PR #11, opened 2020-11-16
Switch works Pre-Faxe

BrandyBuizel:
About to do some breaking shit incorporating Faxe, but this version works well on the switch and does not affect the main version with few exceptions:

- NG API disabled on windows builds(wrong ID, ENC Key, etc.)
- Controls added on controller to restart a song and instantly game over self
- Added Controller key to added health for debugging, both can be disabled in Controls.hx post-merger


ninjamuffin99:
gang

PR #21, opened 2020-12-01
Startskip suboption

lotusotho:
Added an option in the pause menu to skip the start of the song for when the player dies too many times.
It's a little bit buggy and it only works for some songs (dadbattle, south, etc). I tried to detect the song that is playing within the if statements using toLowerCase but it lags a little so I left it there. Still works tho, maybe there's a way to make it fully work?

skipstart


ninjamuffin99:
HMMMMMM something like this i think. Something in the options menu, I like where it's going and I'll play around with it

PR #24, opened 2020-12-03
Fixed issue related with arrows and note trails

lotusotho:
There was an issue related with how the notes were set in the flash xml file. Basically, there's this bug that appears when you hold a note with any of the arrows for a period of time and the camera moves:

trailsbug
 
These are the trails of the notes that are down the confirm arrows and that in the xml they are set too close of each other, so I tweaked it a little bit by changing some pixels without modifying the origin of the animation too much.

image


ninjamuffin99:
ok sure lolol

PR #35, opened 2020-12-06
Fixed loop problem at the end of Boopeebo because of the vocals

lotusotho:
Boopeebo vocals starts right when the song starts playing so when you hit the last note of the song the vocals loop a little bit and continue playing from the start when the code is tying to halt them and load the next.
This does not happen in any other song because they always leave some seconds after the song start.
Anyways, it's a nono to start the vocals at the same time as the song, kinda stinky because of that.


ninjamuffin99:
yolo merge lol

PR #78, opened 2020-12-15
Fixed bf animation when arrows are pressed for too long

lotusotho:
Basically, when any arrow hits correctly any note and you hold it the boyfriend will continue playing the singing animation of that arrow, so I change it with a timer.


ninjamuffin99:
oops, its actually intentional! we'll play with it tho on our own

PR #80, opened 2020-12-15
Fix crash when playing other levels after playing Week 3

dinkstr:
Should fix https://github.com/ninjamuffin99/Funkin/issues/75, https://github.com/ninjamuffin99/Funkin/issues/91, https://github.com/ninjamuffin99/Funkin/issues/83 and possibly https://github.com/ninjamuffin99/Funkin/issues/69

Hopefully I didn't mess this PR up

Fixes the crash that happens if you play a level after playing Week 3, where the game tries to run code on the philly lights but fails because they no longer exist.

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -94,6 +94,8 @@ class PlayState extends MusicBeatState
 
 	override public function create()
 	{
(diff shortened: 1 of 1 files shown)


dinkstr:
Closing because a better fix was committed 😌

PR #169, opened 2021-01-08
fix pico asset paths

Geokureli:
fixes invalid asset paths on the pico stage


ninjamuffin99:
woops messed this up when i was refactoring some path shit for mods the other day

PR #172, opened 2021-01-09
Makes the donate button actually work on Linux.

natis1:
& is not needed as xdg-open will automatically quit after successfully opening the link returning you to the game. Also & doesn't work here. it's given as an argument to xdg-open not an instruction to the system telling it to run the program in the background.

If you absolutely want to run this command in the background you need to use sys.io.process according to this documentation here: https://api.haxe.org/Sys.html

But this isn't even needed. as xdg-open will close on its own after the page starts loading.


ninjamuffin99:
lemme test dis myself and then i'll prob merge hehe

urmum-69:
The version checker also has this problem btw

OPNA2608:
Could you also replace `/usr/bin/xdg-open` with `/usr/bin/env` and an added first argument of `"xdg-open"` please? The current code doesn't even find `xdg-open` on my system because it's in a completely different place.

> `sh: /usr/bin/xdg-open: No such file or directory`

`env` is more portable because it performs a lookup in the user's PATH to find `xdg-open`, and although `/usr/bin/env` isn't a guaranteed path either it's common-enough.

urmum-69:
> env is more portable because it performs a lookup in the user's PATH to find xdg-open, and although /usr/bin/env isn't a guaranteed path either it's common-enough.
why not just straight up use xdg-open without specifying a path? what's the use in having env here?

OPNA2608:
I was assuming it needed an absolute path, but if just `xdg-open` works then yeah go for that.

urmum-69:
yeah, a full path isn't needed
this should just be:
```haxe
 Sys.command('xdg-open', ["https://ninja-muffin24.itch.io/funkin"]);
 ```

natis1:
> yeah, a full path isn't needed
> this should just be:
> 
> ```hx
>  Sys.command('xdg-open', ["https://ninja-muffin24.itch.io/funkin"]);
> ```

Alright. I switched it to this. You’re probably right even if most distros would place it in /usr/bin it’s not something one can be 100% sure of and plus it’s simpler

natis1:
Also added the other change in your pr

Coolguy3289:
Will this be merged soon or has it been fixed upstream since?

NotHyper-474:
Who knew this would eventually turn into https://github.com/FunkinCrew/Funkin/pull/2490 lol

PR #179, opened 2021-01-10
Remove badNoteCheck when hitting too much keys/hitting with no notes on screen

Sakimotor:
Remove maluses when hitting on more notes than required -> closer to the DDR/ITG/StepMania gameplay, makes it somewhat playable on a dance pad

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -1015,6 +1015,11 @@ class PlayState extends MusicBeatState
 					{
 						health -= 0.045;
 						vocals.volume = 0;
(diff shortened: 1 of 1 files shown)


ParzivalWolfram:
I've actually been having an issue with overhits myself. I'd vastly prefer it if they weren't present as well.

Sakimotor (commented):
Resets combo when not hitting a note

ninjamuffin99:
badNoteCheck is intentional for now, but maybe it can be some option later down the line

PR #192, opened 2021-01-14
Fixes timings

KadeDev:
multi-threading is always the solution. Need to optimize this more though.

Otherwise the fix is really good, double, triple, and quad notes are now not based on luck!

Epic!!!!

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -1234,6 +1234,7 @@ class PlayState extends MusicBeatState
 		var downR = controls.DOWN_R;
 		var leftR = controls.LEFT_R;
 
+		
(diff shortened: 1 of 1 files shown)


Thehelper104 (approved):
Coolio!

KadeDev:
> oops, wish i saw this before I did my own thing in my local version!
> Although I will test this also and see which one is nicer hehe

still needs a whole lot of optimizing, its very laggy. Something like this would work a lot though. I'll probably start testing a complete input rework with multi-threading and all.

KadeDev:
Looks like this was already fixed with the new week, great job ninja!

PR #206, opened 2021-01-16
Enable vsync

ActualMandM:
This enables vsync in the project file. Fixes screen-tearing when the game is full screened. (ex. pressing `F` on title screen or `Alt+Enter` anytime when using Windows)


ActualMandM:
Closing as it can just be an option in an options menu if one were to be added

PR #207, opened 2021-01-16
Don't show FPS counter when built with -final

ActualMandM:
Removes the FPS counter at the top left when the game is built with `-final`. This PR also modifies one of the build scripts to build with `-final` instead of `-release`.

This means when someone were to ever build for mobile (it's a pain to set up and crashes after haxeflixel logo lol), it would show the FPS counter as well, unless it were built with `-final`.


ninjamuffin99:
i like the FPS counter so imma keep it :>
will have settings menu eventually where people can disable FPS counter

PR #208, opened 2021-01-16
Shorten prototype line in introText

ActualMandM:
prevents it from cutting off

2021-01-16_14-25-49_Funkin
2021-01-16_14-40-07_Funkin


ninjamuffin99:
hehe, thats part of the joke
its the ONLY line that cuts off like that, because the game IS janky and still a prototype hehehe
gonna close this for now hehehehe

PR #254, opened 2021-01-22
Much better Github-Pages site, but without Jekyll 

regulad:
Since the data required for Jekyll is now absent the page cannot be built locally.


regulad:
I'll be recreating this better at a later date.

PR #214, opened 2021-01-19
can you add keyboard and mouse support.

bruhbruhbruhbruhe:

PR #266, opened 2021-01-23
Suggestions

TheTruestPinky:
I think you should add a straight line control similar to the Left Down Up Right so the new player can easily plays it.


regulad:
Why is this a pull request? It seems to have the same changes as #214, which was closed.

ninjamuffin99:
too many files to check thru im too lazy to see if

PR #249, opened 2021-01-21
Miss counter + accuracy

KadeDev:
Adds a miss counter and accuracy to the score text object on the hud. 

Example:
image

PR #277, opened 2021-01-25
Add outline to score text

ActualMandM:
Allows the score text to be readable in winter stages like Cocoa and Eggnog.

2021-01-24_20-19-56_Funkin
2021-01-24_20-20-10_Funkin

I tried making it easy to read while not looking too bad, but the values can be adjusted if necessary.

(I also tried making the score black for these specific stages instead of using a global outline but I felt like the difference between stages would be jarring)


ActualMandM:
Closed, didn't realize #249

PR #268, opened 2021-01-23
Basic PR/Issue templates

regulad:
Closes #243

Issue #58, opened 2020-12-11
Memory usage issues


Applemunch:
As you keep playing the game normally, (loading in assets, restarts/pauses) the memory usage of the program increases, possibly due to unused information cluttering up the memory. Adding memory dumping for assets after a song (or playlist if you're in story mode) and assets in general could be a good fix.

PR #283, opened 2021-01-25
Fixed memory usage issues

Naishaline:
Basically just tracks assets that are loaded in through add() then removes them when a new scene is loaded. Also some manual asset removal for a few areas that didn't really need all the tracking stuff.

Before this it was re-adding assets on every load, which caused memory usage to build up until the game wouldn't run. Now memory usage is relatively low and stays consistent during gameplay.

I created an issue earlier, but decided I could probably fix it myself and submit a pull request.

Overall not too much was actually added, so should be a pretty small change that makes a big difference. : )


regulad:
Closes #268?

Naishaline:
Also #58! Which I only noticed after making my pull request.

B-Si:
GF assets

nickpythonluajava:
> GF assets
> 
> 

????

AbnormalPoof:
You saw nothing.

PR #288, opened 2021-01-27
Very rudimentary options UI for controls changes

PKPenguin321:
`Commit message:`
```very scuffed, OptionsMenu.hx is now mostly copied from FreeplayState.hx with no actual functionality
currently it reads lines off of controls.txt and displays them on Options menu which is now accessible from main menu
controls.txt has control functions prefixed with "set", IE "setUP", then the corresponding key on the following line IE "W"
```

**This is very much a draft** and currently doesn't actually implement any keybind-changing functionality
Closes https://github.com/ninjamuffin99/Funkin/issues/47

Issues/TODO:
- [ ] Game font does not have characters for `(`, `)`, `:` and possibly more, making the formatting very ugly since I had planned to display lines as `Control - Key` (for example `Restart - Backspace`), either I need new formatting or the artist needs to add more characters to the font
- [ ] Make game read from `controls.txt` at startup and set your bindings correspondingly
- [ ] Actual functionality for the menu buttons that will 1) display text that tells the user it's awaiting input, 2) grabs input, and 3) sets that key and translates the set input back into `controls.txt`
- [ ] Maybe make Left/Right inputs on the options screen tab to other settings like FPS counter toggle, progress reset
- [ ] Allow for more than 1 alternate key per control? On the code side it's doable (and would be cleaner than manually defining an Alternate key for each control) but I'm not sure how the UI would work is the problem
- [ ] Reserve a key that can reset binds to default?
- [ ] Clean up any copypaste errors and commented code

If you're inclined, feel free to take this and tweak/finish it to your own leisure in case I don't get around to finishing it for a while.


PKPenguin321:
Demo: https://streamable.com/qa7f0j
Of note is how lines are just things like `Pause Alternate Enter` rather than the intended `Pause (Alternate) - Enter` due to issue 1
Also lines are too long so formatting will likely need tweaking either way 🤷

ninjamuffin99:
INTRUGING
MAYBE PERHAPS

ninjamuffin99:
I'll merge for now, and clean it up, thanks for gettin these barebones in place!

PR #301, opened 2021-01-31
Allows user to change key binds through options

Injourn:
https://user-images.githubusercontent.com/24728334/106371344-3ab16000-6318-11eb-986f-c0bef7f2b56e.mp4

~~It's still not fully functional unfortunately, But~~ the altered Flxkey is ready for access


Injourn:
I've updated it so now it is fully functional. The only issue so far is that it does not save between session fwiw.

source/Controls.hx
```diff
@@ -492,20 +495,23 @@ class Controls extends FlxActionSet
 		if (reset)
 			removeKeyboard();
 
+		loadControls();
 		keyboardScheme = scheme;
-		
+
+		var keyMaps:Map<String, FlxKey> = FlxMacroUtil.buildMap("flixel.input.keyboard.FlxKey");
+		var i = 0;
```
PKPenguin321, line 502: Unused

PKPenguin321:
Weird that saving between sessions doesn't work, does FlxG.save.data.keyboardMap just not save outside of the current instance of the game? If so you may just have to abandon it and modify controls.txt directly instead. Nice work

Injourn:
Pulled a two for one with the recent commit, now it saves between sessions.

Also yeah forgot to remove the unused code, That's what I get for coding at 1 am.

source/Controls.hx
```diff
@@ -563,6 +567,30 @@ class Controls extends FlxActionSet
 		#end
 	}
 
+	static function loadControls(){
+		if (FlxG.save.data.keyboardMap != null)
+		{
+			keyboardMap = FlxG.save.data.keyboardMap;
+		}
```
PKPenguin321, line 574: Won't doing it like this mean that `controls.txt` isn't used after the first time you launch the game? At that point you could probably just delete it entirely and have the first run generate default controls in this file.

Injourn: Well yes, and honestly I just took the lazy way out. It could be a better idea to remove controls.txt, but as long as it works I don't believe it's too necessary.

PKPenguin321: Leaving controls.txt in is probably misleading to the user since it gives the impression that it can be edited to change controls outside of the game

ParzivalWolfram:
I'd honestly like to see this be expanded on a lot more. I heavily prefer a dual-handed controller layout (D-Pad + face buttons both controlling notes) and I hate having to recompile from source every update (or use antimicro, which drops inputs sometimes) just to allow this.

source/Controls.hx
Injourn: Finally getting back to this, I think that renaming the file to "default controls" would help prevent such a misinterpretation.

source/Controls.hx
Jeffythesmlcool512: case Solo:
				inline bindKeys(Control.UP, [W, FlxKey.UP]);
				inline bindKeys(Control.DOWN, [S, FlxKey.DOWN]);
				inline bindKeys(Control.LEFT, [A, FlxKey.LEFT]);
				inline bindKeys(Control.RIGHT, [D, FlxKey.RIGHT]);
				inline bindKeys(Control.ACCEPT, [Z, SPACE, ENTER]);
				inline bindKeys(Control.BACK, [BACKSPACE, ESCAPE]);
				inline bindKeys(Control.PAUSE, [P, ENTER, ESCAPE]);
				inline bindKeys(Control.RESET, [R]);
			case Duo(true):

PR #327, opened 2021-02-02
Update bug.md

devin503:
add a note to the Local category to tell users to only select one if game is downloaded, and only select for Web if playing on NG or itch.io

I notice a lot of users check a platform in the Local category when playing on web, and vice versa.


regulad (approved):
cool beans

regulad:
@devin503 this should be closed considering i already made a PR that got merged

PR #342, opened 2021-02-03
Update Issue Templates

regulad:
I had made some mistakes when I first submmited this PR as far as formatting and markdown goes.

Similar to #327.

PR #300, opened 2021-01-30
README Rewrite

regulad:
---
labels: documentation
---

Rewrote most of the `README.md` and added a `.vsconfig` to make setting up your build environment easier. If this repo had wikis enabled I would have a wiki page instead, but it doesn't.

PR #343, opened 2021-02-03
Issue Management Workflows

regulad:
There are a lot of issues pouring in, mostly from first time open-source contributors.

These workflows should help clean up all the invalid and spam issues.

Note: This is more of a suggestion than anything else. There are a lot of invalid/spam issues and it may be hard to cut through the noise and find legitmate issues that nee fixing. Semi-dependant on #342 and #300.

### DIFF
--- a/.github/workflows/first-interaction.yml
+++ b/.github/workflows/first-interaction.yml
@@ -1,0 +1,24 @@
+name: First Interaction
+
+on:
+  issues:
+    types:
+      - opened
+      - edited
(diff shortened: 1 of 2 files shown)


WaviestBalloon:
> Note: This is more of a suggestion than anything else. There are a lot of invalid/spam issues and it may be hard to cut through the noise and find legitimate issues that need fixing. Semi-dependant on #342 and #300.

Fixed 3rd sentence

PR #357, opened 2021-02-04
Fixed bug that makes the whole game crash at the start of week 6

lotusotho:
Basically, if you start hitting any key while senpai starts speaking at the start of the week the game will crash because is not fast enough to get the dialogue info before going to the next line:

image

Here's the output:
image

I worked around it by having to check if the dialoglist array has any info before triggering anything else


ninjamuffin99:
GOOD ENUF PROBABLY

PR #359, opened 2021-02-04
Fix the song Spookeez

### LINKED ISSUE #352: Enhancement: Spookeez overhaul
Not really sure whether this counts as a bug report or an enhancement but the game is technically working as intended, so I'll go with enhancement.

The first song of Week 2, Spookeez, has two glaring issues:
1. The voice track is slightly ahead of the music track. For what it's worth, this isn't present in the HTML5 versions of the game but in all other footage I've seen of the song being played, it's desynced.
2. The charts are very inaccurate. A lot of mistimed notes, missing notes, extra notes and some other funky stuff that makes it hard to FC. It was only the second week but then again, South doesn't have these issues so it should still be revised.

I submitted [a mod](https://gamebanana.com/gamefiles/14667) to GameBanana that fixes both of these issues but I would still like some sort of official fix for this. I thought about making a pull request with these changes but it feels a bit weird or improper to PR a mod. Let me know if I should follow up on that.

NickHam13:
Overhauled the charts and fixed the small desync between the voice/music track. Closes #352.


Helper006:
Even if this PR is declined, it is an issue that should be addressed. Frustrating especially in Hard Mode.

NickHam13:
> Even if this PR is declined, it is an issue that should be addressed. Frustrating especially in Hard Mode.

According to someone else, ninjamuffin was planning on recharting it himself so even if this gets closed, at least it'll be fixed at some point

ninjamuffin99:
yeah I would like to redo it myself, just cuz its one of the things that should be MY responsibility.
although i will keep this PR open so that it will haunt me until I do overhaul and fix it

PR #362, opened 2021-02-05
check if newgroups api key actually exists before creating core

Jan200101:
Following the README and using this as the contents for `APIStuff.hx`
```
class APIStuff
{
	public static var API:String = "";
	public static var EncKey:String = "";
}
```
results in the game quitting due to the invalid API key
lets check if the api key is not empty before creating the core so we don't run into this


ninjamuffin99:
looks good nice nice!

AyeTSG:
One problem with this, if you go into the freeplay menu with this implementation, the game WILL crash!

Issue #183, opened 2021-01-11
Debug mode: ChangeBPM problems


ILuvGemz:
I have discovered a bug with the ChangeBPM system. 

I recently found out that when I was charting a 240 BPM song and tried changing the BPM to 180 on section 72, my notes ended up being placed in 96000 instead of 72000.

From what I have gathered, the notes placement is in milliseconds, with this equation:

([Section number * 240 / Normal BPM + (240 / Normal BPM * y / 16)] * 1000) milliseconds, where y is 0 <= x <= 15

y is equal to the position of the note in steps, with the first being 0 and the last being 15.

Therefore, I found out when I use ChangeBPM, it overrides the normal BPM variable of the song, thus making me chart later or sooner than where I expected it to be.

I hope you are able to understand my problem and found a way to fix it!

-Zenokwei a.k.a ILuvGemz

ninjamuffin99:
changing BPM is not supported yet.

ILuvGemz:
Ummm... Okay. I'll wait.

PR #382, opened 2021-02-08
BPM change functionality & implementation

PrincessMtH:
Tweaks changeBPM functionality so that the BPM is not reset back to the original after having been changed.

Worth noting is that this may cause inconsistencies with the chart editor, particularly in the function ``recalculateSteps()``.
A potential fix for this could be done by doing this for its for loop instead:

```haxe
var bpm:Int = _song.bpm;

for (i in 0...curSection)
{
	steps += 16;

	if (_song.notes[i].changeBPM)
		bpm = _song.notes[i].bpm;
	timeShit += (((60 / bpm) * 1000) / 4) * 16;
}
```
But the chart editor scares me so I do not touch it, not that there are many songs I could test with.

### DIFF
--- a/source/ChartingState.hx
+++ b/source/ChartingState.hx
@@ -2,6 +2,7 @@ package;
 
 import Section.SwagSection;
 import Song.SwagSong;
+import Conductor.BPMChangeEvent;
 import flixel.FlxG;
 import flixel.FlxSprite;
 import flixel.addons.display.FlxGridOverlay;
@@ -117,11 +118,9 @@ class ChartingState extends MusicBeatState
 				song: 'Test',
 				notes: [],
 				bpm: 150,
-				sections: 0,
 				needsVoices: true,
 				player1: 'bf',
 				player2: 'dad',
-				sectionLengths: [],
 				speed: 1,
 				validScore: false
 			};
@@ -140,6 +139,7 @@ class ChartingState extends MusicBeatState
 
 		loadSong(_song.song);
 		Conductor.changeBPM(_song.bpm);
+		Conductor.mapBPMChanges(_song);
 
 		bpmTxt = new FlxText(1000, 50, 0, "", 16);
 		bpmTxt.scrollFactor.set();
@@ -366,6 +366,7 @@ class ChartingState extends MusicBeatState
 			vocals.time = 0;
 			FlxG.sound.music.pause();
 			FlxG.sound.music.time = 0;
+			changeSection();
 		};
 	}
 
@@ -425,6 +426,7 @@ class ChartingState extends MusicBeatState
 			else if (wname == 'song_bpm')
 			{
 				tempBpm = Std.int(nums.value);
+				Conductor.mapBPMChanges(_song);
 				Conductor.changeBPM(Std.int(nums.value));
 			}
 			else if (wname == 'note_susLength')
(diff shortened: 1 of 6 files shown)


ILuvGemz:
It doesn't actually work. I tried merging it and it says some variables are missing

PrincessMtH:
After some more testing, I think this requires extra work before it can be merged. Not because the proposed change causes any new issues, but because the rest of the existing system for changing BPM is not yet functional.

Changing the BPM also updates the Conductor's crochet, which is used along with the absolute song position to calculate the current step/beat etc. (curStep, totalSteps, totalBeats). If the BPM changes, the calculation will suddenly be very different, and while the BPM will appear to change visually, it's most likely misaligned with the actual beats of the song. This also messes up calculations for what the current section is, which can eventually lead to the game trying to read from sections that do not exist, and thus crashing.

The solution here is to keep track of when the BPM changes and work relative to the last change. Still thinking about what the best way to do this could be. Perhaps BPM changes should be mapped out on song load? Keeping track of changes as they happen could be impacted by lag, and keeping track of it within the Conductor class is impractical as it doesn't ever seem to reset. Regardless, _something_ is gonna require a bit of an overhaul. Ideas are welcome!

ILuvGemz:
Thank you. This is the EXACT issue I have been suffering from.

PrincessMtH:
In making the ChartingState compatible with this new BPM system I ended up overhauling it a bit more than I intended. Here's a quick demo of how it fared with a rechart of Monster with proper BPM changes (see below):

[Demonstration video](https://youtu.be/eQNhvwhpQbY)

The strumLine follows the tempo nicely and does not typically escape the bounds of the grid. Starting playback anywhere in the middle of the track isn't an issue either. Not much testing done with actual chart editing though, but placing notes seems to work just fine. After setting the BPM for a section, you may need to go back and forth to the previous one to have it reflect the new BPM. Vocals also stop playing after the song loops. Still figuring those out.
PlayState needs some extra changes too. Stuff like the screen zooms and GF dancing seem to update but desync somewhere. Still gotta look into that. So far so good, though!
[(monster rechart here)](https://github.com/ninjamuffin99/Funkin/files/5968764/monster.zip)

ILuvGemz:
Hello! I’m sorry to be rude, but can you post it as a mod when you’re done? If not, I understand.

PrincessMtH:
In-game events (animations, fx etc.) now sync properly even after BPM has changed.
Chart editing requires more testing before I'm confident it can be merged, but I suppose that's what these things are reviewed for, so I'll open it up again. Should fix #183, though.

> Hello! I’m sorry to be rude, but can you post it as a mod when you’re done? If not, I understand.

No problem! If you mean the Monster chart, I suppose I already have in the comments here. It kind of depends heavily on the BPM changes, so unless the code gets merged, I don't think I'll release it anywhere else.

ILuvGemz:
That’s cool! I wanted a version of FNF with changeBPM (under review) and section length changing (which I know you haven’t finished) so much, to the point where I put the entire thing on a 225 GameBanana point bounty! (If the section length is changeable, I would give it to you per request)

If you do not want to continue after the ChangeBPM, then I understand.

PrincessMtH:
[Video demonstration of the Monster map](https://youtu.be/Onr4_xcEO58) (above). Visual quirks like the camera zoom, background elements (philly train/lights, halloween lightning etc.) all sync more accurately after the latest change.

> That’s cool! I wanted a version of FNF with (...) section length changing

In the end, anything is possible, but a lot of the game's logic is built around having 4 steps in a beat and 4 beats in a section, regardless of what lengthInSteps wants you to believe. So it would just require a lot of work, and lengthInSteps would probably have to be replaced with some kind of time signature equivalent so the game knows how long a beat and a step are individually.

ninjamuffin99:
REAL GANGSTER HOURS MUTHUFUKIN BLES UP GOT BLESS U MTH

ILuvGemz:
May I distribute the compiled .exe with changeBPM?

ILuvGemz:
May I distribute the compiled .exe with changeBPM?

ninjamuffin99:
> May I distribute the compiled .exe with changeBPM?

hmm if u want yeah, although later today I WILL update game with a 0.2.7.1 update if u wanted to wait for that hehehe

doom699:
epic

PR #402, opened 2021-02-11
Use haxe switch case 'or'

TheDrawingCoder-Gamer:
Changes the static arrows code to use switch case "or" instead of copy pasting. hopefully will make things neater in the future.


ninjamuffin99:
OH LOL FOR LIKE 3-4 YEARS I DID NOT KNOW THIS LOL I ALWAYS USED 2 OF THESE 
i knew i had to be doing something wrong....

PR #427, opened 2021-02-15
score value ease-out fix + song caching

jvc2:
tiny fix so that when the week score transitions from a smaller value to a greater one, it reaches the intended value. currently, if the score tries going up to a value like 10300, it gets stuck at 10299.
also changed the free play menu score's lerp ratio to be 0.5 like the story menu, just for some uniformity idk


jvc2:
should i have opened a different pull request for that one? i have no idea what i'm doing.
but anyway, caching the songs prevents a very short freeze right when the song begins to play (especially noticeable on bopeebo at dad's first note). i wasn't sure where to add that paragraph so feel free to move it elsewhere.

PR #447, opened 2021-02-17
score value ease-out fix

jvc2:
redoing because i thought i could learn git and now i am experiencing hell

tiny fix so that when the week score transitions from a smaller value to a greater one, it reaches the intended value. currently, if the score tries going up to a value like 10300, it gets stuck at 10299.
also changed the free play menu score's lerp ratio to be 0.5 like the story menu, just for some uniformity idk

### DIFF
--- a/source/FreeplayState.hx
+++ b/source/FreeplayState.hx
@@ -159,7 +159,11 @@ class FreeplayState extends MusicBeatState
 			FlxG.sound.music.volume += 0.5 * FlxG.elapsed;
 		}
 
-		lerpScore = Math.floor(FlxMath.lerp(lerpScore, intendedScore, 0.4));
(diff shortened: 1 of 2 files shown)


PrincessMtH:
freeplay already has code that jumps lerpscore to intendedscore so the added code there is redundant, might be better to just copy that over to the week score stuff instead lol

jvc2:
you mean `if (Math.abs(lerpScore - intendedScore) <= 10) lerpScore = intendedScore;`? i think i was going to remove it but forgot, oops. could have copied it to storymenu but i prefer jumping when diff=1 rather than 10, even though it doesn't matter that much.

jvc2:
closing this because i found a cleaner solution that circumvents all floating point rounding nonsense

PR #412, opened 2021-02-13
check if response result is actually there

Jan200101:
follow up on #362
fixes #410
the thing that was causing problems with an invalid API key was response.result being null since the response Newgrounds gives just tells us that we have an invalid app id.

Since we now actually create NG.core we will always get a console message when anything with NG.core is done that requires the api.

PR #458, opened 2021-02-19
cache songs before playing + smoother game over chord

vinjcc:
redoing this too
> but anyway, caching the songs prevents a very short freeze right when the song begins to play (especially noticeable on bopeebo at dad's first note). i wasn't sure where to add that paragraph so feel free to move it elsewhere.

+ this time also for game over screen
+ game over music fades out for 0.1 secs instead of immediately stopping, to blend in nicer with the retry chord (love that btw)

tiny changes always


ninjamuffin99:
HMMM caching would be useful, I never saw that in the FlxG.sound stuff lolol. I need another good read through of the documentation.

There's a big overhaul to the assets loading that I'm just about to merge, I'm not sure if it addresses the caching stuff, BUT if it doesn't I think this would def be helpful hehehe

jvc2:
[i happened upon it here if you're wondering.](https://api.haxeflixel.com/flixel/system/frontEnds/SoundFrontEnd.html#cache) i know almost nothing about haxe or flixel, i just like messing with code and burying my face in manuals.

also just noticed songs don't freeze in the web version, maybe because mp3s get decoded much faster than oggs. caching is still useful for download versions though.

ActualMandM:
Would it even be possible to implement caching into the freeplay menu? A lot of freezes happen there because it plays the song that is selected.

ninjamuffin99:
it prob would be possible, although I am worried about memory issues on a bit lower end hardware. BUT for now it can prob be implemented and can see if people are running into issues!

jvc2:
in the freeplay menu it might be better to load songs concurrently somehow, so it doesn't freeze waiting for the song to load every time you change the selection (songs wouldn't play immediately though). i might try caching every song anyway just to see what happens.
(ps: pls merge #412)

ActualMandM:
tried caching the freeplay menu on my end, honestly feels like a no-go to me
long load time into the menu and causes a huge memory spike

i added the cache function into addSong if you were curious

```java
	public function addSong(songName:String, weekNum:Int, songCharacter:String)
	{
		songs.push(new SongMetadata(songName, weekNum, songCharacter));
		FlxG.sound.cache(Paths.inst(songName));
	}
```

https://user-images.githubusercontent.com/15317421/109599989-2a380500-7ad1-11eb-8000-9daa7af486ac.mp4

prokube:
>  i.hate.file.size.limits.sometimes.mp4

what mod adds the character icons to the freeplay menu

aflacc:
> > i.hate.file.size.limits.sometimes.mp4
> 
> what mod adds the character icons to the freeplay menu

this is something in the latest version of the github

SimplyMerlin:
I'm pretty sure this doesn't do anything because FlxG.sound.playMusic doesn't try to use a cached sound, you'd have to swap out playMusic to play for that.

ActualMandM:
> I'm pretty sure this doesn't do anything because FlxG.sound.playMusic doesn't try to use a cached sound, you'd have to swap out playMusic with play for that.

freeplay uses playmusic as well and the caching worked fine there (see video above)

VMGuy23:
> it prob would be possible, although I am worried about memory issues on a bit lower end hardware. BUT for now it can prob be implemented and can see if people are running into issues!

Memory issues are quite a big thing. 5% of issues are black sprites or something

SimplyMerlin:
> > I'm pretty sure this doesn't do anything because FlxG.sound.playMusic doesn't try to use a cached sound, you'd have to swap out playMusic with play for that.
> 
> freeplay uses playmusic as well and the caching worked fine there (see video above)

neither the api nor the code itself show anything that would imply that caching actually does anything when used with playMusic

PR #463, opened 2021-02-19
Recenters numscore and rating relative to camfollow

rifxii:
This should fix issues where the numscore and rating are extremely off-centered and are placed near the top of the screen in some stages. 
comparison
This is my first PR so hopefully nothing's messed up.


JulianWhitewood:
good shit

PR #464, opened 2021-02-19
Edited README to fix compiling issue

TentaRJ:
Haxe 4.2.0 wasn't working, so I edited the guide to fix it and download Haxe 4.1.5 instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,8 @@ IF YOU WANT TO COMPILE THE GAME YOURSELF, CONTINUE READING!!!
 ### Installing shit
 
 First you need to install Haxe and HaxeFlixel. I'm too lazy to write and keep updated with that setup (which is pretty simple). 
-The link to that is on the [HaxeFlixel website](https://haxeflixel.com/documentation/getting-started/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,92 +1,91 @@
...
+flixel-ui
+newgrounds
+```
+
+So for each of those type `haxelib install [library]` so shit like `haxelib install newgrounds`
+
+You'll also need to install polymod. To do this, you need to do two things first.
+1. For Windows: [Download GitForWindows](https://gitforwindows.org/) to use gits.
+2. Run `haxelib git polymod https://github.com/larsiusprime/polymod.git` after GitForWindows is installed
+
```
TentaRJ: This is the main edit of the readme

TentaRJ (commented):
Added comments

jvc2:
i've also ran into this issue before and found it described here: https://github.com/HaxeFlixel/flixel/issues/2294
presumably compiling with 4.2.0 will work when lime and openfl are eventually updated, so it isn't specifically an issue with haxe's version. i don't know if @ninjamuffin99 is down for keeping up a list of versions he's using, but it would prevent a lot of trouble with compilation (especially for me when i first tried compiling with 4.0.5 which was too old, and then 4.2.0 which was too new) (thanks haxe)

PR #468, opened 2021-02-20
Added a current song text & added a outline to the score and current song text.

JulianWhitewood:
I have added a current song text, that way, the players would be able to tell which song is playing.
I have also added a outline to both of the score and current song texts, since sometimes its a bit hard to see. (especially for week 5).

Not much, but I think its pretty effective.
Shoutout to VoltrexMaster for basically turning my 75+ line code into a few lines LOL

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -676,12 +676,12 @@ class PlayState extends MusicBeatState
 		add(healthBar);
 
 		scoreTxt = new FlxText(healthBarBG.x + healthBarBG.width - 190, healthBarBG.y + 30, 0, "", 20);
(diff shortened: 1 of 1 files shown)


PrincessMtH:
This is a cool idea, but it seems a bit much to have it on-screen the entire song, for every song, with no option to disable. I think it would fit better on the pause menu, instead.

JulianWhitewood:
> This is a cool idea, but it seems a bit much to have it on-screen the entire song, for every song, with no option to disable. I think it would fit better on the pause menu, instead.

on it ;)

ninjamuffin99:
OOPS
did this with 3d46cdc on the pause menu, but thanks for PR anyways hehe i appreciate it!

PR #521, opened 2021-03-02
Easy modding

VitalieW0WKid:
I came up with and idea to make an even easier modding experience:
It is possible to be a folder of the game in the android data folder,similar to the pc modding technic, instead of using apk editor.Ok maybe is useless but that would be cool


ninjamuffin99:
gon close this cuz it got a lot of wacky little commits and all that sorry gangsta!!!

PR #540, opened 2021-03-05
wait before playing song on freeplay menu

vinjcc:
solves an issue that was mentioned on #458

waits one second before playing the selected song. there's still a short freeze but it's much less of a hindrance.
not a very smart fix but it's an improvement. if a kind, smart soul can figure out how to load+play songs asynchronously (if possible), that would be preferable.

pros:
+ no freezes when scrolling through freeplay menu
+ doesn't load every song in the game to memory

cons:
+ songs won't play immediately when selected


source/FreeplayState.hx
```diff
@@ -271,7 +274,10 @@ class FreeplayState extends MusicBeatState
 		#end
 
 		#if PRELOAD_ALL
-		FlxG.sound.playMusic(Paths.inst(songs[curSelected].songName), 0);
+		songWait.cancel();
```
ActualMandM, line 277: If you want to have the music pause between each selection:
```suggestion
		FlxG.sound.music.stop();
		songWait.cancel();
```

jvc2: (sorry for the delayed response)
i think letting the previous selected song play while scrolling is nicer, but that's just me.
i'll leave the decision to nm99 if he merges

PR #542, opened 2021-03-05
Remove garbage data from end of songs

treacherousfiend:
remove lots of null characters from the ends of almost every song, saves a surprising amount of file space, probably also makes IO slightly quicker?
also some other misc fixes to songs, like removing a new line from the start or end of a song.

edit: the differences don't seem to show up correctly on github, but opening up the current files in most text editors will show the null characters


johnneijzen:
image
image

i wonder does cause NUL at end since it not line ending difference since both of them are using CRLF endings

treacherousfiend:
that was something i had been thinking about as well. the game does not use them at all, as i've been able to play all the songs just fine, and some of the charts didn't have any NUL characters to begin with.

the amount of NUL characters did seem to be based on song length, with longer songs having more NUL characters

ninjamuffin99:
> i wonder what does cause NUL at end since it not line ending difference since both of them are using CRLF endings

It's something to do with the way either Haxe / OpenFL saves files I'm guessing. All of the charts with the ending line stuff has the weird garbage data at the end that is useless. If you wanna maybe look into it, I'd peep the way the game [currently saves the .json song data](https://github.com/ninjamuffin99/Funkin/blob/master/source/ChartingState.hx#L1039-L1088). And then also dig into the [FileReference class of OpenFL ](https://github.com/openfl/openfl/blob/develop/src/openfl/net/FileReferenceList.hx) in general too

ninjamuffin99:
also sure I'll merge hehehe

jvc2:
@ninjamuffin99 the trailing NULs seem to happen only when you save a song in the web version, just tried this out.

maybe see if lime 7.9.0 fixes this? (see https://github.com/haxelime/lime/pull/1438)

PR #543, opened 2021-03-05
Remove readability from some songs json files

treacherousfiend:
Remove the readability from the json files for Eggnog (Hard) and Winter Horrorland (all difficulties), removing the readability also maintains consistency with the other songs json files.
Winter Horrorland is now 1/3 the size it was before.


ninjamuffin99:
the inconsistencies keep anyone digging thru the files on their toes, keeps it SPICY

treacherousfiend:
fair enough lol, i just like micro-optimizations and consistency

treacherousfiend:
just would like to note that all winter horrorland charts were 1 line like this before being regressed in 671dced, and eggnog (hard) was like this before 954dd76.

PR #553, opened 2021-03-08
Compress images via FileOptimizer

exurd:
This pull request compresses the images in the game with the program [FileOptimizer](https://nikkhokkho.sourceforge.io/static.php?page=FileOptimizer). There are no visible differences between the optimized images and the original images.
I used the best possible settings in the program without using lossy compression. Therefore, it _literally_ took over two days or so to complete the process. 

> "boyfriend.png" took over **14 hours or so** and _i almost regret it_

Note that any new assets that are not yet included in the repository will not be compressed (_Y'know_, like week 7, 8, 2000 or whatever). The assets will have to be optimized separately after they're released.


treacherousfiend (approved):
Saves about 19MB. I noticed no differences in quality ingame or when viewing the assets externally

ninjamuffin99:
are you able to post a few side by sides?

exurd:
I'm thinking about **closing this pull request and redoing it,** using [DeveImageOptimizerWPF](https://github.com/devedse/DeveImageOptimizerWPF) instead. This program still uses FileOptimizer but includes a feature that checks for errors in the image; comparing each pixel for differences. It can also process multiple images at the same time (_yes, I optimized each image one by one_).

I'd rather make sure that each image looks the same with no difference in how they look and do this task in a single commit. (Also, I kinda despise the name of the commit. _'Compress images'_ would've sufficed.)

**_However, I won't be closing this PR until I have completed compressed images from that program._**
It might take a while as I'm doing it on my idle pc (which is _slower_) and could take up to three days or so to complete. But I feel like it's worth it to make sure that the images are pixel-perfect.

deve_progress

exurd:
Also, here's the side by side screenshots that you wanted. Some have been taken a few milliseconds apart.

### Unoptimized
dad_unoptimized
mom_unoptimzed
spooky_unoptimized

### Optimized
dad_optimized
mom_optimized
spooky_optimized

I can't see any visible differences, but I'm just making sure that they look correct in-game. Sorry that I took a while to reply,

treacherousfiend:
i feel like there must be a better alternative to the programs you've been using.
i decided to use DeveImageOptimizerWPF and the time it takes to optimize an image just is waaaay too long. my current hardware isn't the best but it would take 30-40 minutes for a hud texture and probably hours for a character.

after doing that i decided to try my image optimizer of choice pingo (which is used in image optimizer) and all of the files were compressed in a little over 8 minutes. (compressed by 19-20MB)
Unfortunately i've had no luck finding a good comparison tool so i'm not 100% sure that there are no visual differences, but i saw none when playing with it.

Side note: image optimizer uses maybe 10 different programs on the images to compress them (with pingo being one of them),
Which may be why it takes so long. i really don't see why that many different compression programs need to be run on an image, seems like they'd just be stepping over each other and wasting time.

exurd:
> i decided to use DeveImageOptimizerWPF and the time it takes to optimize an image just is waaaay too long. my current hardware isn't the best but it would take 30-40 minutes for a hud texture and probably hours for a character.

It's usually due to [ECT](https://github.com/fhanau/Efficient-Compression-Tool) that makes it take a long time. I think that maybe setting it to best (which is 9) might not be the best of ideas; it takes longer for a smaller and smaller difference. It's like lightbulbs. 1 lightbulb to _2_ lightbulbs is a **massive change**, but changing from 99 lightbulbs to _100_ lightbulbs is a small change.

On one hand, I want to get the best compression I possibly do since this game is web-based. On the other hand, it takes too long and gives out a tiny file reduction compared to the other programs, which do more.

> after doing that i decided to try my image optimizer of choice [pingo](https://css-ig.net/pingo) (which is used in image optimizer) and all of the files were compressed in a little over 8 minutes. (compressed by 19-20MB)

That sounds pretty great. Funnily enough, I went for my optimizer of choice, which was FileOptimizer (it also includes pingo). But due to ECT, it takes over 14 hours for a massive image like 'boyfriend.png' to process. Good news, though. I have found out that you can disable that by adding it to the plugin exclude mask in the settings!

However, DIO (Short for DeveImageOptimizerWFP because _it's way too damn long to say_) does change/ignore most (if not all) of the settings in the program you originally had by replacing the .ini file used by FileOptimizer. That means that ECT will still be used. The only way right now to fix this is to delete ECT.exe from ``.\FileOptimizer\Plugins64``. I have created an issue (available at devedse/DeveImageOptimizerWPF#7) about this.

_Anyways._ Thanks to you, @treacherousfiend, this will now only take minutes and **not days**. It might not be as compressed, but at least it's not _as slow as a slow cooker_ anymore and my whole life is less wasted (_than usual_).

exurd:
It's finally finished processing the files. I will be closing this PR in favour of making a new one.

exurd:
Somehow the program missed one file (_"mom_dad_christmas_assets.png"_) and now I have to wait for that one. Unfortunately, it's one of those big ones. Might take a while to finish.

PR #567, opened 2021-03-11
Added Color Blindness Filters to help people that suffers from it play the game better

lotusotho:
Basically the filters work as matrix colors that are applied onto the cameras. They also can be changed in the options menu.

Video showing how the filters work ingame:
https://drive.google.com/file/d/1HyUPJTtWetnijH0XYwo9N7I9SSH9aHSp/view?usp=sharing

- Deuteranopia:
image

- Protanopia:
image

- Tritanopia:
image

The cool thing about this is that it may help people see the arrows better on some stages, test how visible the interface is in some cases, etc. Also the matrix could be used to make your own custom filters.

Right now, this only works on PlayState, but maybe later could also be implemented on the main menu if it's a good idea


Hendicted (commented):
why are you importing everything from PlayState? nothing major just why

lotusotho:
Whoops my bad I thought I changed that.. Was messing around the code and forgot. Thanks!!

PurpleButNotALight:
wait for 0.2.8 so we have a actual settings menu lol

Steviegt6:
> wait for 0.2.8 so we have a actual settings menu lol

This PR was made back in March...

Hendicted:
> wait for 0.2.8 so we have a actual settings menu lol

can easily be implemented in another engine lol

PrincessMtH:
does this actually help for colourblind accessibility? afaik this just makes the game look like how someone with colourblindness would see it, and thus it won't make any difference to those who already have it. i could be wrong but this seems relatively unhelpful?

a different feature we've been toying with for later updates is setting custom note colours and outlines. so you could set it as easily distinguishable colours specific to your variant of colourblindness, or just high contrast, or just do fun customization stuff in general. if a fullscreen colour filter is actually helpful it will definitely be considered though!

PR #586, opened 2021-03-15
Clean up dialogue related stuff in PlayState

ggvbo:
Separates dialogue into text files similar to how its being done in the latest weeks

### DIFF
--- a/assets/preload/data/tutorial/tutorialDialogue.txt
+++ b/assets/preload/data/tutorial/tutorialDialogue.txt
@@ -1,2 +1,2 @@
 :gf:Hey you're pretty cute.
-:gf:Use the arrow keys to keep up \nwith me singing.
+:gf:Use the arrow keys to keep up with me singing.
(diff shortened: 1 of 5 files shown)


assets/preload/data/tutorial/tutorialDialogue.txt
```diff
@@ -0,0 +1,2 @@
+:gf:Hey you're pretty cute.
+:gf:Use the arrow keys to keep up \nwith me singing.
```
ActualMandM, line 2: no need to have the `\n`

GNosii: removed it

hdboye:
It's never actually used in the game right now, so what's the point of doin it?

TentaRJ:
> It's never actually used in the game right now, so what's the point of changing it?

Future use, I guess? I don't suppose this will be used anytime soon... but it might be good to have.

PR #590, opened 2021-03-15
fix combo counter (so it can go above 999)

KadeDev:
Basically, we convert the combo int into a string array and iterate through it and add all of the numbers to separate score instead of some weird-ass math shit.

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -1761,10 +1761,17 @@ class PlayState extends MusicBeatState
 		rating.updateHitbox();
 
 		var seperatedScore:Array<Int> = [];
+		
+		var comboSplit:Array<String> = (combo + "").split('');
 
-		seperatedScore.push(Math.floor(combo / 100));
-		seperatedScore.push(Math.floor((combo - (seperatedScore[0] * 100)) / 10));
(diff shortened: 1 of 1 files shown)


Smokey555:
Based

Le-ECE:
You can set it to just iterate through the indices of 1 string.

var comboString:String = (combo + "");

if (comboString.length == 2)
	seperatedScore.push(0);

for(i in 0...comboString.length)
	seperatedScore.push(Std.parseInt(comboString.charAt(i)));

The-SGPT:
thx now I don't have to hear the shitty blood from haxeflixel command prompt

KadeDev:
> You can set it to just iterate through the indices of 1 string.
> 
> var comboString:String = (combo + "");
> 
> if (comboString.length == 2)
> seperatedScore.push(0);
> 
> for(i in 0...comboString.length)
> seperatedScore.push(Std.parseInt(comboString.charAt(i)));

you realize that's what I'm doing right?

Le-ECE:
I meant instead of creating an array of strings from splitting one string and reading each string, you can iterate through a single string as an array of chars. Saves a little bit of memory but every little bit counts!

PR #626, opened 2021-03-21
Remove Duplicate Notes in Spookeez

meIIoe:
Section 57 in Spookeez has all of its notes duplicated on all difficulties. You only need one input to hit each pair of notes. This affects the combo counter by counting up by 2 every time a pair of these notes are hit and affects the score by awarding double points. This is a minor problem, only affecting 5 notes in Easy, 6 in Normal, and 6 in Hard.


actuallyasmartname (approved):

PrincessMtH:
week7 update recharts fix this issue

PR #628, opened 2021-03-21
vsync pog

itskwartz:
title says it all


CaptainGame17:
doesnt fnf already have vsync (or an fps limit?)

itskwartz:
There's an fps limit on the downloadable versions and vsync on web versions, this is just for people with high refresh rate monitors who want to use the downloadable versions (alternatively, they could replace Lib.application.window.displayMode.refreshRate with any integer they want if they want a specific framerate ig)

ninjamuffin99:
INTRIGUING

itskwartz:
There's also the framerate variable in Main but that only seems to be changeable outside of runtime & it makes the fps locked things even faster than the window.frameRate solution (no idea why)

itskwartz:
And before anyone asks why I've added them to specifically those 3 states; pretty much every interactable state besides AnimationDebug and PauseSubstate extends MusicBeatState (so I don't need to copy and paste one extra line of code into every single state), and losing and regaining focus sets the framerate back to the variable in Main.

itskwartz:
forgot to mention; enable vsync for both 'all targets' & 'windows specific' (this corrects the fps counter and it allows you to freely change your monitor refresh rate during runtime without having to restart)

itskwartz:
yea this pr became a train wreck real quick so to clarify, vsync = true for 'all targets' & 'desktop specific' in Project.xml and the other code goes in MusicBeatState, MusicBeatSubstate and AnimationDebug

itskwartz:
Ok i think i just figured out how vsync on web works so this pr may or may not get closed in ~24 hours just so i can make a fresh one that doesn't look this awful

i swear i know what i'm doing lmao

itskwartz:
Actually, I'll just close it now and see what happens tomorrow

PR #629, opened 2021-03-21
Optimize images with DeveImageOptimizerWPF

exurd:
Redo of #553.

This pull request optimizes all of the images (except for the two bmp files) using [DeveImageOptimizerWPF](https://github.com/devedse/DeveImageOptimizerWPF).

This program uses FileOptimizer to optimize the files. It then checks for errors in the optimized image, comparing each pixel for differences. This means that each image here has no errors in the image. Optimizing them means they take up less space and can make the whole game quicker to download.

This was done quicker due to me removing [ECT.exe](https://github.com/fhanau/Efficient-Compression-Tool) from the process, making it much MUCH faster.

Like before, any new assets that are not yet included in this repository will not be compressed.


oddMLan:
You could optimize the images even further by using pngquant. The game uses a toony style so it should work fine with PNG-8 images. Unless Haxel doesn't support PNG-8 textures.

exurd:
FileOptimizer already includes pngquant. This means that the images have already been optimized with that program, along with 49 other plugins (what programs are called by the developer) as well.

johnneijzen:
one thing that maybe become problem is image decoding time in my opinion is just really small different but some developer have other opinions about it

CyndaquilDAC:
> one thing that maybe become problem is image decoding time in my opinion is just really small different but some developer have other opinions about it

This is for size optimization, not loading times.

johnneijzen:
> This is for size optimization, not loading times.
but it can affect loading times

This is from other a project but based on their analysis it does affect it
image
you can use this link to see their testing methodology how different encoding algorithms can affect decoding speed
https://github.com/peppy/dotnet-benchmarks/tree/master/PngDecodeBenchmark

PR #635, opened 2021-03-23
Excluded the debug mode from Newgrounds API

lotusotho:
It was causing a lot of problems so because is debug mode better to just leave it out


lotusotho:
To be fair, this would work if you had your Newgrounds account attached while debug, the problem is that you cannot go to freeplay menu without the game crashing because of this

PR #640, opened 2021-03-24
feat(performance): Use switch statement

VoltrexMaster:
Use a `switch` statement instead of huge `if .. else if` chains.

### DIFF
--- a/source/PlayState.hx
+++ b/source/PlayState.hx
@@ -227,344 +227,347 @@ class PlayState extends MusicBeatState
 		DiscordClient.changePresence(detailsText, SONG.song + " (" + storyDifficultyText + ")", iconRPC);
 		#end
 
-		if (SONG.song.toLowerCase() == 'spookeez' || SONG.song.toLowerCase() == 'monster' || SONG.song.toLowerCase() == 'south')
+		switch (SONG.song.toLowerCase())
 		{
-			curStage = "spooky";
-			halloweenLevel = true;
+                case 'spookeez' | 'monster' | 'south': 
+                {
+                  curStage = 'spooky';
+	            halloweenLevel = true;
 
-			var hallowTex = Paths.getSparrowAtlas('halloween_bg');
+		    var hallowTex = Paths.getSparrowAtlas('halloween_bg');
 
-			halloweenBG = new FlxSprite(-200, -100);
-			halloweenBG.frames = hallowTex;
-			halloweenBG.animation.addByPrefix('idle', 'halloweem bg0');
-			halloweenBG.animation.addByPrefix('lightning', 'halloweem bg lightning strike', 24, false);
-			halloweenBG.animation.play('idle');
-			halloweenBG.antialiasing = true;
-			add(halloweenBG);
+	            halloweenBG = new FlxSprite(-200, -100);
+		    halloweenBG.frames = hallowTex;
+	            halloweenBG.animation.addByPrefix('idle', 'halloweem bg0');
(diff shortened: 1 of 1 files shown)


TentaRJ (commented):
Seems good! Have you tested this out to check to see if it works?

VoltrexMaster:
> Seems good! Have you tested this out to check to see if it works?

yes

TentaRJ:
> > Seems good! Have you tested this out to check to see if it works?
> 
> yes

Alright I'll try it on my build in a little bit to see if it works as well!

lightbulblighter:
A switch statement rather than an if-else would result in zero to very minuscule performance improvements. We live in an era where such micro-optimizations will make a marginal improvement with the fast processors that we have. Nevertheless, this pull request sacrifices code readability. Switch statements are great for 1-2 line actions, not huge chunks of code. This would make it hard to see where a case ends and starts.

VoltrexMaster:
> A switch statement rather than an if-else would result in zero to very minuscule performance improvements. We live in an era where such micro-optimizations will make a marginal improvement with the fast processors that we have. Nevertheless, this pull request sacrifices code readability. Switch statements are great for 1-2 line actions, not huge chunks of code. This would make it hard to see where a case ends and starts.

In the case of readability, it doesn't make mich difference between the huge block/chains of `if .. else if`s, it wouldn't make it harder to see where the braces start and end since it's almost the same case if huge chunks of `if .. else if` chains, although about the performance, the difference is not so little and miniscule since it just has to iterate through all of the conditions while also casting the string to lowercase all the time, sure it wouldn't make that big of a difference, but it's at least some performance boost, nothing bad at the very end, but I don't disagree with your point, just a little bit performance boost won't do any harm, this might be a huge performance boost in some cases but not entirely, so yea; not a bad point at least.

lightbulblighter:
Usually when trying to give an application a "performance boost" or optimizing in general you will optimize the area where the bottleneck occurs. Switch statements are readable by only tabulation rather than braces. This area has no bottleneck and does not slow down the game at all. My point is that this seems like a useless pull request.

VoltrexMaster:
> Usually when trying to give an application a "performance boost" or optimizing in general you will optimize the area where the bottleneck occurs. Switch statements are readable by only tabulation rather than braces. This area has no bottleneck and does not slow down the game at all. My point is that this seems like a useless pull request.

Performance boost doesn't exactly imply resolving bottlenecks in some areas of the code, it mostly means to just optimize the code as much as possible even if it means to sacrifice readability in anyway, I get what you're trying to say, this replacement of `if .. else if` chains with just a `switch` statement isn't exactly a big performance boost but I don't see anything wrong with even the slightest performance boost, but I get your point.

TentaRJ (approved):
These changes work as normal. The main improvement is readability and not having to use an if/else if string. I approve this. A minor change, but a good one to implement.

PrincessMtH (changes requested):
most of the tabs seem to have accidentally been turned into double spaces.

VoltrexMaster:
> most of the tabs seem to have accidentally been turned into double spaces.

Fixed 👍

TentaRJ:
> most of the tabs seem to have accidentally been turned into double spaces.

Can't believe I didn't see that smh

ninjamuffin99:
oops i actually did this in the private development repo already! I do appreciate it tho hehehe

ninjamuffin99:
I'll merge for now anyways hehe

PR #677, opened 2021-03-30
Fix code smells

ALANVF:
The title says it all. Do note that I didn't try making any *major* code changes since those could probably be done in a separate PR. Oh yeah also some of my changes are similar to those in #640, but I forked the repo before it was merged so idk.

Semi-related question, but why does some code use Haxe 4 features normally and other code checks for Haxe 4 first? I don't even see why you would want to support Haxe 3.x at this point, since Haxe 4 has been out for quite a while and has a lot of nice features like arrow functions and such.

### DIFF
--- a/source/Alphabet.hx
+++ b/source/Alphabet.hx
@@ -97,9 +97,7 @@ class Alphabet extends FlxSpriteGroup
 				if (isBold)
 					letter.createBold(character);
 				else
-				{
 					letter.createLetter(character);
-				}
 
 				add(letter);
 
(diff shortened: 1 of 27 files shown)


ninjamuffin99:
"purplel" is the intentional spelling, I just put it exactly how it is from the adobe flash file

For Nintendo Switch SDK backend stuff, you need to use 3.x, and in November/October we were testing around with it with my friend's devkit. 4.x stuff is cool, but probably gonna be a big fuss to switch over back and fourth. And also I'm not smart enough to use too many of the new fancy features anyways

ALANVF:
Well if the backend stuff ever gets updated to 4.x, lmk and I'd be glad to help transition the code over to 4.x.

I would also recommend trying out some of the newer haxe features on https://try.haxe.org/ sometime when you get the chance, they're really quite useful imo.

FOr now, I suppose I'll convert my code changes to use 3.x features and then it should be fine to merge?

ALANVF:
Status of this?

PR #681, opened 2021-03-30
feat(format): Fix formatting

VoltrexKeyva:
These two file's formatting is scuffed, Mashallah.


KurtisCates:
Where is this file ever used in the code?

VoltrexMaster:
> Where is this file ever used in the code?

https://github.com/ninjamuffin99/Funkin/blob/master/source/Snd.hx#L3

Two-Ai:
Is Snd.hx ever used? It looks like both files were added in d1afc09 but were never used.
If neither file is going to be used, it would be cleaner to remove both files.

VoltrexMaster:
> Is Snd.hx ever used? It looks like both files were added in [d1afc09](https://github.com/ninjamuffin99/Funkin/commit/d1afc09577ec1277d93f5db3081d7a9bc8b2baa6) but were never used.
> If neither file is going to be used, it would be cleaner to remove both files.

Seems like the [Snd.hx](https://github.com/ninjamuffin99/Funkin/blob/master/source/Snd.hx) file is not used literally anywhere at all, maybe its just a placeholder or a pre-state for an upcoming feature?

The-SGPT:
based on the commits it was added for switch support

PR #684, opened 2021-03-30
Update freeplaySonglist to the newer freeplay format

rifxii:
This PR updates freeplay to load the tutorial and week one songs from a json similar to how it was done before [this commit](https://github.com/ninjamuffin99/Funkin/commit/5632736ed462a0f108250dcbbc50da66549c9186). 

Due to this commit freeplaySonglist doesn't work properly and loads every song described in that in that file as a week one song with gf's icon. This PR intends to fix that by making freeplaySonglist a json with each song inside of it having it's own icon and weeknum values. 

(wish i could've used tjson for the fancy formatting but unfortunately that doesn't work on HTML 😔)

### DIFF
--- a/assets/preload/data/freeplayInit.json
+++ b/assets/preload/data/freeplayInit.json
@@ -1,0 +1,1 @@
+{"init": [["Tutorial", 0, "gf"],["Bopeebo", 1, "dad"],["Fresh", 1, "dad"],["Dadbattle", 1, "dad"]]}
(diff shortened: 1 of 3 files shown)


aflacc:
wait what does this do really

rifxii:
> 
> 
> wait what does this do really

In the newer releases (non-github) of the game the game has a txt named freeplaySonglist which loads the tutorial and first week songs into freeplay which I'm gonna be completely honest I have no idea why this is a feature.

But with all the new github stuff this was changed to only load the tutorial song because of what I assume is because it's harder to import songs from there due to it needing an icon and week num so I just fixed it so you can load those in now with their proper icon / weeknum.

PR #701, opened 2021-04-03
Add proper donation screen

Leo40Git:
This lil' PR simply adds a donation screen. Nothing special here.

Only thing that needs to be done is adding "Give a Lil' Bit Back" and playing it here, and also maybe finding a way to open a link without opening an ugly command prompt on Windows.


CyndaquilDAC:
this is a nice little pr. my only recommendation is adding "give a little bit back" as the bg music since it kinda fits, though that's up to ninjamuffin since there might be some secret, unannounced use for it.

PR #702, opened 2021-04-03
bunch of fps stuff

itskwartz:
you can now easily switch between framerates (30fps, 60fps, 120fps and 240fps) through the main menu using the numpad

so i edited certain frame-based values for every fps locked thing i could find to compensate for the extra / loss of frames

yeah that's it, really

edit: if this gets merged and put into another state, the mainmenustate.hx camera movement code can go back into create()


actuallyasmartname (approved):
i like this

👏

PR #703, opened 2021-04-04
Loading gf-christmas as a character will no longer cause the game to crash

CyndaquilDAC:
adds gf-christmas to healthicon.hx so the game doesn't crash when you attempt to load her.


CyndaquilDAC:
(sorry for it using my org, i have a kade engine fork on this account so i couldn't refork the og. github why havent you allowed multiple forks when one of the forks is a fork of a fork AHHHHHHHHHHHHHHHHHHHHHHHHH that would be so useful)

aflacc:
I mean, we could just remove her. She doesn't exactly have a purpose outside of decoration

VMGuy23:
> I mean, we could just remove her. She doesn't exactly have a purpose outside of decoration

Yeah true but the decoration counts. Preventing game crashes when people try to experiment is a good thing

CyndaquilDAC:
has been fixed in the week 7 release

Issue #279, opened 2021-01-25
Bug Report: Polymod doesn't use binary assets in mods


regulad:
# Please check for duplicates or similar issues, as well performing simple troubleshooting steps (such as clearing cookies, clearing AppData, trying another browser) before submitting an issue.

### Web
- [ ] [Newgrounds](https://www.newgrounds.com/portal/view/770371)
- [ ] [Itch.io](https://ninja-muffin24.itch.io/funkin)

#### If you are using a web build, what browser are you using?
- [ ] Chrome, or a Chromium based broswer (Edge, Opera, etc.)
- [ ] Firefox, or a Gecko based broswer
- [ ] Safari
- [ ] Other, please specify:

### Local
- [ ] Windows x86
- [x] Windows x86_64
- [ ] Linux
- [ ] macOS

## What version of the game are you using? Look in the bottom left corner of the main menu.
v0.2.6

## Have you identified any steps to reproduce the bug? If so, please describe them below. Use images if possible.
1. Copy a mod into the `assets/mods` folder that changes binary assets, like `.ogg` or `.png`
1. Declare the mod in `modList.txt`
1. Launch the game

## Please describe your issue. Provide extensive detail and images if possible.
Only text files, like `.txt`, `.json`, and `.xml` can currently be changed by Polymod, not binary files like `.ogg` or `.png`.

AyeTSG:
Is anyone aware why this happens? If we can fix this, this would make the polymod modding system fully complete, no?

AyeTSG:
I think I've found out the problem. Polymod hooks into the lime ``Assets`` class, but most of the images don't use that, rather they use ``FlxSprite`` and ``FlxAtlasFrames``. This may instead be a polymod issue.

regulad:
> 
> 
> I think I've found out the problem. Polymod hooks into the lime `Assets` class, but most of the images don't use that, rather they use `FlxSprite` and `FlxAtlasFrames`. Sounds also aren't replaced as they're being loaded with `FlxG.sound.playMusic`. This may instead be a polymod issue.

If it's not too much of a pain, could you consider making a pull request?

AyeTSG:
I would, but I don't know enough about Haxe to fix this on FNF or Polymod's end...

PR #710, opened 2021-04-05
Fix PolyMod Binary Integration

Ji-Rath:
---
name: Bug Fix
about: Fix poly mod not replacing binary files
title: 'Bug Fix: Poly Mod Binary Integration'
labels: bug, polymod, modding, binary
---
## Does this PR close any issue(s)? If so, link them below.
#279 
Apologies if this pull request is not appropriate as **this fix is not completely finished**. I simply wanted to get the ball rolling on this issue since I did not see relevant solutions to this problem.
This essentially addresses some of the issues with the current poly mod integration. Currently, this fixes replacing binary files through the mods folder, as it did nothing before. While I was only successful with replacing images/animation on Windows 10 build, it should be possible to replace sounds/music with a few tweaks.
On a side note, the current implementation adds ALL mods in the mods folder and each mod requires a _polymod_meta.json file. In addition, I modified the name of the base folder to 'contents' as keeping it as 'assets' would not let me replace binary files. This should be looked into further.


Ji-Rath:
After further investigation, it looks like I stumbled upon the issue of why poly mod was not replacing files correctly. It seems that FNF uses Asset Libraries. I essentially cheated on 4240103 by modifying the Project.xml file to have: `<assets path="assets" rename="contents"/>` which creates a set of all the assets onto contents without using libraries. I then modified Paths.cs and removed the library references. ex. "'$library:contents/$library/$file'" to 'contents/$library/$file'.
This is further enforced by a [comment](https://github.com/larsiusprime/polymod/issues/21#issuecomment-813807785) by larsiusprime, the author of polymod:
> As it happens ninjamuffin is an old acquaintance of mine and I've been helping him fix some bugs with Polymod, did some digging on that tonight in fact. As it turns out he's doing some stuff with asset libraries I wasn't expecting, so it's going to take some refactoring to make FNF compatible with Polymod. I know now what needs to be done but it might take a while.

PR #733, opened 2021-04-09
chore(source): Cleanup

VoltrexKeyva:
Cleaned up [ChartParser.hx](https://github.com/ninjamuffin99/Funkin/blob/master/source/ChartParser.hx) and [line 18](https://github.com/ninjamuffin99/Funkin/blob/master/source/ChartParser.hx#L18) in this file doesn't do anything since it doesn't mutate the original string.


Two-Ai:
ChartParser is unused, it is only referenced by this commented out line: https://github.com/ninjamuffin99/Funkin/blob/6cf4523/source/PlayState.hx#L1040
(The current chart parsing code is in `Song.hx`.)
If ChartParser is not used in v0.2.8.0 (in the private dev repo), it would be simpler to delete the file and the commented out line.

Otherwise the changes seem good, except for style change of the 1-line if statements. (I prefer the full-brackets style, it takes more space but is clear and less error-prone.)

VoltrexMaster:
> ChartParser is unused, it is only referenced by this commented out line: https://github.com/ninjamuffin99/Funkin/blob/6cf4523/source/PlayState.hx#L1040
> (The current chart parsing code is in `Song.hx`.)
> If ChartParser is not used in v0.2.8.0 (in the private dev repo), it would be simpler to delete the file and the commented out line.
> 
> Otherwise the changes seem good, except for style change of the 1-line if statements. (I prefer the full-brackets style, it takes more space but is clear and less error-prone.)

I think this file is deprecated and they forgot to delete it, no idea lol.

PR #735, opened 2021-04-09
fix(source): Don't allow score to go to negatives

### LINKED ISSUE #728: petition
petition to eliminate the negative points

is very anoying that you complete a week but with negative points

the week record will be 0 points

Captura de pantalla (134)

VoltrexKeyva:
When missing a note, your current song score gets subtracted by `10` without any checks which can lead to negative scores, this PR fixes this issue and closes #728.


source/PlayState.hx
```diff
@@ -2142,7 +2142,7 @@ class PlayState extends MusicBeatState
 			}
 			combo = 0;
 
-			songScore -= 10;
+			songScore -= (songScore - 10) >= 0 ? 10 : songScore;
```
Two-Ai, line 2145: Keeping the zero lower bound code separate from the score modification would be a bit simpler and easier to read.
```suggestion
			songScore -= 10;
			if (songScore < 0)
			{
				songScore = 0;
			}
```

VoltrexMaster: Tbh, the source code uses the ternary operator in many places so not really that hard to read, the thing with this is that there's a rare chance where this could still go to negatives if you `update()` function is called before that if statement is passed.

Two-Ai: Oh ouch, that would be a nasty concurrency bug.
This code is only ever run from within `update()` so I assume it should be safe. (Specifically, `update > keyShit > (noteCheck >) badNoteCheck > noteMiss`.)
I don't know for sure if HaxeFlixel is single-threaded (the original Flixel was written in ActionScript 3 which is single-threaded), but either way calling `update()` before the previous `update()` call completes seems like a recipe for total chaos.

VoltrexMaster: yea, this might be an issue later on if they add anything to call it multiple times, so thanks for suggestion anyway.

aflacc:
honestly I see no problem with this. I don't get how you can finish a level with negative score without dying.

VoltrexMaster:
> honestly I see no problem with this. I don't get how you can finish a level with negative score without dying.

The point is not finishing the level with negative score, this PR is meant to fix the negative score during the runtime of the level.

aflacc:
> > honestly I see no problem with this. I don't get how you can finish a level with negative score without dying.
> 
> The point is not finishing the level with negative score, this PR is meant to fix the negative score during the runtime of the level.

yeah but theres no reason to not have this you get out of the negatives very quickly by hitting like 1-2 notes

VoltrexMaster:
> > > honestly I see no problem with this. I don't get how you can finish a level with negative score without dying.
> > 
> > 
> > The point is not finishing the level with negative score, this PR is meant to fix the negative score during the runtime of the level.
> 
> yeah but theres no reason to not have this you get out of the negatives very quickly by hitting like 1-2 notes

A bug-like issue doesn't imply how fast you can get rid of it, it is still an issue at the end.

actuallyasmartname (approved):
looks good

Salladk-but-no-mo-flaged:
tmy problem is the game make hard to get a high score with negative score

also bad make you lose many points (im not very sure but i sus like 350 points)

PoketrainerMS:
dont know who needs to know this, but the lowest you can go without dying is -240. one more miss and you die.

PR #750, opened 2021-04-13
Added support for h, j, k, l bindings

flber:
This should just add the ability to h, j, k, and l for the movement. I find that this makes a bit more sense to me, as the in-game arrows are displayed in a linear format similar to these keys. Anyway, feel free to reject if you don't think adding more keybindings is a good idea, and thank you for making such a good game!

### DIFF
--- a/source/Controls.hx
+++ b/source/Controls.hx
@@ -498,10 +498,10 @@ class Controls extends FlxActionSet
 		switch (scheme)
 		{
 			case Solo:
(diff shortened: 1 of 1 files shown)


hdboye:
Rebindable keys are already being worked on and will release in the next update.

ecksemzee:
This just replaces WASD keybinds

nickpythonluajava:
its cool but its gonna be in da next update

PR #528, opened 2021-03-03
unseparate characters from their weeks

ActualMandM:
makes the game no longer crash if you use a character not from the song's specified week

this is a necessary change for custom songs (ex. [Dad Battle (In-Game Version)](https://youtu.be/2d0U8RHH3-E))

https://user-images.githubusercontent.com/15317421/109759423-d517f400-7ba1-11eb-92c3-96b3832af841.mp4

PR #767, opened 2021-04-16
unhardcode stages from songs + additional changes for custom songs

ActualMandM:
this pr unhardcodes the stages from their songs in the code so songs can now define a stage of their choice in their chart file

this requires a major modification to the asset loading system: basically instead of weeks having their own libraries, stages have their own libraries

chartingstate allows you to change the stage and will go through loadingstate first instead of going directly to playstate. to make sure that the game doesn't crash on web from leaving chartingstate after changing the stage, we check to make sure if the library for said stage is loaded first

due to how loadingstate reads from a lowercase input, mallEvil and schoolEvil have the -evil suffix instead

to make sure the game doesn't crash from custom charts not having a stage variable, a fallback is included so that the correct stage will be loaded for older custom charts (this can be removed once enough time has passed)

additionally, custom songs can now be added to freeplay via customSonglist and will appear at the bottom with bf-old. because of this, tutorial is added to the freeplay menu via the method that every song built into the game uses

additional changes:
 - #528 is automatically a part of this pr as grouping the characters per stage would be a dumb idea
 - pixel boyfriend's game over is now applied to his character instead of being restricted to the school stages
 - girlfriend will always draw behind the limo even if she is chosen as player 2
 - gf-christmas will no longer crash the game and will properly take over gf if she is chosen as player 2
 - girlfriend versions are now connected to boyfriend versions instead of being restricted to their respective stages
   - remind me to allow this be changed in the chart file
 - browser will always transition in playstate now so that it smoothly enters and exits from loadingstate
 - some additional code optimizations/organizations have been made (ex. we don't set gf-car twice if limo is the current stage)
 - most importantly, the stage switch case is no longer poorly formatted

showcase:

- changing stage via chartingstate

https://user-images.githubusercontent.com/15317421/114972055-8c0abf00-9e32-11eb-96cc-14de1e163e13.mp4

- loading stage via chart file or resorting to fallback (using modified [SiIva Pico rip](https://youtu.be/vh6jf5d-ut4) as a test)

https://user-images.githubusercontent.com/15317421/114972112-b3618c00-9e32-11eb-93bb-ef58ba23c138.mp4

- custom song in freeplay via customSonglist

2021-04-16_15-13-16_Funkin


hdboye (approved):
this looks like a great addition!

CyndaquilDAC (approved):
Swaggy

The-SGPT (approved):
real good addition; would help with customssongs no longer needing to have to be an existing song to get that stage

ActualMandM:
Will be refactoring once 0.2.8 code is pushed.

ActualMandM:
Converted to draft until week 7 source code is out and I can update the PR to that

ActualMandM:
just realized that you cna restore branches :letsfuckinggo:

polybiusproxy:
> just realized that you cna restore branches :letsfuckinggo:

lol

PR #776, opened 2021-04-17
week 6 texture for story mode fix

trayfellow:
there was a small issue with the texture for week 6's logo in story mode so i fixed it i guess

im not sure how pr's work since this is my first time trying this but all it does is replace the current week6 texture file with the one i uploaded


hdboye (approved):
the white dots at the top of the week 6 text are so frustrating, lgtm

trayfellow:
isseu has been fixed in the week 7 update

PR #860, opened 2021-04-22
OCD fix

Th3SnowyOwl:
These pixels were triggering me, might update this more later (
Screenshot 2021-04-21 10 32 48 PM Annoying.


Th3SnowyOwl:
image
The xml file is slightly incorrect, leading to the down arrow having one column of pixels from the right arrow.

Th3SnowyOwl:
And some other ocd triggering fixes

PR #874, opened 2021-04-22
OCD fix

Th3SnowyOwl:
These pixels were triggering me, might update this more later.Screenshot 2021-04-21 10 32 48 PM
The xml file is slightly incorrect, leading to the down arrow having one column of pixels from the right arrow.
image


Th3SnowyOwl:
im assuming this may have been fixed with week 7?

trayfellow:
> im assuming this may have been fixed with week 7?

yea its fixed now

PR #897, opened 2021-04-23
Stopped old bf from toggling off

AUserOnGitHub:
When you change your health icon to the old bf icon it would revert back when going into another song (even if it is in the same week). This fixes that issue by carrying a variable over to the next song.


actuallyasmartname:
Is this really needed?

AUserOnGitHub:
Not specifically but it would make the hidden Easter egg look more professional.

Issue #932, opened 2021-04-25
Bug Report: Not every track on a week is getting displayed.


polybiusproxy:
### If you are playing the game in a browser, what site are you playing it from?
[Newgrounds](https://www.newgrounds.com/portal/view/770371) or [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below.

### If you are playing the game in a browser, what browser are you using?
Chrome (or a Chromium based browser, i.e. Edge, Opera, etc.), Firefox, or Safari? Specify b**e**low.

### If you are playing a downloaded version of the game, what operating system are you using?
Windows (`x86`), Windows (`x86_64`), Linux, or macOS? Specify below.

Windows (64 bits).

## What version of the game are you using? Look in the bottom left corner of the main menu.

0.2.7.1

## Have you identified any steps to reproduce the bug? If so, please describe them below. Use images if possible.

1. Download source code and do "lime test windows" to compile it.
2. Once game is running, just go on Story Mode.

## Please describe your issue. Provide extensive detail and images if possible.

In Story Mode, not every track shows.

Example: On week 1, it just shows "Bopeebo" and "Fresh" on the Tracks section. WHERE IS DADBATTLE AHHH-
![image](https://user-images.githubusercontent.com/47796739/115997954-5493d380-a5e5-11eb-9351-87f0f8a81cd3.png)
![image](https://user-images.githubusercontent.com/47796739/115997957-5c537800-a5e5-11eb-9128-62f372158ff4.png)

KitOdin:
simple... ish, first you duplicate the storyplaylist to another array, you'll see why in a moment, also be sure to rename the duplicated array to some like "songNames" or "weekTracks" to avoid compiler confusion, then at the end of each tracklist(the arrays IN the array) at a (, " Reserved ") without the rounded brackets, you can also set the "Reserved" text to anything including nothing, the reason it needs a duplicated array is that storyPlaylist(containing ALL of the tracks witch is used by the track list text to render all of weeks songs), is needed in order for a week to be playable, if you change one of the songs in a tracklist, it would apply to gameplay, so adding a extra space to a tracklist without cloning the storyPlaylist array would cause the game add that to the song list, and when its time to load it up, well it basically crashes.

PR #940, opened 2021-04-26
turn the week's final track visible

kyandora:
Solves #932 (already fixed in v0.2.8). Looks like every track needs to end on a line break to be visible.


rifxii:
This is only a problem if you're using the latest version of OpenFL (9.1.0) from what I've looked at, if you use `haxelib git openfl https://github.com/openfl/openfl` in the same way to how you install stuff like discord_rpc in your terminal this problem should be fixed.

notdonxd:
thanks for this fix! lol
i suggest this being added to readme.md so it doesnt happens anymore .-.

notdonxd (approved):

PR #954, opened 2021-04-27
Difficulty-Specific Vocals Demo

T0paz:
Each difficulty has different charts. Instead of redoing vocal tracks to match the charts, the in-game engine can selectively mute the vocals to match it to the chart. This provides backwards compatibility and mod compatibility. This is a demonstration, but it could see better implementation if the hard mode chart were always loaded, so note lengths could be adjusted.
Uses fadeOut() to asynchronously silence the vocal track after 0.2 seconds whenever a good hit is made by Boyfriend or the opponent (before the note is destroyed).
Known issues:
- fadeOut() quiets the track, making it hard to hear alongside the music. This is exacerbated if the chart does not perfectly align with the vocal track.
- Some notes that are very close in Hard mode can be heard on certain notes.
- Certain notes sound cut-off or held notes replace the fast inputs of Hard mode, so the representation is not perfect.


doggy-dentures:
Instead of:
`vocals.fadeOut(0.2);`
You can use:
`vocals.fadeOut(Conductor.stepCrochet/1000, 1, function(_){vocals.volume = 0;});`.
It's not the prettiest, but it does two things:
1. Rather than having vocals play for an arbitrary amount of time (0.2 seconds), they play for the length of...an eighth note, I think? Either way, the timing is now dependent on the song's tempo.
2. Vocals don't progressively fade out into silence, but instead play at full volume until they need to be muted.

Even with the above change, the system remains a bit flawed.
Still, I think this is a rather interesting idea. Maybe it can added as an option in the Options Menu or something.
Here are some recordings for all difficulty levels in Thorns with this implementation:
[Thorns Hard](https://www.youtube.com/watch?v=HaSFyRThIaQ)
[Thorns Normal](https://www.youtube.com/watch?v=ZFx0LbmWglk)
[Thorns Easy](https://www.youtube.com/watch?v=xO2FY24DWNU)

T0paz:
Thanks for the function parameters, I was having a hard time figuring out how to get the fadeOut() tween to play at full volume until the duration ends. The demo videos also look great and help demonstrate the modification.
Now that the system is working mostly correct, I'm thinking that the note duration could be adjusted to be longer, based on whether another note exists in a time frame only in the Hard mode chart (essentially used as a "perfect" map of the vocals themselves). This would require loading both charts at once somehow, but I think it would help fix the choppiness of the muting (though Thorns will inevitably sound the same because of its design).

CyndaquilDAC:
Is it alright if I make this an option in Kade Engine as a PR?

PR #1011, opened 2021-05-02
Fix update prompt

VMGuy23:
Fixes the update prompt on self-compiled versions


OPNA2608:
Duplicate of #941.

VMGuy23:
OK, I'll close this

PR #1019, opened 2021-05-04
Tween Ins Tween Outs and Linear Tweens

ILuvGemz:
I made all of the eases with the modchart effects to provide more capabilities for the notitg modding scene.


Kiffolisk:
me when you literally import kade engine code

PR #1020, opened 2021-05-04
newer changelog

Kiffolisk:
l


rifxii (changes requested):
You should probably mention the addition of the options menu (customizable controls). input rewrite by MtH, and replace "Cute little icons onto freeplay menu" with something along the lines of "Prettied up the freeplay menu" as there's also the background changing colors per week.

hdboye:
this will probs be added when the source code will be released.. nothin needs to be written by contributors

ActualMandM:
This is already changed in the week 7 assets.
2021-05-05_10-55-32_notepad++

Issue #1021, opened 2021-05-04
Bug Report: My game poster black boxes instead of boyfriend and girlfriend 


no-us:
### If you are playing the game in a browser, what site are you playing it from?
[Newgrounds](https://www.newgrounds.com/portal/view/770371) or [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below.

### If you are playing the game in a browser, what browser are you using?
Chrome (or a Chromium based browser, i.e. Edge, Opera, etc.), Firefox, or Safari? Specify blow.

### If you are playing a downloaded version of the game, what operating system are you using?
Windows (`x86`), Windows (`x86_64`), Linux, or macOS? Specify below.

VMGuy23:
Memory issue, not game bug.

no-us:
> Memory issue, not game bug.

how to fix the bug and i play on newgrounds in qwant

VMGuy23:
You cant fix it, you dont have enough ram, only way to fix it is to get more ram

VMGuy23:
please close its not a bug

NotHyper-474:
There's the possibility that the game just loaded incorrectly, try clearing your browser's cache and see if it works

Issue #992, opened 2021-05-01
Bug Report: Some of the sprites and backgrounds appear as black boxes


KaidenPhillips:
The GF and BF sprites appear as black boxes (except in week 6 and the the week 7 cutscenes)  and as far as I know the spookies background appears as a black box and there's probably more I haven't looked at yet

VMGuy23:
Where? What browser?

KaidenPhillips:
Brave, and on Newgrounds

VMGuy23:
Tried Chrome? Edge? I have this issue but only on Chrome for Android in Desktop Site.

KaidenPhillips:
I have tried Chrome and I'd rather die than use Edge, I have a Chromebook which is a hybrid yet still mainly an Android

VMGuy23:
Try Chrome, then try Edge (I mean, its actually good now)

The-SGPT:
Memory issue caused by browsers, wont get fixed because thats just how it is

KaidenPhillips:
Well shit

VMGuy23:
Thats hy it happened on my 3GB of RAM Android tablet and not my 16GB RAM laptop or my freinds iPad

Issue #856, opened 2021-04-22
Bug Report: Unusually High Memory Usage When Loading Scenes


T0paz:
**Playing from:**
[Newgrounds](https://www.newgrounds.com/portal/view/770371)

**Browser:**
Firefox (Linux)

**Version:**
v0.2.8 (Newgrounds exclusive preview)

**To Reproduce:**
Play through any week, or through several levels in Free Play. Keep a resource monitor visible, such as through your OS's task manager. When the game switches scenes, the memory usage will dip, then increase to a higher value before the previous scene was unloaded. This also applies to every time a song is restarted. I found this most noticeable when playing through Week 5.

**Description:**
When playing FNF in browser, the memory usage can increase rapidly as scenes are reloaded either through normal play or having to retry a song. Looking at a resource manager, it's clear that the largest spikes in memory usage are a result of the game loading in new scenes. Even worse, needing to retry a song adds to the memory usage rapidly, and gigabytes of memory can be consumed in seconds simply by restarting a song over and over.

I checked previous issues regarding the game's memory usage, but this seems much more severe than what I've experienced with previous versions of the game (downloaded, Windows and Linux), and the cause of widespread "black box" or missing asset issues. My computer has 16 GB of memory, and for it to run out so quickly is typically a sign of a memory leak or irresponsible multitasking.

**Images:**
These are screenshots of my task manager monitoring memory usage. The top of the graph represents the 16 GB of memory my computer can manage.
[Week 5 (Christmas)](https://cdn.discordapp.com/attachments/459406668690620416/834602736443850773/unknown.png)
The memory usage maps pretty clearly to the loading screens and songs, the latter of which are represented by the plateaus.

[Restarting Week 5 (Cocoa) Rapidly](https://cdn.discordapp.com/attachments/459406668690620416/834611420515532830/unknown.png)
Memory usage increases very quickly (same time scale as the first graph), and can cause your computer to run out of memory in seconds. I quickly closed the tab running the game before it slowed my computer way down.

Issue #280, opened 2021-01-25
Memory useage problems


aiden3c:
While playing the game, memory it uses constantly goes up. After maybe 20-30 minutes of play and I go to task manager and see 10 gigs being used. 
This is either from reloading assets that are already loaded, or a memory leak of some kind.
This is definitely something that should be fixed asap, even when paused my memory usage went up about a gig every minute or two.
It seems to happen on all web versions of the game, but I assume it'll be in any build of the game.

PR #1083, opened 2021-05-13
Fixed major memory usage issues

aiden3c:
When assets are loaded into the game, they are never unloaded. This can cause many issues related to significant RAM usage. Most issues that complain about characters loading as black boxes are from running out of memory. This also causes slowdowns as the system runs out of RAM. I found these issues myself playing the game normally for the first time, and it seems like a very critical issue. 
I'll note that in the cases where assets are loading as black boxes, that's more from devices with 4 gigs of RAM or so, so it may not be able to load the assets initially at all when loading the game. Because of that, this request may not fix all instances of black boxes from happening, but I think it'll help/prevent it in a lot of cases because it'll keep memory as low as it can.

Anyways, my solution overwrites the Haxe `add` command, and adds the object you're adding to an array of of tracked assets, then it loads the asset in as it would normally. Then in key areas (like when a song gets finished, or when going from a menu to a song), it'll unload the assets using the Haxe `remove` command. There are a few areas I manually unloaded some assets, like the GameOverSubstate, just because there's only one or two assets ever loaded.

TL;DR: This cleanly and dynamically keeps track of and removes unneeded assets to stop clogging up memory. This fixes some pretty significant issues.

(Memory usage outright)
#856 #280 #58 

(Black boxes/Characters just not loading)
#1021 #992 (There are more I just don't want to search through it all lol)

Let me know if you have any questions or comments! Pretty happy with how clean the implementation is.


PurpleButNotALight (approved):
we must support the low end gamers, ninjamuffin pls merge.

nickpythonluajava:
raspi 4: **My time has come**

nickpythonluajava (approved):

hdboye (approved):
ninja, if u see this, merge or ur a poopy fart

nickpythonluajava:
> ninja, if u see this, merge or ur a poopy fart

insult of the year

hdboye:
i know im a baddie

VMGuy23:
> raspi 4: **My time has come**

I have a Raspberry Pi 4 (4GB RAM) and it would be cool to see the FPS on that thing, and play the game without squares

trayfellow:
> ninja, if u see this, merge or ur a poopy fart

hdboye when ninjamuffin doesn't merge this : https://youtu.be/EEuFbmmuqK4

VMGuy23 (approved):
Nice

I need this

trayfellow (approved):

NotHyper-474 (approved):

Rs232k:
Sad, maybe it's works in desktop, but in mobile created crash.

Naishaline:
> Sad, maybe it's works in desktop, but in mobile created crash.

How did you go about compiling for mobile?

Hendicted:
what about the songs that get loaded in? from what i'm seeing they dont seem to be unloading

Steviegt6 (changes requested):
There appears to be quite a lot of boiler-plate code used here. Would a single class that serves as an asset manager not work better? I see no reason for using what appear to be the same methods over and over in a few different classes.

Edit: Nevermind, missed the fact that you're overriding methods. My bad.

Hendicted:
actually since the game uses openfl libraries you could try to use [https://api.openfl.org/openfl/utils/Assets.html](the unloadLibrary() function) of openfl.utils.Assets. i could make a proof of concept out of it

Steviegt6:
Frankly, if you were to make a PoC then you may as well make an alternative PR with similar goals.

notdonxd (changes requested):
You should add this function in other states, like the Pause menu.

Hendicted:
> Frankly, if you were to make a PoC then you may as well make an alternative PR with similar goals.

i tried it! but unfourtanetly the results were dissapointing

basically what the function actually does is unload the library and it's paths, not unload the library from memory, which i find dissapointing

Michael82548:
#1222

MemeHoovy (approved):

TheGaloXx (approved):
THIS IS SO GOOD, IM TRYING THIS OUT

APRO716 (approved):

PR #1150, opened 2021-05-21
Added Override Update Check

TechMaster85:
Modders don't want to have to deal with players getting links to the vanilla game whenever the mod is based on an older version. This pull request adds a toggle to let modders choose whether or not they want an update screen, and if enabled, new versions are completely ignored.

Update: I want to make this the foundation for a potential in-game setting to disable update notifications, not just a toggle exclusively for modders as I originally intended.

### DIFF
--- a/source/TitleState.hx
+++ b/source/TitleState.hx
@@ -32,6 +32,9 @@ using StringTools;
 
 class TitleState extends MusicBeatState
 {
+	static var overrideUpdateCheck:Bool = false;
+	// ^^ For modders that don't want players do deal with an update notification all the damn time
+
 	static var initialized:Bool = false;
 
 	var blackScreen:FlxSprite;
@@ -292,7 +295,7 @@ class TitleState extends MusicBeatState
 
 				var version:String = "v" + Application.current.meta.get('version');
 
-				if (version.trim() != NGio.GAME_VER_NUMS.trim() && !OutdatedSubState.leftState)
+				if (!overrideUpdateCheck && version.trim() != NGio.GAME_VER_NUMS.trim() && !OutdatedSubState.leftState)
 				{
 					FlxG.switchState(new OutdatedSubState());
 					trace('OLD VERSION!');


Steviegt6:
Modders can just add this themselves? Pretty redundant.

TechMaster85:
> Modders can just add this themselves? Pretty redundant.

I know modders can add it but it would be nice if there was a universal way to disable these notifications. Others (or me if I'm up to it) can potentially extend this by implementing an in-game setting so notifications can be toggled without recompiling. That way it'll be useful to others than just modders. I now see this as more of a foundation than a complete feature.

Steviegt6:
> > Modders can just add this themselves? Pretty redundant.
> 
> I know modders can add it but it would be nice if there was a universal way to disable these notifications. Others (or me if I'm up to it) can potentially extend this by implementing an in-game setting so notifications can be toggled without recompiling. That way it'll be useful to others than just modders. I now see this as more of a foundation than a complete feature.

And, frankly, I don't see it as a complete feature or foundation for an idea. That stuff is pretty specific to modders and it should be their job to implement extra checks are stuff. An outdated repository has no reason to cater to mods in this context.

TechMaster85:
> > > Modders can just add this themselves? Pretty redundant.
> > 
> > 
> > I know modders can add it but it would be nice if there was a universal way to disable these notifications. Others (or me if I'm up to it) can potentially extend this by implementing an in-game setting so notifications can be toggled without recompiling. That way it'll be useful to others than just modders. I now see this as more of a foundation than a complete feature.
> 
> And, frankly, I don't see it as a complete feature or foundation for an idea. That stuff is pretty specific to modders and it should be their job to implement extra checks are stuff. An outdated repository has no reason to cater to mods in this context.

Now that I've thought about it, I completely agree with you that if this is catered to only modders a commit like this is pointless. Do you think that if I create an in-game setting that does the same thing I'll have a good chance of people finding it useful?

I'm sorry as I'm really new to this stuff. Thank you so much for your helpful feedback.

Steviegt6:
> > > > Modders can just add this themselves? Pretty redundant.
> > > 
> > > 
> > > I know modders can add it but it would be nice if there was a universal way to disable these notifications. Others (or me if I'm up to it) can potentially extend this by implementing an in-game setting so notifications can be toggled without recompiling. That way it'll be useful to others than just modders. I now see this as more of a foundation than a complete feature.
> > 
> > 
> > And, frankly, I don't see it as a complete feature or foundation for an idea. That stuff is pretty specific to modders and it should be their job to implement extra checks are stuff. An outdated repository has no reason to cater to mods in this context.
> 
> Now that I've thought about it, I completely agree with you that if this is catered to only modders a commit like this is pointless. Do you think that if I create an in-game setting that does the same thing I'll have a good chance of people finding it useful?
> 
> I'm sorry as I'm really new to this stuff. Thank you so much for your helpful feedback.

I don't find the setting useful purely because the lack of an up-to-date repository seems to be a unique case for week 7 as of now. If ninjamuffin were to merge a PR that adds the option then I wouldn't be opposed to it, though.

TechMaster85:
Ok then, when the Week 7 branch becomes public I'll consider another PR with a new setting. Thank you for your suggestions.

PR #1227, opened 2021-06-03
Adds the ability to repeat songs and AB Repeat.

Injourn:
The loop functionality is too long to show here, but the AB repeat looks like this:

https://user-images.githubusercontent.com/24728334/120572823-79901900-c3d1-11eb-8fc9-28155d68c233.mp4

Useful for practicing difficult sections in certain songs.

It uses the pause menu for options.

image


ActualMandM:
is it also possible to disable score incrementation while it's looping?

Injourn:
> is it also possible to disable score incrementation while it's looping?

Yeah it is, I had that planned from the start, but I forgot to actually implement it. I'll add it to the pr later today.

PR #1255, opened 2021-06-10
Allow providing Newgrounds API info via compiler flags

TheEssem:
This PR provides a way to specify a Newgrounds application ID and encryption key on compile-time via the command line instead of hardcoding them. Example:
```sh
$ lime test html5 -debug -Dng_api="application ID" -Dng_enc="encryption key"
```
If these arguments aren't provided, the variables will fall back to empty strings. This removes the need to recreate the APIStuff.hx file on every clone and can possibly help a little bit with security (e.g. preventing accidental commits containing the keys).


VMGuy23:
das good

polybiusproxy (changes requested):
you should also edit "build-NG-HTML-RELEASE.bat" in art folder.
just that, this pr seems good

PR #1331, opened 2021-07-03
fix week 6 week icon

ThePlank:
week 6's week icon has a couple of pixels on top of the image so i fixed them
image

I added a black background in editing so i can see where it is (not in the final edit)
this happened because of cropping to multiple files


ThePlank:
well now its fixed in 0.2.8.0 so im just gonna close

PR #1334, opened 2021-07-03
Make dad appear when pressing NINE

Cheemsandfriends:
So yeah
BF old without dad is kinda lame so How about adding dad?
(The code is kinda lame xd)


ACardboardBox1:
it's still dad tho ❓

Cheemsandfriends:
> it's still dad tho ❓

It is dad xd
but it's with the Icon grid thing.
I'll test some things for the separated icons and stuff but yeah xd

PR #1421, opened 2021-08-05
Make it actually show the creators names instead of just showing "PRESENT"

ACrazyTown:


VMGuy23:
It's already in the NG version

MythsList:
It shows but depending on your computer's performance it will skip that text.

ACrazyTown:
> It shows but depending on your computer's performance it will skip that text.

I'm pretty sure it has to do with the music being played before all the intro assets were created, this PR just plays the music after everything's been made and loaded so the text should show up on mostly all devices

cuckydev:
the actual reason this is an issue is because MusicBeatState sucks and skips steps, and if a step on a beat is skipped it also skips beats. stupid dum dumb

PrincessMtH:
this is fixed in week7 update already through a more robust method but thank you!

PR #1424, opened 2021-08-08
Describe what the txt file is for because why not

MisterAlexanderB:


josetzunun:
BISH

Steviegt6:
There are many `txt` files following the same naming format here, this isn't necessary.

ninjamuffin99:
aight

Steviegt6:
Quite an insignificant amount of playful and innocent tomfoolery has been committed.

PR #1467, opened 2021-08-22
fix the arrows pls i beg you

ultraivan12:
fix the arrow trails so that they do not overlap or separate


AbnormalPoof:
stop merging kade engine please ty
@ninjamuffin99 can you close this

PR #1690, opened 2021-08-26
Remove the extra haxeflixel link

theblobscp:
If you install Flixel using haxelib as the instructions say, you don't need to go to the link to install it again


AbnormalPoof:
Unnecessary since lime needs to be installed via the link, and the repo assumes you did that

theblobscp:
Actually, you can install lime using haxelib too, so there's no need for the extra link. I'm gonna add that too now.

AbnormalPoof (changes requested):
maybe add `haxelib run lime setup flixel` and `haxelib run lime setup` below `So for each of those type` `haxelib install [library]` `so shit like` `haxelib install newgrounds`

AbnormalPoof:
and maybe add openfl to the dependency list while you're at it lol

theblobscp:
When did I do this

PR #1696, opened 2021-08-28
Add the -easy and the -hard so you don't need to Rename

Cheemsandfriends:
Hello. So I wanted to make this since there is no way to save the things if you went from the difficulties. You can only rename if you want to make easy or hard.
this can be a solution to that problem lol

### DIFF
--- a/source/ChartingState.hx
+++ b/source/ChartingState.hx
@@ -1039,7 +1039,12 @@ class ChartingState extends MusicBeatState
 		var json = {
 			"song": _song
 		};
-
+		
+		var diffic:String = "";
+		if (PlayState.storyDifficulty == 0)
+			diffic = "-easy";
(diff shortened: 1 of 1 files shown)


ThePlank (approved):
cool, but we need a difficulty dropdown now

dum-du-m:
pls merge...

somefnfplayerlol (changes requested):
add a difficulty dropdown to the charting menu

ninjamuffin99:
the way different difficulty stuff is being handled is gonna be different in new update, so gon close dis for now since its kinda bein addressed hehehe

Cheemsandfriends:
> 
> 
> the way different difficulty stuff is being handled is gonna be different in new update, so gon close dis for now since its kinda bein addressed hehehe

I'll definitely gonna remaster my thing when the thing comes out lol

PR #1733, opened 2021-09-06
stage offset debug

G4merz:
changing stage offsets sucks, a stage positioning debug but for the actual stage, not the characters, would be really nice because STAGE OFFSETS S U C K S


Cheemsandfriends:
Can You please be more Specific? I don't want to eat 70 billion kade dev PR's lol

Cheemsandfriends:
Also. Why KE? this is Fukin master lol

PR #1758, opened 2021-09-17
POLYMOD UPDATED LOL FIX IT

EyeDaleHim:


EyeDaleHim:
saw this from a polymod PR
`https://github.com/larsiusprime/polymod/pull/39`

felt like it was better to change the link because the current one breaks fnf(?)

ninjamuffin99:
already updated in funkin development fork so imma close dis

PR #1760, opened 2021-09-17
Easter Eggo + AnimationDebug Tweaks.

KitOdin:
Added a little easter egg in the menu ;)
also tweaked the hidden ass animation debug menu only exclusive to debug build targets lmao.


Infinity48Games:
Does not work. It gives these errors on compile:
```
source/LatencyState.hx:19: characters 26-40 : Type not found : FlxGridOverlay
source/LatencyState.hx:66: characters 7-20 : Type not found : controls.BACK
source/AnimationDebug.hx:164: characters 7-22 : Type not found : controls.ACCEPT
```
To fix this just add:
```
private var controls(get, never):Controls;

	inline function get_controls():Controls
		return PlayerSettings.player1.controls;
```
To the beginning of both classes and add `import flixel.addons.display.FlxGridOverlay;` to the beginning of LatencyState.hx.

KitOdin:
fixed :)

source/AnimationDebug.hx
```diff
@@ -60,12 +59,11 @@ class AnimationDebug extends FlxState
 			add(bf);
 
 			char = bf;
-			bf.flipX = false;
 		}
 
 		dumbTexts = new FlxTypedGroup<FlxText>();
 		add(dumbTexts);
-
+
```
somefnfplayerlol, line 66: tf is this?

somefnfplayerlol (changes requested):

source/AnimationDebug.hx
KitOdin: dunno.

prokube: me when
``` diff
-
+                 
```

PR #1824, opened 2021-10-03
Make a Dockerfile

theblobscp:
Docker is cool


DotWith:
How does this work? I've never used Dockerfile.

theblobscp:
Fire ship has a good video on it, watch it and you’ll probably understand it 
Link: https://www.youtube.com/watch?v=Gjnup-PuquQ

DotWith:
> Fire ship has a good video on it, watch it and you’ll probably understand it Link: https://www.youtube.com/watch?v=Gjnup-PuquQ

This needs to be merged

theblobscp:
I'm still working on some bugs, but thanks for the feedback

AbnormalPoof:
why only linux

DotWith:
> why only linux (in the build command)

Docker is a Linux **based** thing

AbnormalPoof:
oops
what about html5

AbnormalPoof (changes requested):
maybe add this after linux idk

```
# Generate the built game for html5
CMD [ "haxelib", "run", "lime", "test", "html5" ]
```

.gitignore
```diff
@@ -1,3 +1,3 @@
 export/
 .vscode/
-APIStuff.hx
```
Steviegt6, line 3: If you're gonna do this, you should instead use environment variables, not a class containing two strings.

Steviegt6 (changes requested):
Using a docker for this is actually really dumb. Use a GitHub workflow for building. You can use encrypted environment variables to avoid leaking API keys (though they're easy to retrieve regardless due to source compilation & I believe Ninja is taking a different approach in the private fork).

theblobscp:
> 
> 
> maybe add this after linux idk
> 
> ```
> # Generate the built game for html5
> CMD [ "haxelib", "run", "lime", "test", "html5" ]
> ```

CMD can only be used once

.gitignore
theblobscp: Do you have a block of code I can add to it

theblobscp:
Also html would be useless, as it auto-runs it. Docker is for CLI, not GUI.

theblobscp:
I realized that for some reason the builds don’t work due to an error in the final build. I’m gonna close the pr until I can find a solution to this. If you need to test it, git pod is where I did it.

PR #1831, opened 2021-10-09
HashLink compatibility

libauror-a:
fuck neko, embrasse hashlink, just some problems, NGio dosen't work because you'll get this error `FATAL ERROR : Failed to load library ssl.hdll` so i had to jankly fix the error, (probally modders shoudn't use the NG shit in their mods/engines and shoud be coded out) probally gimme suggestions because this is kida bad, but it still works!

### DIFF
--- a/source/Discord.hx
+++ b/source/Discord.hx
@@ -7,7 +7,7 @@ using StringTools;
 
 class DiscordClient
 {
-	#if cpp
+	#if discord_rpc
 	public function new()
 	{
 		trace("Discord Client starting...");
--- a/Project.xml
+++ b/Project.xml
@@ -169,6 +169,6 @@
 
 	<!-- <haxedef name="SKIP_TO_PLAYSTATE" if="debug" /> -->
 	<haxedef name="NG_LOGIN" if="newgrounds" />
-	<haxedef name="newgrounds" if="newgrounds" />
+
 	
 </project>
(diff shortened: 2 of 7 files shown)


source/Discord.hx
```diff
@@ -84,4 +85,9 @@ class DiscordClient
 
 		//trace('Discord RPC Updated. Arguments: $details, $state, $smallImageKey, $hasStartTimestamp, $endTimestamp');
 	}
+	#else
+	public function new() {
+		trace("Discord Client isn't going to start because this is a cpp client, go home.");
```
ACrazyTown, line 90: looks good, but shouldn't it be "because this is *not* a cpp client"?

ACrazyTown (changes requested):

DotWith:
Also do the same with newgrounds

ThePlank:
> OpenGLRenderer crash

i think lime and openfl are outdated

DotWith:
> > OpenGLRenderer crash
> 
> i think lime and openfl are outdated

Thanks I was installing a dev openfl from git. It now works with non git version.

KitOdin:
since when was there a secret repo for fnf?

Steviegt6:
> since when was there a secret repo for fnf?

For quite some time, how else would they collaboratively develop it?

PR #1887, opened 2021-11-23
Add a new param for the new stuff so it doesn't crash

Cheemsandfriends:
Polymod added some new stuff and inserted on the haxelib, The case is that it may crash without the frameworkParams
This PR just adds it


AbnormalPoof:
i dont really think this is necessary because the bug was already fixed in a pr by mastereric

Cheemsandfriends:
> i dont really think this is necessary because the bug was already fixed in a pr by mastereric

It's for the mod support to work, And also that (At least on my HL experience)
It crashes without it, And I downloaded the haxelib one but yeah.
In fact, It's a current [Issue](https://github.com/larsiusprime/polymod/issues/74) that Eric knows right now and he's trying to make it optional, Not obligatory to use on Polymod.init

DotWith:
The polymod init is supposed to be in the Main new func

Cheemsandfriends:
> The polymod init is supposed to be in the Main new func

Why? If there's gonna be a polymod modding support how are you going to init?
And you can use it on more places rather than Main

ThePlank:
alr in week 7

PR #1952, opened 2022-02-23
Adjust speed in playstate

kudorado:
This might helpful to debug mid song event easier, and useful in some case i dunno, but it make your life easier, I guess.

ezgif com-gif-maker (4)


ThePlank:
adding to my engine 100%

ThePlank:
of course, not adding the controls, just adding the code for timeScale compatibility.

PR #2046, opened 2022-10-01
Cool Things

Gidk-g:
polymod shits


MemeHoovy:
not bad, but MasterEric is making a better system though

MemeHoovy (changes requested):
not bad, but MasterEric is making a better system though

Gidk-g:
> not bad, but MasterEric is making a better system though

true

Gidk-g:
goofy ahhhh

PR #2103, opened 2023-01-31
Fix week7 source code errors（I can't fix stress's cut animation）

salumod:
Music from dialogueBox.hx is not allowed to play after entering PlayState from freeplay


MemeHovy:
what the absolute hell happened to that commit @salumod

salumod:
> what the absolute hell happened to that commit @salumod

I didn't clone the updated library, so the updated thing is attached to the source code of v0.2.7

salumod:
But it's not a big problem. Just download the source code

salumod:
https://github.com/FunkinCrew/Funkin/pull/2124

PR #2137, opened 2023-03-27
Update readme.txt

cooldude2144:


cooldude2144:
why not

WizardMantis441:
geometry dash is pretty cool but ur making robtop look bad homie >:(

MemeHovy:
> why not

cuz fuck you is why

Issue #1968, opened 2022-03-27
Enhancement: Fixing unintentional resets on controller.


BlubParadox:
## What is your suggestion?
Changing line 633 in Controls.hx from 'Control.RESET => [Y]' to 'Control.RESET => [6]'

### Why should it be implemented?

This is going to sound a little stupid, but I play on a Guitar Hero Controller, in which the frets correspond to A,B,Y,X. The suggestion I am putting forward is to change the Y button to the Select button on controller. I am not the only person who plays the game on a Guitar Hero Controller, and I feel like it shouldn't have been set to Y in the first place, as some people would probably prefer to use A,B,Y,X instead of the D-Pad on a normal controller.

Select is a button that also wasn't in use to begin with, and would probably be a better fit for resetting.

### Why is it [6] and not [BACK] or [SELECT]?
This is the ID for the select button, as trying to use either [BACK] or [SELECT] simply won't work and the game won't compile.

#### why do you care
because i just wanna play the base game on a guitar hero controller (because im shit on a keyboard)

CPPxfnfmodsmaker:
hi bitch

CPPxfnfmodsmaker:
@CPPxfnfmodsmaker

PR #2157, opened 2023-06-15
Update Controls.hx

BlubParadox:
Changed Control.RESET from [Y] to [6]. 

Why? Because that means instead of pressing Y to reset the song, you instead press the select button, which prevents accidental resets if you are playing on a non-standard xbox controller. 

I brought up [this issue](https://github.com/FunkinCrew/Funkin/issues/1968) back in 2022 and it never got solved so I just decided I'd make the pull request myself.


MemeHovy (commented):
wouldn’t it have been better to made it R? and plus the repo is outdated which means they’re not accepting PRs

PR #2300, opened 2024-05-08
FIX WINDOW TWEEN

TheGaloXx:
the tween glitches out when pressing Y multiple times. just cancel the tween if it already exists


WizardMantis441:
i think that personally this shouldn't be fixed as it is way funnier but idk that's just my opinion

PR #2310, opened 2024-05-08
Better offsets handling for gameplay props.

Sword352:
Currently, Funkin directly changes the `x` and `y` properties for offsets. This is a bit tedious because it can break stuff, such as tweens.

This pull request changes the way animation offsets are handled without changing `x`, `y` or `offset`.
Ideally this could be implemented on other classes too.

### DIFF
--- a/source/funkin/play/character/BaseCharacter.hx
+++ b/source/funkin/play/character/BaseCharacter.hx
@@ -1,6 +1,7 @@
 package funkin.play.character;
 
 import flixel.math.FlxPoint;
+import flixel.FlxCamera;
 import funkin.modding.events.ScriptEvent;
 import funkin.play.character.CharacterData.CharacterDataParser;
 import funkin.play.character.CharacterData.CharacterRenderType;
@@ -118,19 +119,17 @@ class BaseCharacter extends Bopper
    */
   public var cameraFocusPoint(default, null):FlxPoint = new FlxPoint(0, 0);
 
+  /**
+   * Defines the animation offset.
+   */
+  public var animOffset:FlxPoint = FlxPoint.get();
+
   override function set_animOffsets(value:Array<Float>):Array<Float>
   {
     if (animOffsets == null) value = [0, 0];
     if ((animOffsets[0] == value[0]) && (animOffsets[1] == value[1])) return value;
 
-    // Make sure animOffets are halved when scale is 0.5.
-    var xDiff = (animOffsets[0] * this.scale.x / (this.isPixel ? 6 : 1)) - value[0];
-    var yDiff = (animOffsets[1] * this.scale.y / (this.isPixel ? 6 : 1)) - value[1];
-
-    // Call the super function so that camera focus point is not affected.
-    super.set_x(this.x + xDiff);
-    super.set_y(this.y + yDiff);
-
+    animOffset.set(value[0], value[1]);
     return animOffsets = value;
   }
 
@@ -570,9 +569,23 @@ class BaseCharacter extends Bopper
     }
   }
 
+  // override getScreenPosition (used by FlxSprite's draw method) to account for animation offsets.
+  override function getScreenPosition(?result:FlxPoint, ?camera:FlxCamera):FlxPoint
+  {
+    var output:FlxPoint = super.getScreenPosition(result, camera);
+    output -= animOffset;
+    return output;
+  }
+
   public override function onDestroy(event:ScriptEvent):Void
   {
     this.characterType = OTHER;
+  }
+
+  override function destroy():Void
+  {
+    animOffset = flixel.util.FlxDestroyUtil.put(animOffset);
+    super.destroy();
   }
 
   /**
--- a/source/funkin/play/character/BaseCharacter.hx
+++ b/source/funkin/play/character/BaseCharacter.hx
@@ -1,7 +1,6 @@
 package funkin.play.character;
 
 import flixel.math.FlxPoint;
-import flixel.FlxCamera;
 import funkin.modding.events.ScriptEvent;
 import funkin.play.character.CharacterData.CharacterDataParser;
 import funkin.play.character.CharacterData.CharacterRenderType;
@@ -118,20 +117,6 @@ class BaseCharacter extends Bopper
    * Set the position of this rather than reassigning it, so that anything referencing it will not be affected.
    */
   public var cameraFocusPoint(default, null):FlxPoint = new FlxPoint(0, 0);
-
-  /**
-   * Defines the animation offset.
-   */
-  public var animOffset:FlxPoint = FlxPoint.get();
-
-  override function set_animOffsets(value:Array<Float>):Array<Float>
-  {
-    if (animOffsets == null) value = [0, 0];
-    if ((animOffsets[0] == value[0]) && (animOffsets[1] == value[1])) return value;
-
-    animOffset.set(value[0], value[1]);
-    return animOffsets = value;
-  }
 
   /**
    * If the x position changes, other than via changing the animation offset,
@@ -569,23 +554,9 @@ class BaseCharacter extends Bopper
     }
   }
 
-  // override getScreenPosition (used by FlxSprite's draw method) to account for animation offsets.
-  override function getScreenPosition(?result:FlxPoint, ?camera:FlxCamera):FlxPoint
-  {
-    var output:FlxPoint = super.getScreenPosition(result, camera);
-    output -= animOffset;
-    return output;
-  }
-
   public override function onDestroy(event:ScriptEvent):Void
   {
     this.characterType = OTHER;
-  }
-
-  override function destroy():Void
-  {
-    animOffset = flixel.util.FlxDestroyUtil.put(animOffset);
-    super.destroy();
   }
 
   /**
--- a/source/funkin/play/stage/Bopper.hx
+++ b/source/funkin/play/stage/Bopper.hx
@@ -1,6 +1,7 @@
 package funkin.play.stage;
 
 import flixel.FlxSprite;
+import flixel.FlxCamera;
 import flixel.math.FlxPoint;
 import flixel.util.FlxTimer;
 import funkin.modding.IScriptedClass.IPlayStateScriptedClass;
@@ -68,6 +69,11 @@ class Bopper extends StageProp implements IPlayStateScriptedClass
   }
 
   /**
+   * Internally used to define the animation offsets to apply.
+   */
+  var _currentAnimOffset:FlxPoint = FlxPoint.get();
+
+  /**
    * The offset of the character relative to the position specified by the stage.
    */
   public var globalOffsets(default, set):Array<Float> = [0, 0];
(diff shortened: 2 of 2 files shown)


BlueColorsin (approved):

EliteMasterEric (changes requested):
I'm bewildered that this doesn't immediately break things, especially since you deleted a bunch of code that was supposed to keep Week 6 working, but base game characters do seem to work during my playthrough.

My main concerns are:
- Further playtesting of this change with mods
- Better variable naming because I got very confused thinking `animOffset` and `animOffsets` were the same (it should also be private so people don't accidentally edit it).
- Moving this code to `Bopper.hx`; you are overriding the behavior of stage props in the character class, and by doing so making characters and stage props behave differently.

Sword352:
I wouldn't be doing a pull request if I haven't tried, of course!

Re-tried the changes I did with some mod songs and seems to be working as excepted.

https://github.com/FunkinCrew/Funkin/assets/99792114/a635c04d-7364-4b22-8a88-d1856ad8183b

Raltyro:
review bugged

Sword it would be cool if you make it compatible with angles using the fast trigonometries function from FlxMath

```hx
final rad:Float = angle * FlxAngle.TO_RAD;
output.x -= animOffset.x * FlxMath.fastCos(rad) - animOffset.y * FlxMath.fastSin(rad);
output.y -= animOffset.y * FlxMath.fastCos(rad) - animOffset.x * FlxMath.fastSin(rad);
```

something like this, i probably screwed something writing this in phone but yeah

BlueColorsin:
> review bugged
> 
> Sword it would be cool if you make it compatible with angles using the fast trigonometries function from FlxMath
> 
> ```hx
> final rad:Float = angle * FlxAngle.TO_RAD;
> output.x -= animOffset.x * FlxMath.fastCos(rad) - animOffset.y * FlxMath.fastSin(rad);
> output.y -= animOffset.y * FlxMath.fastCos(rad) - animOffset.x * FlxMath.fastSin(rad);
> ```
> 
> something like this, i probably screwed something writing this in phone but yeah
shouldn't this have a 
```hx
if(angle != 0) {
    //code
}
```

Sword352:
Can't get it working, perhaps in another pr?

Raltyro:
idea from blue: normalized scale?

source/funkin/play/stage/Bopper.hx
```diff
@@ -95,12 +101,7 @@ class Bopper extends StageProp implements IPlayStateScriptedClass
     if (animOffsets == null) animOffsets = [0, 0];
     if ((animOffsets[0] == value[0]) && (animOffsets[1] == value[1])) return value;
 
-    var xDiff = animOffsets[0] - value[0];
-    var yDiff = animOffsets[1] - value[1];
-
-    this.x += xDiff;
-    this.y += yDiff;
-
+    _currentAnimOffset.set(value[0], value[1]);
```
EliteMasterEric, line 104: Is there a reason you can't just use `animOffsets` in place of `_currentAnimOffset` here? The values are identical.

source/funkin/play/character/BaseCharacter.hx
```diff
@@ -118,22 +118,6 @@ class BaseCharacter extends Bopper
    */
   public var cameraFocusPoint(default, null):FlxPoint = new FlxPoint(0, 0);
 
-  override function set_animOffsets(value:Array<Float>):Array<Float>
-  {
-    if (animOffsets == null) value = [0, 0];
-    if ((animOffsets[0] == value[0]) && (animOffsets[1] == value[1])) return value;
-
-    // Make sure animOffets are halved when scale is 0.5.
-    var xDiff = (animOffsets[0] * this.scale.x / (this.isPixel ? 6 : 1)) - value[0];
```
EliteMasterEric, line 127: Can you give an explanation as to how/why your implementation doesn't require this custom scale logic?

EliteMasterEric (changes requested):
Still looking for clarification on this one

source/funkin/play/character/BaseCharacter.hx
Sword352: To be honest I'm not quite sure why it was there in the first place.
People says this somewhat makes the character progressively "fly" because scale isn't applied on `value`.
If this was meant for week 6 then I don't see how that has an effect as it just scales then divide by 6. However if this is meant to normalize the offset by scale (which I thought funkin doesn't) then I can re-implement that back.

source/funkin/play/stage/Bopper.hx
Sword352: Probably an oversight... will remove that.

gamerbross:
I love this solution, but it has an issue, the base game system somehow ignores the idle offsets, no matter what offset you put in idle, it will behave as it doesnt have any (but the rest of animations work just fine) so you would have to implement that.
(i think im not explaining correctly, lemme know)

Sword352:
> I love this solution, but it has an issue, the base game system somehow ignores the idle offsets, no matter what offset you put in idle, it will behave as it doesnt have any (but the rest of animations work just fine) so you would have to implement that. (i think im not explaining correctly, lemme know)

not sure I'm following... maybe re-explain it yeah

gamerbross:
I realized this while porting VS Entity to the base game. In the mod, Nikusa has offsets in her idle animation (instead of position offset)
image
And I just did the same in the port, but putting an animation offset to idle, just does nothing and retargets all of the animation offsets to be 0, 0
(If idle offsets are [195, 501], then they "are changed" to be [0, 0] and (for example) singLEFT offsets were [181, 474], and they changed to [-14, -27] (got substracted by original idle ofssets))
So the problem with this implementation is that it doesnt have the same behavior, it just renders Nikusa at the original offsets
(isnt really your fault, more like a weird thing done in base game)
Resume: somehow you should make idle offsets to [0, 0] and substract idle original offsets to every animation offsets

gamerbross:
whats the status of this? we really need it 🙏

EliteMasterEric:
I was porting some mods this week and encountered the issues people have been having with offsets, so I'm looking at this PR now.

EliteMasterEric:
I finally reviewed and approved this branch

Since this invalidates both #2332 and #2972, I manually implemented a fix for the scaling issue myself, and will be closing those issues.

EliteMasterEric (approved):

gamerbross:
could you show the fix? so I can use it meanwhile

EliteMasterEric:
> could you show the fix? so I can use it meanwhile

```haxe
  function applyAnimationOffsets(name:String):Void
  {
    // I don't take globalOffsets into account here anymore, I moved that to getScreenPosition.
    var offsets = animationOffsets.get(name);
    this.animOffsets = offsets;
  }

  override function getScreenPosition(?result:FlxPoint, ?camera:FlxCamera):FlxPoint
  {
    var output:FlxPoint = super.getScreenPosition(result, camera);
    // Combine the offsets then apply scale before adding.
    output.x -= (animOffsets[0] - globalOffsets[0]) * this.scale.x;
    output.y -= (animOffsets[1] - globalOffsets[1]) * this.scale.y;
    return output;
  }
  ```

**Comment by gamerbross** [comment]
> > could you show the fix? so I can use it meanwhile
> 
> ```hx
>   function applyAnimationOffsets(name:String):Void
>   {
>     // I don't take globalOffsets into account here anymore, I moved that to getScreenPosition.
>     var offsets = animationOffsets.get(name);
>     this.animOffsets = offsets;
>   }
> 
>   override function getScreenPosition(?result:FlxPoint, ?camera:FlxCamera):FlxPoint
>   {
>     var output:FlxPoint = super.getScreenPosition(result, camera);
>     // Combine the offsets then apply scale before adding.
>     output.x -= (animOffsets[0] - globalOffsets[0]) * this.scale.x;
>     output.y -= (animOffsets[1] - globalOffsets[1]) * this.scale.y;
>     return output;
>   }
> ```

add the isPixel thing, it broke senpai offsets lmao

**Comment by gamerbross** [comment]
also characters with idle offsets break when restarting the song because of this 
image
Just remove the `+ character.animOffsets[x]`

**Comment by gamerbross** [comment]
and why substracting globalOffsets? the stage already does that, just causes it to duplicate it. I feel like it should change the position and not act like a visual offset

**Comment by gamerbross** [comment]
my code end to this so it works properly:
```haxe
  override function getScreenPosition(?result:FlxPoint, ?camera:FlxCamera):FlxPoint
  {
    var output:FlxPoint = super.getScreenPosition(result, camera);
    // Combine the offsets then apply scale before adding.
    output.x -= animOffsets[0] * (!isPixel ? this.scale.x : 1);
    output.y -= animOffsets[1] * (!isPixel ? this.scale.y : 1);
    return output;
  }
```

gamerbross:
it alsoooo breaks the spirit position due to idle offsets (-220, -280)
image
fixed offsets:
```json
  [
    {
      "name": "idle",
      "prefix": "idle spirit_",
      "offsets": [0, 0]
    },
    {
      "name": "singLEFT",
      "prefix": "left_",
      "offsets": [20, 0]
    },
    {
      "name": "singDOWN",
      "prefix": "spirit down_",
      "offsets": [390, 390]
    },
    {
      "name": "singUP",
      "prefix": "up_",
      "offsets": [0, 40]
    },
    {
      "name": "singRIGHT",
      "prefix": "right_",
      "offsets": [0, 0]
    }
  ]
```

gamerbross:
@EliteMasterEric sorry for mention but maybe you had notifications off on this PR

KutikiPlayz:
I mean correct me if I'm wrong but the way I see all this isPixel bs is that initially when they were adding the scale multiplication, it broke pixel characters cuz they're scaled up, and instead of just changing the pixel character's offsets to work with the new system they just slapped a bandaid on it by hardcoding a "fix" for pixel characters, which just turns off the scale multiplication.
And I know they didn't change the offsets because if you look back at the legacy 0.2.x version, the spirit character offsets for example, are the exact same as they are now.

If it were me I'd just readjust the offsets in each of the pixel character files, and then you don't need any of that hardcoded bandaid fix.

With a sprite that in reality is, lets say 64x64, an animation offset of 30 would work at a higher scale like it originally did, but now that the offsets are scaled as well, that turns into 180, and then it's incorrect.
So just divide the old offset that used to work by the scale and boom you've now got better numbers that actually work for the small sprite.

spirits new working offsets would be this for example:
```
  [
    {
      "name": "idle",
      "prefix": "idle spirit_",
      "offsets": [0, 0]
    },
    {
      "name": "singLEFT",
      "prefix": "left_",
      "offsets": [3, 0]
    },
    {
      "name": "singDOWN",
      "prefix": "spirit down_",
      "offsets": [65, 65]
    },
    {
      "name": "singUP",
      "prefix": "up_",
      "offsets": [0, 7]
    },
    {
      "name": "singRIGHT",
      "prefix": "right_",
      "offsets": [0, 0]
    }
  ]
```
based on gamerbross's updated spirit offets which adjust for the removal of the idle offsets.

Now I do understand the fact that doing this would require everyone to readjust the offsets of any scaled character that they're porting to the game from elsewhere, but personally I see that as a minor inconvenience at best. Whoever made that mod porter can just automatically divide any animation offsets by the character scale in the program, and then boom. No issues for anyone.

PR #2312, opened 2024-05-08
Spelling and grammar fixes in documentation

tposejank:
Small typo, but confusing for phone readers like me lol.
7 key definitely works so I see no point to mark it as not implemented.


tposejank:
I added `-clean` as a suggestion to prevent the hassle of deleting the export folder with the file explorer. Many files in there.

PR #2316, opened 2024-05-08
Fix for AMD

TheSPREEGuy:
Fixes AMD gpu users


EyeDaleHim:
Instead of removing things entirely simply due to incompatibility with certain hardware, we should find a way to make the code compatible with all hardware as much as we can as a community.

Ralsin:
there's a better fix to this that is even easy to do in the compiled code itself

EliteMasterEric:
Superceded by #2357.

PR #2322, opened 2024-05-08
Field Read and Write Overloading for Prefereneces

saturn-volv:
This adds an abstracted `PreferenceMap`, which behaves as a StringMap but with some field read and write overloading.

So now instead of needing to call:
```haxe
Save.instance.options.downscroll = true;
```
You can accomplish the same thing with:
```haxe
Preferences.get().downscroll = true;
```

There are other use cases for this but I'm tired and cannot bring myself to list them all right now

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -1,12 +1,70 @@
 package funkin;
 
+import haxe.ds.StringMap;
 import funkin.save.Save;
+
+/**
+ * An abstract class which behaves as a StringMap, caching all getters of options,
+ * while also providing a more versitile way of modders adding their own preference setters and getters,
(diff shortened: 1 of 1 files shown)


Raltyro (approved):
Pretty sick
Though i think you can instead immediately do return get().value = value;

saturn-volv:
> return get().value = value;

I initially did but the code style doesn't want inner assignment used.
image

Raltyro:
i forgot codestyle exists whoopsm..

EliteMasterEric:
This looks like a great bit of code cleanup!

Also, the codestyle is bugged, it should be allowing inner assignment in return statements.

EliteMasterEric:
I just looked into this more and I actually don't think this is much of an improvement over the existing `Preferences` map.

PR #2328, opened 2024-05-08
New judge application logic

nebulazorua:
Does a couple of things:
- Adds hitDiff to HitNoteScriptEvent
- Adds isComboBreak to HitNoteScriptEvent
- Adds doesNotesplash to HitNoteScriptEvent
- Moved code related to applying the judgement stats (score, health, etc) to a new applyScore function
- popupScore now only has 2 arguments, judgement and combo.

So, why these changes?
popupScore was a bit of a misleading title, as it doesn't just display the judges, but it also applied logic related to them, so I have split the functionality between popupScore and applyScore. This, in my opinion, makes a lot more sense and allows for scripts to display judgements seperately from score application, i.e in the case of mine/hurt notes where maybe they'd want to display a custom judgement to show that you hit a mine. Perhaps HitNoteScriptEvent.showScore can be a thing, which'd toggle whether popupScore is called or not (though in scripts you could just event.cancel and call applyScore yourself without calling popupScore to hide the judgement display)

The changes to HitNoteScriptEvent to add doesNotesplash , hitDiff and isComboBreak is to give more freedom to modders when doing things on note hit, such as special note kinds with lower hitboxes, like mines, or even custom judgements.

Mines, when implemented in FNF mods, generally combo break and have a smaller hitbox, so an example Mine script's onNoteHit could look a bit like this:
```haxe
override function onNoteHit(event){
  if(event.note.kind == 'mine'){
    var noteDiff = event.hitDiff;
    if(Math.abs(noteDiff) > 60){ // Lowers the hitbox from the full 166ms to 60ms.
      event.cancel();
      return;
    }
    event.isComboBreak = true;
    event.healthChange = -0.2;
    event.judgement = 'mine';
    event.score = -150;
    event.doesNotesplash = false;
  }
}
```

This would not be as easily possible without these changes to HitNoteScriptEvent and how score is applied.

### DIFF
--- a/source/funkin/modding/events/ScriptEvent.hx
+++ b/source/funkin/modding/events/ScriptEvent.hx
@@ -140,16 +140,29 @@ class HitNoteScriptEvent extends NoteScriptEvent
    */
   public var score:Int;
 
-  public function new(note:NoteSprite, healthChange:Float, score:Int, judgement:String, comboCount:Int = 0):Void
+  /**
+   * Whether the judgement caused a combo break.
+   */
+  public var isComboBreak:Bool = false;
+
+  /**
+   * The time difference when the player hit the note
+   */
+  public var hitDiff:Float = 0;
+
+  public function new(note:NoteSprite, healthChange:Float, score:Int, judgement:String, isComboBreak:Bool, comboCount:Int = 0, hitDiff:Float = 0):Void
   {
     super(NOTE_HIT, note, healthChange, comboCount, true);
     this.score = score;
     this.judgement = judgement;
+    this.isComboBreak = isComboBreak;
+    this.hitDiff = hitDiff;
   }
 
   public override function toString():String
   {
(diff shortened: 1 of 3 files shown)


crowplexus:
i was gonna do this myself so I could make the killer judge work but here we go

EliteMasterEric:
The diffs for this looked good, I'm going to have to test this in-game but otherwise this seems like a good set of features.

nebulazorua:
Any update on if/when this might get merged? I've been thinking of updating it to add `NoteHitScriptEvent.isSplash` so you can turn on/off the note-splash. Would allow scripts to add custom judgements, or force certain note types to always splash for i.e explosions for mines, etc.

nebulazorua:
Updated the PR to include doesNotesplash in the list of changes i've made

EliteMasterEric (approved):

nebulazorua:
seems to have gotten undone at some point since the merge

nebulazorua:
(ignore my last reply that i removed i was on the wrong branch teehee)

PR #2335, opened 2024-05-08
Mini repo cleanup

doggogit:
- Move `CHANGELOG`, `CODESTYLE` and `Modding` from root to `docs/`.
- Make sure `Project.xml` generates the `changelog.txt` file from `docs/CHANGELOG` and not from `root/CHANGELOG`.
- Attach a hyperlink to the Apache and custom `funkin-assets` license in `LICENSE`.


EliteMasterEric:
I don't think any of these changes are relevant to us, we probably just want to rework those files entirely or just delete them.

Plus we don't want ANYBODY to touch the LICENSE files without permission, even to make small fixes.

doggogit:
> I don't think any of these changes are relevant to us, we probably just want to rework those files entirely or just delete them.
> 
> Plus we don't want ANYBODY to touch the LICENSE files without permission, even to make small fixes.

Gotcha, thanks for the reply :)

Issue #2289, opened 2024-05-06
Bug Report: "Memory Leak" when opening screenshots folder


gamerbross:
## Game Version
- [ ] [Newgrounds](https://www.newgrounds.com/portal/view/770371)
- [x] [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below
- - [x] Windows (Desktop)
- - [ ] Mac
- - [ ] Linux

## What version of the game are you using? Look in the bottom left corner of the main menu. (ex: 0.2.7, 0.2.1, shit like that)
0.3.2, (Version 11)

## Have you identified any steps to reproduce the bug? If so, please describe them below in as much detail as possible. Use images if possible.
- Screenshot with the built-in thing
- Click the screenshot preview to open the folder

## Please describe your issue. Provide extensive detail and images if possible.
It's not really a memory leak, but it creates a separated `explorer.exe` process, which doesn't terminate when you close the window, creating unused process that take memory.
![image](https://github.com/FunkinCrew/Funkin/assets/55158797/c808ca73-1396-4d06-a50e-56cc059197b3)

This could be fixed changing the function `funkin.util.FileUtil.openFolder` to run some C++ code that open a folder without creating a new process.
This was the solution for my own engine.
```haxe
#if windows
@:functionCode('
static HMODULE hShell32 = LoadLibraryW(L"shell32.dll");
static const auto pShellExecuteW = (decltype(ShellExecuteW)*)GetProcAddress(hShell32, "ShellExecuteW");
pShellExecuteW(NULL, L"open", folder.__WCStr(), NULL, NULL, SW_SHOWDEFAULT);
')
#end
public static function openFolder(folder: String)
{
	#if !windows
	// run not windows code to do this
	#end
}
```
```haxe
#if windows
@:cppFileCode("
#include <windows.h>
#include <shellapi.h>
")
#end
class FileUtil
```

moxie-coder:
that’s literally how windows work, it’s not the games problem, it’s your OSes problem

NotHyper-474:
Actually that happens when clicking any "Open folder" button, not only for screenshots. Opening folders by right-clicking explorer icon in taskbar just allocates more memory in "main" instance instead of creating others.

moxie-coder:
> Actually that happens when clicking any "Open folder" button, not only for screenshots, but your code still applies. Opening folders by right-clicking explorer icon in taskbar just allocates more memory in "main" instance instead of creating others.

[...]

PR #2353, opened 2024-05-08
Use `ShellExecute` for `FileUtil.openFolder` to avoid "memory leaks"

gamerbross:
As mentioned in this issue (#2289), the created `explorer.exe` process by `FileUtil.openFolder` doesn't get terminated by it's own, taking up memory.
Maybe my C++ code implemention is a little bit "raw" and should be made in a more "softer" way lol, but kinda shows the point.


gamerbross:
I saw Eric's stream and yeah, this maybe isn't an important issue because so many programs do the same

Issue #2279, opened 2024-05-05
Bug Report: Can't select players to the right in the chart editor (see video)


ytggobs:
[weed]: <> (FILL THIS ISSUE THING OUT AS MUCH AS POSSIBLE)
[weed]: <> (OR ELSE YOUR ISSUE WILL BE LESS LIKELY TO BE SOLVED!)
[weed]: <> (DO NOT POST ABOUT ISSUES FROM OTHER FNF MOD ENGINES! I CANNOT AND PROBABLY WON'T SOLVE THOSE!)
[weed]: <> (GO TO THEIR RESPECTIVE GITHUB ISSUES AND REPORT THEM THERE LOL!)

[weed]: <> (ALSO MAKE SURE THAT YOU USE PROPER LABELS, IF YOU'RE RUNNING INTO COMPILER ISSUES, USE THE compiler issue LABEL!!!)

### If you are playing the game in a browser, what site are you playing it from?

[weed]: <> (Put an X in the [ ] thingies to fill out checkbox!)
[weed]: <> (something like [x] pretty much, don't screw up or you will look stupid)

- [ ] [Newgrounds](https://www.newgrounds.com/portal/view/770371)
- [x] [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below
- - [x] Windows
- - [ ] Mac
- - [ ] Linux

## What version of the game are you using? Look in the bottom left corner of the main menu. (ex: 0.2.7, 0.2.1, shit like that)
0.3.2

## Have you identified any steps to reproduce the bug? If so, please describe them below in as much detail as possible. Use images if possible.
Go to the chart editor and create new chart and open metadata then select a character to the right and it doesn't change
## Please describe your issue. Provide extensive detail and images if possible.
It won't change the character when you click one on the right in metadata you can hear that I am clicking

https://github.com/FunkinCrew/Funkin/assets/62442697/82c0aa80-0578-4985-a3a0-045f22bb175d

MadBear422:
#2370
The issue is briefly explained here, with a potential solution to make it possible to select these characters!

ytggobs:
fixed in v0.3.3

ytggobs:
fixed in v0.3.3

PR #2370, opened 2024-05-08
Character selection fix for chart editor

MadBear422:
image
This corrects an issue within the character dropdown grid of the chart editor, where the user cannot select them because of the previous hitbox width of the grid. Mention of this bug is found here: #2279 
image
With this change it makes these characters selectable. No issues known so far


EliteMasterEric:
We were trying to figure this one out for ages, I thought for sure it was HaxeUI's fault.

Thanks for the PR. We will review and report back.

Pietrodjaowjao:
It's bad that we cant compile to test.

EliteMasterEric (approved):
This worked great! Tested on stream.

PR #2383, opened 2024-05-09
Fix windows compilation issue

Pietrodjaowjao:


Raltyro:
Windows compile fine already though, i was thinking of instead of making A-Bot Visuals ftt only works for web, why not somehow making windows work too somehow

Pietrodjaowjao:
makes sense

Pietrodjaowjao:
i will try to do this instead and rename the pr lol

Raltyro:
Welp since this uses funkin.vis which is a seperate repo, i think you should close this pr instead, also the ftt work seem to be hard though to make it work for native

EliteMasterEric:
The game already builds on desktop without this, this doesn't fix anything.

PR #2391, opened 2024-05-09
[ENHANCEMENT] timer sequence

cyn0x8:
qol class for a making batches of flxtimers, so you can do stuff like
```haxe
new Sequence([
    {time: 1, callback: some_func},
    {time: 2.5, callback: () -> {
        some_var = true;
    }}
], Conductor.beatLengthMs / 1000);
```
instead of
```haxe
new FlxTimer().start(1 * (Conductor.beatLengthMs / 1000), some_func);
new FlxTimer().start(2.5 * (Conductor.beatLengthMs / 1000), () -> {
    some_var = true;
});
```

### DIFF
--- a/source/funkin/Conductor.hx
+++ b/source/funkin/Conductor.hx
@@ -6,6 +6,7 @@ import flixel.math.FlxMath;
 import funkin.data.song.SongData.SongTimeChange;
 import funkin.data.song.SongDataUtils;
 import funkin.save.Save;
+import funkin.util.TimerUtil.SongSequence;
 import haxe.Timer;
 import flixel.sound.FlxSound;
 
@@ -501,6 +502,8 @@ class Conductor
       prevTime = this.songPosition;
       prevTimestamp = Std.int(Timer.stamp() * 1000);
     }
+
+    if (this == Conductor.instance) @:privateAccess SongSequence.update.dispatch();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


cyn0x8:
im unsure if a pr is the right place for a feature request... but posting all that code into a feature request issue feels unfit

Raltyro:
Oh wait instead of it depend on deltatimes, why not also make it for conductor songPosition too

cyn0x8:
> Oh wait instead of it depend on deltatimes, why not also make it for conductor songPosition too
> 
> (This sounds much for a feature actually and there prob already a equivalent that is events)

doesnt have to be used in just playstate tho...
but ill try looking for a way to make it optionally follow song pos, or just pause/resume/stop when the song pauses/resumes/stops

Pietrodjaowjao:
Maybe you want to use it everywhere, not only inside songs

nebulazorua:
Could add an option or make a variant that uses Conductor.instance.songPosition then

cyn0x8:
i believe this should do it? do tell me if any of you find potential errors

ninjamuffin99:
this looks lovely, maybe can even be in base Flixel code ! gonna poke @Geokureli to see if this might be cool, otherwise I'll mark this as reviewing internally!

EliteMasterEric:
Going to suspend this and wait to see if it gets merged into Flixel.

Geokureli:
> Going to suspend this and wait to see if it gets merged into Flixel.

I don't plan on adding this in the immediate future, and if I did I'm considering something less adhoc

If people want this feature now they can borrow this implementation, which seems good, for now

PR #2405, opened 2024-05-09
improved conductor time updating in playstate

Snirozu:
this pr should make note movement smoother on screens that have a higher refreshing rate than 60FPS

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -413,6 +413,11 @@ class PlayState extends MusicBeatSubState
   var startingSong:Bool = false;
 
   /**
+   * True if the song has finished playing.
+   */
+  var endingSong:Bool = false;
+
+  /**
    * Track if we currently have the music paused for a Pause substate, so we can unpause it when we return.
    */
   var musicPausedBySubState:Bool = false;
@@ -909,6 +914,12 @@ class PlayState extends MusicBeatSubState
       Conductor.instance.update(Conductor.instance.songPosition
         - (Conductor.instance.instrumentalOffset + Conductor.instance.formatOffset + Conductor.instance.audioVisualOffset)
         + elapsed * 1000 * playbackRate); // Normal conductor update.
+
+      // FlxG.sound.music.onComplete may sometimes not get fired up lol
+      if (Conductor.instance.songPosition >= FlxG.sound.music.length)
+      {
+        endSong(false);
+      }
     }
 
     var androidPause:Bool = false;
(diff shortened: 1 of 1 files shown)


biomseed:
> this pr should make note movement smoother on screens that have a higher refreshing rate than 60FPS

is this why things feel slightly delayed but the input feels fine?

ninjamuffin99:
would you be able to make a short before/after demo video of differences?

Snirozu:
it's probably not noticeable here because obs may have done something with the framerate and i don't want to configure it twice but here are the recorded clips

https://github.com/FunkinCrew/Funkin/assets/72814880/b42aebe0-7dc6-4d60-81d5-d2e8c80bfee4



https://github.com/FunkinCrew/Funkin/assets/72814880/33c3ed52-8cbd-4630-883c-489edddc1f99

KutikiPlayz:
my obs records at my actual monitor refresh rate, and I'm playing a song with a higher scroll speed so it should be more noticeable here

https://github.com/FunkinCrew/Funkin/assets/73810550/a2180cae-f433-4128-b124-d130bc8b3638

https://github.com/FunkinCrew/Funkin/assets/73810550/0275367d-5a47-4337-bc27-7e3a30138550

I've got no other changes to my build besides this pr (in fact it's still 0.3.2)

gamerbross:
in `resyncVocals` function, you dont need a time for argument in `FlxG.sound.music.play`,
`FlxG.sound.music.play(Conductor.instance.songPosition);`
can be
`FlxG.sound.music.play(Conductor.instance.songPosition);`
because if the sound is paused, this argument is ignored

```haxe
	public function play(ForceRestart:Bool = false, StartTime:Float = 0.0, ?EndTime:Float):FlxSound
	{
		if (!exists)
			return this;
			
		if (ForceRestart)
			cleanup(false, true);
		else if (playing) // Already playing sound
			return this;
			
		if (_paused) // here, it just uses StartTime when the sound is not paused
			resume();
		else
			startSound(StartTime);
			
		endTime = EndTime;
		return this;
	}
```

EliteMasterEric:
We looked at this and can't really see the difference here, and the way this changes the Conductor to be no longer based on the FlxSound object makes me concerned about desyncing the song.

Rejected.

gamerbross:
The problem of using music time is visible in higher refresh rates, in 165hz or even 144hz I can clearly notice a difference, it's so much smoother. Probably this wasn't the best implementation but should get changed so `elapsed` is used instead

Issue #2298, opened 2024-05-08
[COMPILING] Installation of funkin.vis fails with "Execution error"


EliteMasterEric:
When attempting to build the project on Windows, some users experience the following error:

![image](https://github.com/FunkinCrew/Funkin/assets/4635334/89c18fcb-c873-4dcf-ad7c-4dca6189b860)

```
PS C:\Users\godee\Desktop\Funkin> hmm reinstall --force funkin.vis
reinstalling git library: funkin.vis from https://github.com/FunkinCrew/funkVis (ref: Some(98c9db09f0bbfedfe67a84538a5814aaef80bdea), dir: None)
haxelib --never remove funkin.vis
Error: Library funkin.vis is not installed
Warning: command "haxelib --never remove funkin.vis" failed with status: 1 in dir: C:\Users\godee\Desktop\Funkin/
haxelib --never git funkin.vis https://github.com/FunkinCrew/funkVis 98c9db09f0bbfedfe67a84538a5814aaef80bdea
Installing funkin.vis from https://github.com/FunkinCrew/funkVis branch: 98c9db09f0bbfedfe67a84538a5814aaef80bdea
Error: std@sys_remove_dir
Execution error: command "haxelib --never git funkin.vis https://github.com/FunkinCrew/funkVis 98c9db09f0bbfedfe67a84538a5814aaef80bdea" failed with status: 1 in cwd: C:\Users\godee\Desktop\Funkin/
Use "hmm help" to see usage
PS C:\Users\godee\Desktop\Funkin>
```

This issue is related to the game trying to download the code for Nene's visualizer (which is in a separate dependency library) and failing to do so.

TheJustAnto:
So is there a way to fix this error ?

Eviethecoder:
you could try to manually download the library through the hmm files GitHub link for it. and add it to the haxelib folder manually. at least till this issue gets solved

Gato-chistoso:
`haxelib remove funkin.vis` then `haxelib git funkin.vis https://github.com/FunkinCrew/funkVis`

prplnorangesoda:
Getting the same error
![image](https://github.com/FunkinCrew/Funkin/assets/60671867/41129523-da9d-4b8c-b658-b4fe96e6d48e)
Original command: `hmm install`

```
Download complete: 295.46KB in 0s (3110KB/s)
Installing format...
  Current version is now 3.5.0
Done
[...]

PR #2407, opened 2024-05-09
Replace hmm lib funkin.vis commit hash with 'new' one

NotHyper-474:
As I mentioned in https://github.com/FunkinCrew/Funkin/issues/2298, currently `hmm.json` contains an outdated hash for a funkVis commit which causes `hmm install` to fail.

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -49,7 +49,7 @@
       "name": "funkin.vis",
       "type": "git",
       "dir": null,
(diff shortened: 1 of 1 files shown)


ImCodist:
This is fixed in the develop branch. 1e359030c31840850c55c06dd9d8e8f394cb8049
Use the develop branch.

PR #2451, opened 2024-05-11
abstract enum pagename

nebulazorua:
turns PageName into an abstract enum, which should make adding custom options pages easier and allow for more compatibility between mods that would like to add custom options pages, as they won't have to leech off an existing, unused PageName (like CoolHUD does) and also won't have to rewrite code related to options to add a custom page in a more compatible way.

Note that this will break current mods that add custom options, though


nebulazorua:
please 🙏 i wanna be able to make options pages in scripts and not do it jankily

EliteMasterEric:
Will do a bunch more PR review tomorrow, if this doesn't immediately break the options menu then it should be good to merge.

AbnormalPoof:
How would this look in a mod/script? I'm mildly confused.

PR #2456, opened 2024-05-11
[BUGFIX] Fix Charting Game Over

gamerbross:
Pressing Escape in gameover screen while charting crashed because its closes PlayState but it tries to access it's instance after.


gamerbross:
just fixed minimal mode crash

gamerbross:
bruh i have to redo this pr just because i renamed the branch

PR #2457, opened 2024-05-11
[BUGFIX] Freeplay Crash when song is invalid

gamerbross:
Caused when the level lists a non-existing song.


tposejank:
this bug is caused by incorrectly naming songs in levels. song folders currently have no exceptions to try and see if their names can be reformatted (lowercased, dashed etc). cool!

tposejank (approved):

EliteMasterEric (approved):
I've had this happen if a song's JSON data doesn't parse properly and then a week tries to reference it.

PR #2464, opened 2024-05-11
Character Pixel Icon Data Accessed via Character Metadata

MadBear422:
Freeplay and the chart editor now read from character metadata instead of hard-coding assigned character pixel icons. Character metadata files updated with respective icons previously assigned in the source code. Additionally, the ability to flip the icons horizontally and adjusting the origin offset is now included via the metadata.


MadBear422:
I messed up; let me recreate this PR

PR #2465, opened 2024-05-11
[ENHANCEMENT] Character Pixel Icon Data Accessible via Character Metadata

MadBear422:
Freeplay and the chart editor now read from character metadata instead of hard-coding assigned character pixel icons. Existing character metadata files have updated with respective icons previously assigned in the source code . Additionally, the ability to flip the icons horizontally and adjusting the origin offset is now included via the metadata.

Situation:
For the modders who have multiple versions of some character and would like to use the same character icon, it was previously not possible simply due to the fact that prior to this change, the pixel icon is determined by the character's id name rather than a specified pixel ID. In this case, I have a characters labeled as "hugenate-opponent" and "hugenate-playable". I also have an icon file named "hugenate", and I want both of these characters to use the same icon. However, neither will work since they would have to be the same name as the freeplay icon file.
image
Now with this change, it corrects the issue and allows for me to individually assign the icon id to the players in their metadata. NOTE: If the id is left as null, the game will automatically set it to the same name as their character id.
image
After modifying it with this data, this is the new result!
image
New parameters such as flipX and origin are also included to adjust icons if they overlap the text:
image
Before:
image
After:
image


MadBear422:
Sorry about the other PR; I chose the wrong branch that I wanted to create the PR for!

tposejank:
pretty cool, should be looked into

EliteMasterEric:
At an initial glance, this looks great and I wouldn't have implemented this differently (save maybe naming it `freeplayIcon` but that's too much of a nitpick and I think it's fair to just leave it as is).

Do definitely want to do more review first though.

MadBear422:
Currently testing the resolved merge conflicts

AbnormalPoof:
Merge conflicts.

AbnormalPoof:
@MadBear422 Do you mind rebasing this PR to squash all the commits into one? The commit history is a mess.

MadBear422:
@AbnormalPoof Working on that right now don't worry!

MadBear422:
I'll close this for now. Git is giving me a hard time

MadBear422:
This PR is dead. I'm submitting a new one of this shortly.

MadBear422:
#4497 Moved to this PR

PR #2466, opened 2024-05-11
Fixed Freeplay Scrolling Issue

JugieNoob:
Before this fix, scrolling through songs in the freeplay menu using the scroll wheel would not function as expected. This pull request fixes this problem.


ninjamuffin99:
very nice thank you! merged this internally for the 0.3.3!

PR #2481, opened 2024-05-12
[BUGFIX?] SaveData using a player 2 controls instead of player1 controls

gamerbross:
SaveData was using player2 controls because of an incorrect check (assumed playerId 0 is for player1, so always used player2 instead).

> [!WARNING]  
> This change will wipe user's controls.
> Could be fixed with some kind of migration of the save data, maybe increasing savedata version and moving player2 controls to player1 controls, and emptying player2 controls


Matriculaso:
Esto es bien mierda pq no 
tiene var singSong = "Douh homero";

gamerbross:
if it works dont fix it

PR #2490, opened 2024-05-13
[Linux] Fix "Unexpected argument '&'" error when trying to open URL

NotHyper-474:
Fixes the background thingie not being passed properly to the command line due to how Sys works, which causes an error and prevents the link from being opened

~~Could be something with my system for all I know, not sure if being on Ubuntu WSL2 changes how the command behaves.
If anyone else also compiles in Linux (no WSL) please confirm if this error happens to you too.~~

Edit: I tested this through native Ubuntu and also got the same error, so not related to WSL, unless it's related to Ubuntu in general.
Edit 2: This error is caused by how Haxe's Sys.command appends the arguments to the console, encasing them with single quotes. So the original code would cause the command to be `xdg-open 'https:/link.here' '&'` which makes xdg think that & is an argument, so not related to any Linux distro.

### DIFF
--- a/source/funkin/util/WindowUtil.hx
+++ b/source/funkin/util/WindowUtil.hx
@@ -24,7 +24,7 @@ class WindowUtil
   {
     #if CAN_OPEN_LINKS
     #if linux
-    Sys.command('/usr/bin/xdg-open', [targetUrl, '&']);
+    Sys.command('/usr/bin/xdg-open ' + targetUrl + ' &');
     #else
     // This should work on Windows and HTML5.
     FlxG.openURL(targetUrl);
--- a/source/funkin/util/WindowUtil.hx
+++ b/source/funkin/util/WindowUtil.hx
@@ -24,7 +24,7 @@ class WindowUtil
   {
     #if CAN_OPEN_LINKS
     #if linux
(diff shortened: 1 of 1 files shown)


gamerbross:
using string interpolation should be more readable
```haxe
Sys.command('/usr/bin/xdg-open $targetUrl &');
```

NotHyper-474:
Alright, thanks, gonna do that.

ninjamuffin99:
if someone can show/record some before/after of this change, that would be very helpful :) 
also im linux noob still, this is consistent across distros ? i was always curious about that command...

NotHyper-474:
xdg-open is part of the xdg-utils package, which generally is [installed by default in distros with a desktop manager](https://stackoverflow.com/a/27871581), so it's not guaranteed to work in _every single_ distro, but there doesn't seem to exist a better alternative.


And here's the before and after, tested in Kali Linux (WSL)

https://github.com/FunkinCrew/Funkin/assets/40342021/667ce8b5-d48e-4aab-a9e8-37561ad8d61c

https://github.com/FunkinCrew/Funkin/assets/40342021/0173b1d2-cfc6-4f7a-a577-a22e554a89c7

ninjamuffin99:
very nice !

PR #2499, opened 2024-05-13
Fix Web Loading Bar

lemz1:
I noticed (at least for me) that the loading bar starts from the middle which is probably because previously scale.x was used. And another issue is that it just doesn't move at all.
I fixed this issue by using setGraphicSize and updateHitbox afterwards
I removed the lerp since it made the loading bar look weird


ninjamuffin99:
nice! internally I'll be adding the lerp back in, but thank you! I'll merge this here and add the 0.3.3 milestone

PR #2500, opened 2024-05-14
Small FreeplayState optimization

richTrash21:
- State beneath FreeplayState won't render unless player goes back.
- Remove and destroy Freeplay camera after FreeplayState got destroyed.

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -527,6 +527,9 @@ class FreeplayState extends MusicBeatSubState
       orangeBackShit.visible = true;
       alsoOrangeLOL.visible = true;
       grpTxtScrolls.visible = true;
+
+      // render optimisation
+      _parentState.persistentDraw = false;
     });
 
     generateSongList(null, false);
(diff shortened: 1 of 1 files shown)


source/funkin/ui/freeplay/FreeplayState.hx
```diff
@@ -527,12 +528,15 @@ class FreeplayState extends MusicBeatSubState
       orangeBackShit.visible = true;
       alsoOrangeLOL.visible = true;
       grpTxtScrolls.visible = true;
+
+      // render optimisation
+      _parentState.persistentDraw = false;
```
EliteMasterEric, line 533: This is probably broken? Will need to test more thoroughly later.

Go into a song in Freeplay, exit to the Freeplay menu, then move back to the main menu. When this line is here you probably get a pink screen during the transition

EliteMasterEric (changes requested):

source/funkin/ui/freeplay/FreeplayState.hx
richTrash21: Tested this out, no pink screen. It's may also be that I'm missed something (?)

https://github.com/FunkinCrew/Funkin/assets/87835336/13e129d6-b639-467f-81ad-4bed9a8e05e9

Same test but exiting right after getting in Freeplay

https://github.com/FunkinCrew/Funkin/assets/87835336/5eb45a87-ff00-471f-b434-e09327312a52

EnterTheVoid-x86: It did break something, from what I can see the sticker transition is not rendering

richTrash21: I have used a transparent stickers mod for debugging "pink screen" issue, so there should not be any problems with sticker transition

EliteMasterEric:
tested on stream

PR #2533, opened 2024-05-17
update to 0.3.3

Eviethecoder:


CharlesCatYT:
the f- wait is this an accidental pr or-

EliteMasterEric:
- Pull request is vague; it touches a wide variety of files but does not elaborate on what it modifies and why. Distinct changes which are independent of each other should be divided into separate pull requests to improve the review process.

- You turned off asset redirection, that's a useful feature. Change your build configuration if you want to disable it.
- The commit ID you changed to on funkVis is wrong.
- Player and opponent are already accessible from PlayState (try `PlayState.instance.currentStage.getBoyfriend()` and `PlayState.instance.currentStage.getDad()`).
- I don't even understand the changes you made to the health icon, there is already a system for displaying Winning icons.

- Health bar colors are staying green and red, we have no plans to change that.
- As it's a gameplay change, ghost tapping shouldn't be an option, it should be either on or off. Additionally, your implementation of ghost tapping is completely broken, it just disables anti-mash entirely instead of querying if there are notes nearby first.
- The `-end` animation suffix is interesting but should either be re-implemented or made into its own stand-alone pull request.

Overall, I see only one useful change in this PR that does anything helpful (the `-end` animation). If you want to have it added in, please make it a separate PR, and do not include anything else.

Eviethecoder:
omg this was supposed to be for a fork i was messing with huh? im so sorry everybody

PR #2537, opened 2024-05-17
Add shader and color support for SustainTrail

Falkimore:
The drawTriangles function in SustainTrail omits the colorTransform and shader arguments, which makes changing those properties do nothing


Sword352 (approved):
YESSS

lemz1:
im pretty sure, we can remove the processedgraphic, if we just use the colortransform in drawTriangles

Kade-github:
oops I think I fixed this internally by accident i'm so sorry lol. we had an issue exactly how lemz described and I ended up fixing it the same way :(

PR #2539, opened 2024-05-17
[BUGFIX] Fix TitleState bugs + crash

gamerbross:
Fixes TitleState in general
- When going back to TitleState, the music gets muted
- Spamming enter when getting to TitleState crashed (wasn't initialized yet)
- No 1 second black screen when going back to TitleState


ninjamuffin99:
code looks good! will test to double check

PR #2540, opened 2024-05-17
[BUGFIX] Allow Volume Keys while watching Toy Commercial (AttractState)

gamerbross:
When pressing any volume key (mute, up or down) the toy commercial closes and was really frustrating


EliteMasterEric:
This made my friend so mad when it happened to them, thanks for contributing!

PR #2563, opened 2024-05-19
Center strumlines

EnterTheVoid-x86:
Self explanitory. Moves strumlines to the middle of the screen, fixing the odd offset.

### DIFF
--- a/source/funkin/util/Constants.hx
+++ b/source/funkin/util/Constants.hx
@@ -522,7 +522,7 @@ class Constants
   /**
    * The horizontal offset of the strumline from the left edge of the screen.
    */
-  public static final STRUMLINE_X_OFFSET:Float = 48;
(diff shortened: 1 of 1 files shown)


CharlesCatYT:
agreed but to be fair... the devs preferred the original offset way more, feels just right.

EnterTheVoid-x86:
I honestly don't really understand why the devs prefer the original offset. It just looks wrong

CharlesCatYT:
im just saying

image

EnterTheVoid-x86:
I might try to see if I can get this to be an option in the settings menu so the user can choose between fixed offsets or how it was originally

richTrash21:
Fun fact: there is actually a commented line in PlayState that does that!

image

EnterTheVoid-x86:
Currently working on adding an option in the settings for this. Will make a new pr when done.

PR #2564, opened 2024-05-19
[FEATURE] Add Ghost Tapping to the Preferences Menu, and an option to center the strumline

EnterTheVoid-x86:
image
image

This is a follow up to my original pull request that fixed the strumline offset so it was no longer off center, I noticed that ghost tapping support was toggle-able from the source code so I also added an option for that too.

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -46,6 +46,25 @@ class Preferences
   }
 
   /**
+   * If enabled, the strumline is moved to the middle of the screen.
+   * @default `false`
+   */
+  public static var centerStrumlines(get, set):Bool;
+
+  static function get_centerStrumlines():Bool
+  {
+    return Save?.instance?.options?.centerStrumlines;
+  }
+
+  static function set_centerStrumlines(value:Bool):Bool
+  {
+    var save:Save = Save.instance;
+    save.options.centerStrumlines = value;
+    save.flush();
+    return value;
+  }
+
+  /**
    * If disabled, flashing lights in the main menu and other areas will be less intense.
    * @default `true`
    */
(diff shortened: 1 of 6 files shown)


CharlesCatYT:
but they didnt want ghost tapping to be in anyone's hands, its all about difficulty, why dont you just make your own engine at that point

EnterTheVoid-x86:
> but they didnt want ghost tapping to be in anyone's hands, its all about difficulty, why dont you just make your own engine at that point

I'm not THAT knowledgable lmao

EnterTheVoid-x86:
Also, where exactly is the proof of your statement?

EnterTheVoid-x86:
> Also, where exactly is the proof of your statement?

PhantomArcade literally said themselves that Ghost Tapping was good...

EnterTheVoid-x86:
If they didn't want it to be in anyone's hands why would they even make it an option during compilation???

JackXson-Real:
Can you tell me what strum lines are I literally don't notice anything different

EliteMasterEric:
Ghost tapping is an option during compilation because it's been in testing, on and off.

Since this is a nuanced feature which requires a lot of playtesting and tweaking (I was messing around with it earlier and it was NOT behaving in a way I felt was intuitive or fair), it should be definitely implemented by us.

Furthermore, since it's a gameplay change which directly affects how the game performs scoring, it should ABSOLUTELY not be a toggleable option.

Oh yeah, plus the centered strumline is on purpose.

Marcosayfi:
can't you atleast add the ghostapping from mobile? why do they get ghostapping?

MightyTheArmiddilo:
> can't you atleast add the ghostapping from mobile? why do they get ghostapping?

It is significantly easier to accidentally hit the screen on mobile which can lead to unnecessary ghost misses. Also, the ghost tapping isn't exactly perfect yet iirc.

PR #2594, opened 2024-05-23
add simple PR auto labelling action

ninjamuffin99:
This should add a github action that automatically labels PR

Currently it:
- Labels `Documentation` to any PRs that change anything in the `docs/` folder, or any `.md` file.
- Labels `Haxe` to any PRs that change `.hx` files

We can do further config as we please, for now I think this one should be enough for the lil prototype 

NOTE: right now this aciton runs on `pull_request`, this is purely for review/testing. After approval, and before merging, we need to change it to `pull_request_target`. See [here](https://github.com/marketplace/actions/labeler#updating-major-version-of-the-labeler)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,12 @@
+# Add Documentation tag to PR's changing markdown files, or anyhting in the docs folder
+Documentation:
+- changed-files:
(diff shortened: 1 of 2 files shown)


EliteMasterEric (approved):
Very useful

Issue #2496, opened 2024-05-13
Implement a "Game is Outdated" message


aflacc:
## What is your suggestion, and why should it be implemented?
It would be nice if the game had a way of letting the user know that theres a newer version of the game available, and prompting them to download the newer version, I guess think kind of like how Psych Engine will have a message on startup if theres a new version I guess? The way the itchio page has versions labeled is different from in-game, so it isn't exactly super clear if you're using the latest version

ninjamuffin99:
There is code in there we used to have! Shouldn't be too hard to re-implement it! Then it's a matter of us updating it via the NG api. We could also potentially update and keep track of versioning via itch or github...

Starexify:
The `status: pending pull request` should be changed to planned or just pending triage (check https://github.com/FunkinCrew/Funkin/pull/4456#issuecomment-4049471794)

PR #2631, opened 2024-05-30
[FEATURE] Re-add out of date screen 

Ethan-makes-music:
This works on all platforms from my knowledge!

I re added the out of date screen referenced in #2496 and it may be pretty janky (sorry about that).
The way you can tell it the newest version is by changing the gitVersion.txt to the newest version. And it will show the screen once you press enter in the Title state if the versions do not match 

Also did more code clean up in this PR!

I'm probably missing a few instructions but a simple look through the code should kinda explain it.

If this PR ends up getting merged, you need to change the path of the gitVersion.txt to the one in the actual funkin repo instead of it being the path to my fork!

I have only been coding in haxe for a year so apologies :)


aflacc:
> re added

wait that was a thing?? what does it look like??

Ethan-makes-music:
> > re added
> 
> wait that was a thing?? what does it look like??

It kind of looks like the Kade and psych screen. It was unused to I just re-added it. Heres what it looks like!

image_2024-05-30_162601492

Ethan-makes-music:
Bro im so fucking stupid.. accidently just fucking discarded all 14 commits

PR #2633, opened 2024-05-30
[FEATURE] Add OutdatedSubState

lemz1:
i've noticed that there exists an `OutdatedSubState` which is never used, so i thought about implementing it.
it checks the  project.xml version of the current build and compares it to the project.xml version from the main branch.
this check is only done for desktop versions.

we could also use the latest commit hash from the main branch to compare against, but that be more annoying, so i didnt do the check that way.

(i have noticed that #2631 already implemented the `OutdatedSubState`, however i dont like that implementation, so im gonna keep this open)


EliteMasterEric:
As much as I generally like this implementation's use of XML parsing, I have an internal PR which removes the XML entirely in favor of another system which uses Haxe code instead.

The version check file should definitely be a separate file. The code should also definitely use the semantic version checking systems found in `funkin.util.VersionUtil`.

EliteMasterEric:
NOTE: This pull request is similar to #2720.

PR #2655, opened 2024-06-06
Recycle Credit Lines

lemz1:
this pr changes the behaviour of the credits line, to not be all created at the beginning, but rather that they get recycled once a new line appears on the screen.
this should improve memory usage, especially when backers entries are supported


CyndaquilDAC (approved):
LGTM

PR #2662, opened 2024-06-07
Update COMPILING, adding one step

An-enderman:
I added the step that says to download Git, because the command "hmm install" requires Git to work.


ninjamuffin99:
merged internally and will come out in the 0.4.1 patch :)

ninjamuffin99:
actually you know what, why am i dumb, i should just merge this now lol

PR #2683, opened 2024-06-08
Fix botplay sustain release bug

Hundrec:
Copied opponent's holdTimer code to the player's side to process hold note animations correctly in Botplay mode


EliteMasterEric:
Tested on stream and LGTM

PR #2684, opened 2024-06-08
[BUGFIX] Pause Camera Follow when pausing

gamerbross:
Yeah, i dont think it even needs a video
(Thx @Matriculaso for showing me this way to fix it, credit him not me!!)

it also uses `Constants.DEFAULT_CAMERA_FOLLOW_RATE` instead of 0.04


EliteMasterEric:
Tested on stream, worked great! I saw zero issues with this.

PR #2709, opened 2024-06-08
[BUGFIX] Leaving the Game Over while also Retrying

DMMaster636:
Before, you were able to leave the game over while also retrying, which could actually lead to a crash with the sticker transition sometimes.
This just fixes it by not letting the player leave after confirming the death.


NotHyper-474:
Perhaps this also fixes #2715?

DMMaster636:
> Perhaps this also fixes #2715?

Yep, it does, since it's also tied to the Game Over

ninjamuffin99:
i believe this is merged in ! i assumed github would auto close the issue, so i will have to! but I believe since i rebased them onto the branch it changed the commit ids, apologies!

PR #2713, opened 2024-06-08
[BUGFIX] Fixed memory counter displaying negative numbers

KarimAkra:
when the Gc memory usage exceeds what's around 2GB, the memory counter dies and display negative numbers which is due to the limit of Int32

Screenshot_20240507_223802_dev beefers vendetta

(Note for testing: easiest way to achieve this is by going through every stage in the game)

now the bytes are retrieved as a Float (on cpp targets only) so the memory counter will likely never show negative mem usage (unless you're playing on a nasa PC and got a huge memory leak lol?)

Screenshot 2024-06-09 011647

### DIFF
--- a/source/funkin/ui/debug/MemoryCounter.hx
+++ b/source/funkin/ui/debug/MemoryCounter.hx
@@ -36,7 +36,7 @@ class MemoryCounter extends TextField
   @:noCompletion
   #if !flash override #end function __enterFrame(deltaTime:Float):Void
   {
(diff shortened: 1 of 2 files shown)


NotHyper-474 (approved):
LGTM

tposejank (approved):

EliteMasterEric (approved):

EliteMasterEric:
approved

Issue #2715, opened 2024-06-09
Bug Report: Game crashes after pressing escape right after restarting the song on the chart editor


Prod-42:
### If you are playing the game in a browser, what site are you playing it from?

- [ ] [Newgrounds](https://www.newgrounds.com/portal/view/770371)
- [ ] [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below
- 
  - [X] Windows
- 
  - [ ] Mac
- 
  - [ ] Linux

## What version of the game are you using? Look in the bottom left corner of the main menu. (ex: 0.2.7, 0.2.1, shit like that)

0.4.0

## Have you identified any steps to reproduce the bug? If so, please describe them below in as much detail as possible. Use images if possible.

1. Open the chart editor
2. Test the chart ( Press enter or Shift+Enter)
3. Die while testing the chart
4. Restart the song when you get a game over
5. Press Escape right after the transition starts
6. Null Object

## Please describe your issue. Provide extensive detail and images if possible.

While testing a song on the chart editor it's possible to make the game crash after dying on the chart editor, you need to get a game over and restart the song, then you can press escape right after you restart, this makes game crash and get Null Object Reference.

## If you're game is FROZEN and you're playing a web version, press F12 to open up browser dev window, and go to console, and copy-paste whatever red error you're getting

[https://github.com/FunkinCrew/Funkin/assets/85413903/f7c3cad3-02a3-4d25-bf22-40a5eb3d5e74](https://github.com/FunkinCrew/Funkin/assets/85413903/f7c3cad3-02a3-4d25-bf22-40a5eb3d5e74)

[...]

PR #2718, opened 2024-06-09
[BUGFIX] When testing a chart, pressing R (or dying) then pressing ESC to go back to the chart editor, it no longer crashes

zacksgamerz:
When testing a song via the chart editor, dying (or pressing R), then pressing ESCAPE it crashes. This is because of `playstate.instance` returning `null`.

Fixes issue #2715

### DIFF
--- a/source/funkin/play/GameOverSubState.hx
+++ b/source/funkin/play/GameOverSubState.hx
@@ -276,7 +276,7 @@ class GameOverSubState extends MusicBeatSubState
     }
     else if (boyfriend != null)
     {
-      if (PlayState.instance.isMinimalMode)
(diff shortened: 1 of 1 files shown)


zacksgamerz:
Also added one more fix where you could quit gameoversubstate after confirming to restart

tposejank:
> Also added one more fix Now you cant quit gameoversubstate after confirming to restart

Sir that would be a duplicate of #2709

zacksgamerz:
> > Also added one more fix Now you cant quit gameoversubstate after confirming to restart
> 
> Sir that would be a duplicate of #2709

Oh! I'll revert it, my bad

EliteMasterEric:
I don't experience this bug on v0.4.1 so I assume it got fixed earlier.

PR #2729, opened 2024-06-09
Use preview Timings for FreePlay Song Preview

PurSnake:
This pr makes previewTimings not useless
and makes preview louder, I can’t hear a damn thing.

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -1891,6 +1891,7 @@ class FreeplayState extends MusicBeatSubState
     else
     {
       var potentiallyErect:String = (currentDifficulty == "erect") || (currentDifficulty == "nightmare") ? "-erect" : "";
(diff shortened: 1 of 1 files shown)


ninjamuffin99:
FlxPartialSound doesn't get audio based off millisecond/second timing, but rather it gets the audio based on bytes and byte positions. the 0.05 start and 0.25 end refer to getting all the bytes between 5% into the song, and 25% into the song. Getting the actual second based time from a slice of bytes is a bit more complicated than just using these values, since we parse the MP3 + OGG "frames" directly, which both could have varying bitrates or sample rates that you'd have to figure out.

AbnormalPoof:
I know this PR was closed 3 months ago, but this PR doesn't directly give FlxPartialSound the timings in milliseconds. Rather, it divides both `previewStart` and `previewEnd` by 10,000 to convert it to a decimal.

It is slightly inaccurate though, but that's likely due to FlxPartialSound's way of loading stuff

PR #2734, opened 2024-06-09
Dont SNAP camera on CLASSIC focus event

PurSnake:
If there are several CLASSIC events nearby, camera WILL JERK.
If you need to SNAP camera, then use INSTANT ease for it.


PurSnake:
oh for god sake

PR #2740, opened 2024-06-10
[FEATURE] Run the garbage collection process after purging cache.

MAJigsaw77:
By running the garbage collector after destroying some textures from the `previousCachedTextures` map, some memory will be freed.


tposejank (approved):
w

MaybeMaru (approved):

PR #2741, opened 2024-06-10
[BUGFIX] Only play `scrollMenu` when there's a real change

gamerbross:
Changes some `scrollMenu` play code to only play when last selection isn't the same as the new one. (like when opening freeplay or pause menus)
(also reduce the volume in options menu sound to match the others)


ninjamuffin99:
very nice! i've merged this internally and it will be out in the 0.4.1 patch :) thanks!

ninjamuffin99:
this was merged into public repo now, but i think since i rebased them the commit ids changed and github dont recognize it! so will close this but it's merged in, thanks!

PR #2742, opened 2024-06-10
[BUGFIX] Made freeplay use the metadata to get the instrumental suffix

AppleHair:
(I accidentally messed up the previous branch and had to make the PR again. Sorry about that)
Song previews in freeplay will now use the instrumental suffix from the current difficulty's corresponding song variation metadata instead of using the variation id as an instrumental suffix and checking only the "erect" variation.


EliteMasterEric:
I approve!

EliteMasterEric (approved):

PR #2767, opened 2024-06-12
[IDK WHAT TO CATAGORISE THIS] minor pause menu optimization

Burgerballs:
Tsk tsk tsk... variables dont need to be made for this!


EliteMasterEric:
This change results in an unmeasurable performance, and personally I find "if current entry is not equal to previous entry" makes intent more clear than "if change is not zero".

PR #2813, opened 2024-06-15
[DOCS] Make compiling guide more user friendly

Hundrec:
Moved “download Git” from the middle of the guide to the setup step 

Should prevent errors with Git before installing Git

### DIFF
--- a/docs/COMPILING.md
+++ b/docs/COMPILING.md
@@ -3,12 +3,13 @@
 0. Setup
     - Download Haxe from [Haxe.org](https://haxe.org)
     - Download Git from [git-scm.com](https://www.git-scm.com)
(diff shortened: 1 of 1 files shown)


trayfellow (approved):

NotHyper-474:
I think it could also be good adding that users should **not** download the repository from the Download ZIP button, since I've seen a lot of cases where people would do that and the compilation fail.

Hundrec:
Great suggestion! I've added the ZIP button warning in the Setup step and restructured sentences to be easier to read and follow.
Hopefully people run into fewer problems in the future!

EliteMasterEric:
Nice!

Merging directly into `main` so people get linked to this new documentation.

PR #2814, opened 2024-06-15
FunkinCamera improvements

cyn0x8:
there are a few things about `FlxCamera`/`FunkinCamera` id like to change...

---

for one, rotation!! you can move the camera frame around, or move around its world space with `scroll`. you can scale the camera frame, or scale its world space with `zoom`. you can rotate the camera frame... but there is no equivalent for rotating its world space as of right now.

using `angle` to rotate the camera also breaks any shaders applied to the camera. it resizes the bitmap used for the shader, often ending up larger than the actual bounds of the camera for some reason, and also lag spikes at every arbitrary resize?? i first encountered this while working with Psych Engine and spent a while trying to figure out the cause of this to no avail.

https://github.com/FunkinCrew/Funkin/assets/58795684/6f759e68-97a6-4804-b7d9-2560bcd92227

the bandaid fix i settled on was to scale the camera such that itll always fit the screen while rotated, and then use a shader for rotation, but this comes with its own downsides:
- the game has to render ~4x as many pixels (cuz its ~2x the size) which is prettyy bad for performance, especially with shaders.
- to look the same, shaders need modified tex coords + modified texture function to scale back down to fit screen space.

id rather not have to do this, though... the solution im pursuing is to rotate anything drawn to the camera internally during rendering with a `rotation` camera property.

---

second, i would like to change the shaking behavior such that it only affects the world space position during rendering than acting on the camera itself, which will get rid of the "black border" issue

### DIFF
--- a/source/funkin/graphics/FunkinCamera.hx
+++ b/source/funkin/graphics/FunkinCamera.hx
@@ -47,6 +47,9 @@ class FunkinCamera extends FlxCamera
 
   public var shouldDraw:Bool = true;
 
+  // The camera's world rotation in degrees.
+  public var rotation:Float = 15;
+
   // Used to identify the camera during debugging.
   final id:String = 'unknown';
 
@@ -182,6 +185,16 @@ class FunkinCamera extends FlxCamera
       ?shader:FlxShader):Void
   {
     if (!shouldDraw) return;
+
+    if (rotation % 360 != 0)
+    {
+      // rotate about camera center
+      final centerX:Float = width / 2;
+      final centerY:Float = height / 2;
+      matrix.translate(-centerX, -centerY);
+      matrix.rotate(rotation * Math.PI / 180);
+      matrix.translate(centerX, centerY);
+    }
 
     if ( switch blend
       {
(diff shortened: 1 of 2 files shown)


cyn0x8:
i would like to point out that i am not familiar at all to these inner workings of flixel and flash and how it renders stuff, soo im gonna be kinda stumbling my way through this lol
any insight/criticisms/suggestions/edits are appreciated

current issues with the state its in:
- `FlxBackdrop` does not rotate
<br><br>
- sustains also do not rotate
<br><br>

Raltyro:
For Sprite Culling, i think for this it need the function for detect isOnCamera to revamp
or easy way somehow make the camera resize its boundary for sprites

For Notes and FlxBackdrop, these uses drawTriangles, copyPixels, startQuadBatch, so FunkinCamera would need to override those both too

cyn0x8:
`flashSprite.rotation` is already the current behavior i think



but ty lots for the info!! ill experiment with it tomorrow

Raltyro:
Oh shit i meant either _scrollRect or canvas, i got confused w flashSprite mb!

cyn0x8:
well thats... unfortunate

image

ill have to close this and make a separate pr in [FunkinCrew/flixel](https://github.com/FunkinCrew/flixel) since it now requires direct edits to `FlxCamera`

PR #2820, opened 2024-06-15
[ENHANCEMENT/BUGFIX] Few animation editor bugfixes and improvements

gamerbross:
I know this editor isn't finished, but would be great to have this fixes meanwhile.

List of changes
 - The background now is a FlxBackdrop and it's scroll factors are set to 1, so it's not confusing between moving the character and moving the camera.
- Changed flixel-ui commit to latest one so the FlxUIDropDownMenu is fixed and works properly
- Changed the mouse behaviour a little bit so now clicking the animation dropdown doesn't move the camera and it just changes the click behaviour on click, so if you click the character and drag while passing through a ui element, it keeps dragging the character until you unclick. (just feels more comfortable)


EliteMasterEric:
> * Changed flixel-ui commit to latest one so the FlxUIDropDownMenu is fixed and works properly

I was investigating earlier and apparently this menu is the ONLY place in the game that uses `flixel-ui` instead of `haxeui` (the other places are unused imports). I'd like to remove that dropdown entirely at some point (replacing it with a dropdown in `offset-editor-view.xml`), but it's currently non-trivial to do that so I haven't bothered.

gamerbross:
it always shows up for me

EliteMasterEric:
I will review and approve this if it's good, but my point was that the dropdown should be part of the dialog on the left rather than floating in space.

EliteMasterEric (approved):

PR #2885, opened 2024-06-22
[FEATURE] Bopper Speed Event

anysad:
- This PR aims to make players control the speed of each prop in the stage that is a _Bopper_ type.
- Players can control: characters (boyfriend, girlfriend, opponent/dad) and other Boppers within the stage.

Video Showcase:

https://github.com/FunkinCrew/Funkin/assets/69017727/16589409-f08f-4985-9e70-86a4f5eb4f48

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -51,6 +51,7 @@ import funkin.play.notes.SustainTrail;
 import funkin.play.scoring.Scoring;
 import funkin.play.song.Song;
 import funkin.play.stage.Stage;
+import funkin.play.stage.Bopper;
 import funkin.save.Save;
 import funkin.ui.debug.charting.ChartEditorState;
 import funkin.ui.debug.stage.StageOffsetSubState;
@@ -191,6 +192,11 @@ class PlayState extends MusicBeatSubState
    * Gets disabled once resetting happens.
    */
   public var needsReset:Bool = false;
+
+  /**
+   * Map of the boppers that will reset their bop speed once the restart occurs.
+   */
+  public var resetBoppers:Map<Bopper, Int> = new Map<Bopper, Int>();
 
   /**
    * The current 'Blueball Counter' to display in the pause menu.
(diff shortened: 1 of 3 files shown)


source/funkin/play/PlayState.hx
```diff
@@ -192,6 +193,11 @@ class PlayState extends MusicBeatSubState
    */
   public var needsReset:Bool = false;
 
+  /**
+   * Map of the boppers that will reset their bop speed once the restart occurs.
+   */
+  public var resetBoppers:Map<Bopper, Int> = new Map<Bopper, Int>();
```
EliteMasterEric, line 199: This should certainly be part of `Stage.hx`, since all Boppers are either stage props or characters, and therefore are children of the stage.

You could also just reset bopping for all characters and all stage props on stage restart and not require a Map to store that data.

source/funkin/play/event/BopperSpeedEvent.hx
```diff
@@ -0,0 +1,85 @@
...
+
+  static final DEFAULT_DANCE_EVERY:Int = 1;
+
+  public override function handleEvent(data:SongEventData):Void
+  {
+    if (PlayState.instance == null || PlayState.instance.currentStage == null) return;
+
+    var bopperName:String = data.getString('bopper');
+
+    var speed:Int = data.getInt('danceEvery') ?? DEFAULT_DANCE_EVERY;
```
EliteMasterEric, line 32: We can just leave this as is for now, but in the private branch this has changed to a float which snaps to multiples of `0.25` rather than multiples of `1`.

EliteMasterEric (changes requested):
I was going to say this needed a rebase but I fixed it by fiddling with Git.

Had one idea for a refactor but otherwise I like this.

source/funkin/play/PlayState.hx
anysad: Moved the bop speed reset into `Stage.hx` as you requested alongside cleaning some code inside `resetStage` to cycle through every single character in the `characters` Map.

Hundrec:
Merge conflict, whoop!

AbnormalPoof:
This pull request has been closed for inactivity. It will be re-opened once you resolve any existing conflicts.

PR #2893, opened 2024-06-23
[BUGFIX/ENHANCEMENT] Sustain trail alpha fix

Itsgithubfellow:
Sustain trail alpha would always default to 1 regardless of what alpha value was defined. Simply moving the alpha value into `update()` fixes this issue. This is similar to pre-rewrite sustain trails which were also semi-transparent.

### DIFF
--- a/source/funkin/play/notes/SustainTrail.hx
+++ b/source/funkin/play/notes/SustainTrail.hx
@@ -125,8 +125,6 @@ class SustainTrail extends FlxSprite
 
     flipY = Preferences.downscroll;
 
-    // alpha = 0.6;
-    alpha = 1.0;
     // calls updateColorTransform(), which initializes processedGraphic!
     updateColorTransform();
 
@@ -151,6 +149,7 @@ class SustainTrail extends FlxSprite
       triggerRedraw();
     }
     previousScrollSpeed = parentStrumline?.scrollSpeed ?? 1.0;
+    alpha = 0.6;
   }
 
   /**


nebulazorua:
Why do it in the update? Also why remove the existing alpha = 1.0?

lemz1:
> I tried editing the existing `alpha = 1.0` and it did not change anything

in Strumline.hx there is this line: note.holdNoteSprite.alpha = 1.0;
this is why changing it in create didnt do anything
you would need to change this to 0.6

lemz1:
then you will have to look for all lines of code that change `holdNotesprite.alpha`. Probably most of it will be found in strumline and playstate

anysad:
> Just tested it out, it only applies the alpha when the note is being pressed

maybe you forgot to change the alpha in _buildHoldNoteSprite()_ function in Strumline.hx?
image

anysad:
https://github.com/FunkinCrew/Funkin/assets/69017727/627a117e-bd6a-4f75-8374-d84e1aeb4943

if you change it to the desired 0.6 alpha, then it becomes just like in the good old days

EliteMasterEric:
> which could be why @EliteMasterEric didn't reject this.

For the record, `status: pending triage` means I haven't looked into an issue, I have been just tagging issues as they come in because the issue template isn't doing that for me right now.

I messaged Dave about it but my recollection is that opaque holds were an intentional change. I will definitely merge the other PR that fixes the issue with mods changing opacity regardless.

EliteMasterEric:
Yeah the opacity change was intentional.

Lasercar:
When I was getting some clips for [this](https://github.com/FunkinCrew/funkin.assets/pull/111) yesterday, I thought that transparent hold notes looked way cooler than opaque ones. 

If the issue is that short hold notes are harder to see with the transparency, then maybe only make short holds opaque, but make everything else have the transparency? Something like this:
Screenshot 2025-01-29 162941
The only problem really with doing it like this is that it doesn't really account for scroll speeds.

PR #2894, opened 2024-06-23
Looping of character sing and strumline confirm animations on hold notes

Davvex87:
Now, whenever the player or CPU/NPC hit a hold note, the character corresponding to that player and the strumline which the note belongs to will now play their sing and confirm animations respectively on "loop".
This used to be a feature in funkin versions prior to 0.3 but has been lost due to the re-write of how the game renders, stored and uses hold notes, but with this re-implementation, this lost feature has been revived into the new engine, with the added bonus of animation loop rates not being song BPM dependent but instead being a constant value that determines at what frame of the animation it should loop back to the starting frame.

### Note:
At line 448 of `Strumline.hx`, the `holdConfirm` function has been commented out as its functionality was replaced with the new `playConfirmHold` function, however the function declaration for `holdConfirm` was left in the code-base as I do not have any idea if the presence of that function call is critical for the game's function, therefore anything related to that function was not tinkered with, but if decided that the function should no longer have a use in the code-base, I made sure that it could be removed without worry of breaking other functionalities.
image

https://github.com/FunkinCrew/Funkin/assets/102032123/22da1f99-a97d-4b03-95b8-a2ad9d6dc6fe

### DIFF
--- a/source/funkin/play/character/BaseCharacter.hx
+++ b/source/funkin/play/character/BaseCharacter.hx
@@ -346,6 +346,16 @@ class BaseCharacter extends Bopper
     }
   }
 
+  /**
+   * Stored the direction of the last note that was hit by this player.
+   */
+  var lastNoteDirection:Null<NoteDirection> = null;
+
+  /**
+   * Stores the time at which the player should stop looping the sing animation when pressing a hold note.
+   */
+  var lastHoldFinish:Null<Float> = null;
+
   public override function onUpdate(event:UpdateScriptEvent):Void
   {
     super.onUpdate(event);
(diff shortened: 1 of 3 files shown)


EliteMasterEric:
> This used to be a feature in funkin versions prior to 0.3

This used to be a BUG in Funkin' versions prior to v0.3.

MemTheBun:
Im 5km from your home Davvex87

tposejank:
I thought they would look better but this looks like they are skeletons

CyndaquilDAC:
> > This used to be a feature in funkin versions prior to 0.3
> 
> This used to be a BUG in Funkin' versions prior to v0.3.

Should've been a feature. It looks substantially better and more impactful than just leaving the characters on a still frame after the first notehit

Keoiki:
Why was this PR even made in first place? "-loop" animation suffix functionality is a thing now and this PR would've defeated the purpose of said animations, right? Just use "-loop" animations. (And the strum notes look better when the flash animation isn't played that fast on sustains.)

Davvex87:
The "-loop" sufix works differently, as when an animation is set to loop in HaxeFlixel, it will wait until the animation has reached its last frame and which it will restart the animation at the first frame, but they way I implemented it waits until the 4th frame is reached and once it does it will restart at the first frame, not taking into account the true last frame of the animation.

"But again, why not use the sufix?" you may ask, because certain characters with more that just 6 frames of animation waiting to loop will look unatural. Take Psychic or smooth BF (idk what its name is), they have like 20 frames, which will make it super weird!

Davvex87:
And yeah I also hate to see the characters look like a statue when singing a hold note, thats mostly why I made this Post Request.

PR #2902, opened 2024-06-24
Enable auto-pause for Video Cutscenes

AbnormalPoof:
A pretty small issue I noticed with Video Cutscenes is that they don't seem to pause when the game loses focus, even with the "Auto Pause" option enabled in the Options menu. Looking at the source code, it's kind of obvious why (autoPause is set to false).
Video Cutscenes will now pause on focus loss depending on if "Auto Pause" is enabled or not!

For HTML5, FlxVideo (the method of Video Playback HTML5 uses) doesn't seem to have an autoPause parameter like hxCodec does, so we're just manually pausing and resuming the Video in PlayState's `onFocus()` and `onFocusLost()` functions.

Another thing I'd like to mention is that there's a weird bug with autoPause where if you're on the Pause Screen and regain focus, the Video will resume in the background despite the game being paused. I took the liberty of fixing that, basically we just re-pause the video when we regain focus. It's not *that* elegant, but it does seem to work from the testing I've done.

Before:

https://github.com/FunkinCrew/Funkin/assets/86753001/9621dc72-60ce-43e8-ae33-e14d445793e5

After:

https://github.com/FunkinCrew/Funkin/assets/86753001/85262462-d9df-42d1-b426-45a2a3123b2f

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1301,12 +1301,18 @@ class PlayState extends MusicBeatSubState
     super.closeSubState();
   }
 
-  #if discord_rpc
   /**
    * Function called when the game window gains focus.
    */
   public override function onFocus():Void
   {
+    if (VideoCutscene.isPlaying() && FlxG.autoPause && isGamePaused) VideoCutscene.pauseVideo();
+    #if html5
+    else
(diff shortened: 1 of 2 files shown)


AbnormalPoof:
Accidentally merging into main... please hold.

PR #2913, opened 2024-06-25
Modcharting Tools

lemz1:
Feedback and tips are appreciated.

## ADDITIONS
- Cut sustain trail into multiple slices, which is useful for modcharting
- Add `NotePathModifier` interface for specifying the path that the note should follow

## TODO
- [x] Rework sustain trail rendering
- [x] Add sustain end trail back
- [x] Fix stuttering when hitting notes (For some reason calling triggerRedraw in the draw method seems to make it stutter)
- [x] Maybe add a "trail quality" option, which would determine how many segments the trail is cut into
- [x] Fix downscroll
- [ ] Fix chart editor hold
- [ ] Add modcharting tools (Modcharting will be done via scripts probably)
- [ ] Optimize the shit out of this (The performance doesn't seem to take a hit, but i will try to optimize regardless)

### assets/shared/images/NOTE_hold_assets.xml
```xml
<?xml version="1.0" encoding="utf-8"?>
<TextureAtlas imagePath="NOTE_hold_assets.png">
	<SubTexture name="blue hold end0000" x="157" y="0" width="50" height="64" />
	<SubTexture name="blue hold piece0000" x="105" y="0" width="50" height="87" />
	<SubTexture name="green hold end0000" x="261" y="0" width="50" height="64" />
	<SubTexture name="green hold piece0000" x="209" y="0" width="50" height="87" />
	<SubTexture name="purple hold end0000" x="53" y="0" width="50" height="64" />
	<SubTexture name="purple hold piece0000" x="1" y="0" width="50" height="87" />
	<SubTexture name="red hold end0000" x="365" y="0" width="50" height="64" />
	<SubTexture name="red hold piece0000" x="313" y="0" width="50" height="87" />
</TextureAtlas>
```

### assets/week6/images/weeb/pixelUI/arrowEndsNew.xml
```xml
<?xml version="1.0" encoding="utf-8"?>
<TextureAtlas imagePath="arrowEndsNew.png">
	<SubTexture name="blue hold end0000" x="28" y="0" width="7" height="6" />
	<SubTexture name="blue hold piece0000" x="19" y="0" width="7" height="6" />
	<SubTexture name="green hold end0000" x="46" y="0" width="7" height="6" />
	<SubTexture name="green hold piece0000" x="37" y="0" width="7" height="6" />
	<SubTexture name="purple hold end0000" x="10" y="0" width="7" height="6" />
	<SubTexture name="purple hold piece0000" x="1" y="0" width="7" height="6" />
	<SubTexture name="red hold end0000" x="64" y="0" width="7" height="6" />
	<SubTexture name="red hold piece0000" x="55" y="0" width="7" height="6" />
</TextureAtlas>
```


tposejank:
Fuck no, use shaders instead!

lemz1:
> Fuck no, use shaders instead!

If you are talking about using fragment shaders to do positional effects for sustain trails, then that is just not viable. We would need to create a new camera or bitmap for each sustain trail. Which would be memory and performance intensive.

If you are talking about vertex shaders, then we literally need to create more vertices to do positional effects.

tposejank:
I've seen the use of shaders in sustains before. But in my opinion, the adding of separated parts to the sustains may complicate it

lemz1:
Do you know what type of shaders they were? Meaning fragment or vertex. When people use shaders for sustain effects, i've always only seen fragment shaders. And from looking at their code, it was always kind of weird and limiting

tposejank:
Fragment is everywhere, I've never seen anyone use vertex. (in the end, everything came out of shadertoy)

lemz1:
Hmm, like I said, fragment shaders are limiting, because either you have to apply the effect to all sustain trails, or you need to create a new camera or bitmap for each sustain trails, which is going to be bad for memory usage. I could also imagine that it would be pretty costly for the performance, since you would be going through ~1million pixels for each trail, which is pretty bad in my opinion.

tposejank:
Yep. I do think they'll say it's best for the sustains to be one piece
only, might want to see when it is reviewed

KutikiPlayz:
what kind of modcharts specifically need segmented sustains?
I couldn't imagine a single thing that'd look remotely cool if you segment them again
just use shaders like tpose says
you shouldn't need an entire camera for the sustains anymore since they're one piece, you just need to apply the shader to all the sustains

lemz1:
when you use the shader on the sustain itself, the fragment shader will affect the graphic that is used for drawing. meaning you can do any positional effects

lemz1:
to demonstrate what i mean:
I used a shader and added 0.125 to uv.x

https://github.com/FunkinCrew/Funkin/assets/87707926/d6589c5c-e50f-45fd-a0fe-7524e4ee09e6

As you can see it doesnt change the position of the trail, rather it changes where to sample from the hold texture. the reason why i said that we need a camera is because the camera covers the entire screen and if we apply the 0.125 to uv.x then it would offset the entire graphic of the camera, which would be the expected behaviour.

lemz1:
closing for now, because i wont work on this for a while

anysad:
:(

lemz1:
reopening this, because i have an issue and maybe while im not working on it, someone can find a solution to it.

polymod error: Could not import class funkin.play.notes.modifier.NoteTransform

script:
```haxe
import funkin.play.notes.modifier.NotePathUtil;
import funkin.play.notes.modifier.NotePathModifier;
import funkin.play.notes.modifier.NoteTransform;
import funkin.play.song.Song;

class CoolModifier implements NotePathModifier {
	public function new() {

	}

	public function calculateTransform(time:Float):NoteTransform {
		return new NoteTransform(0, time);
	}
}

class HighSong extends Song
{
	var modifier:CoolModifier;

	function new()
	{
		super('high');
	}

	function onCreate(event:ScriptEvent) {
		modifier = new CoolModifier();
	}

	function onUpdate(event:UpdateScriptEvent) {
    for (note in NotePathUtil.getNotes()) {
			note.modifier = modifier;
    }
  }
}
```

tposejank:
Cannot import enums/abstracts/whatever in modules

tposejank:
hscript does support multiple classes in a single file, like class One
extends Module, class Two extends Module, etc, but not two class One and
class One

lemz1:
hmm, maybe its something else then.

currently i have an issue with this:
 modifier = new CoolModifier();

im not on my pc right now so i cant give you the exact error, but it was something along the lines of: target class null

lemz1:
ive tested it out with another script in base funkin:

i get an error here:
```haxe
var testClass = new TestClass();
```
error: Could not determine target class for "null" (unknown type?)


full script:

```haxe
import funkin.graphics.adobeanimate.FlxAtlasSprite;
import funkin.Paths;
import flixel.FlxG;
import funkin.audio.FunkinSound;
import funkin.play.character.MultiSparrowCharacter;
import funkin.play.GameOverSubState;

class TestClass {
	public var cool:Int;

	public function new() {
		cool = 5;
	}
	
	public function print() {
		trace(cool);
	}
}

class BoyfriendCharacter extends MultiSparrowCharacter {
	function new() {
		super('bf');
	}

	override function playAnimation(name:String, restart:Bool, ignoreOther:Bool) {
		if (name == "fakeoutDeath") {
			doFakeoutDeath();
		} else {
			super.playAnimation(name, restart, ignoreOther);
		}

		var testClass = new TestClass();
		testClass.print();
	}

	function doFakeoutDeath():Void {
		FunkinSound.playOnce(Paths.sound("gameplay/gameover/fakeout_death"), 1.0);

		var bfFakeout:FlxAtlasSprite = new FlxAtlasSprite(this.x - 440, this.y - 240, Paths.animateAtlas("characters/bfFakeOut", "shared"));
    FlxG.state.subState.add(bfFakeout);
		bfFakeout.zIndex = 1000;
    bfFakeout.playAnimation('');
		// We don't want people to miss this.
		FlxG.state.subState.mustNotExit = true;
    bfFakeout.anim.onComplete = () -> {
      bfFakeout.visible = false;
      this.visible = true;
			FlxG.state.subState.mustNotExit = false;
      this.playAnimation('firstDeath', true, true);
      // Play the "blue balled" sound. May play a variant if one has been assigned.
      GameOverSubState.playBlueBalledSFX();
    };
    bfFakeout.visible = true;
		this.visible = false;
	}
}
```

nebulazorua:
This feels like something fit for mods/frameworks to implement, and not implemented into base code itself

lemz1:
i never thought of making this a framework, but i guess that might just be better.

Issue #2822, opened 2024-06-15
Bug Report: Girlfriend no longer plays the sad animation when you lose a large combo.


AbnormalPoof:
## Describe the bug
As of the WeekEnd 1 update, Girlfriend no longer plays the sad animation when dropping a combo. Before, if you missed a note with a combo of 10 or higher, she would cry. Now, that animation seems to not work anymore.

## Screenshots/Video
Both videos are recorded on the HTML5 build of the game, but this also occurs on desktop as well.

- Pre-WeekEnd 1 (0.2.8)

https://github.com/FunkinCrew/Funkin/assets/86753001/564b60da-ce59-43f5-8b87-035489d15b51

- WeekEnd 1 (0.4.1)

https://github.com/FunkinCrew/Funkin/assets/86753001/c6e885d3-0e40-427c-90e9-536ddd3e4e4c

## Desktop
 - OS:
macOS
 - Browser
Firefox
 - Version:
0.4.1

## Additional context
Looking in GF's json file, there's an animation set for dropping a combo of 70.
![Screenshot 2024-06-15 at 5 47 33 PM](https://github.com/FunkinCrew/Funkin/assets/86753001/a076954a-9744-4291-9c6c-a2a753c31747)

(Nene also has similar entries!)
![Screenshot 2024-06-15 at 5 47 55 PM](https://github.com/FunkinCrew/Funkin/assets/86753001/638e1f11-b4ab-45c5-b4ca-1fa3b441531d)

In the source code, they're meant to play, but they aren't for some reason? I'm not sure why.
![Screenshot 2024-06-15 at 5 50 09 PM](https://github.com/FunkinCrew/Funkin/assets/86753001/355c851c-4c7f-42ab-b53b-54aee879543b)

Hundrec:
Just dug through the 0.2.x branch and found this:

![image](https://github.com/FunkinCrew/Funkin/assets/170126004/a1055b52-401a-4609-96a3-1e78241bcecf)

I had no idea the crying animation played with a combo of **6** or higher rather than 10.

Anyway, I'm going to attempt to reimplement the animation for combos of 70 or higher.

AbnormalPoof:
I think I ironed out the issue? I noticed that the drop code was using event.comboCount. I changed that to use Highscore.tallies.combo instead and that seemed to work! I also took the liberty of implementing the combo### animations as well. I'll make a PR with these changes later today.

EliteMasterEric:
[...]

PR #2919, opened 2024-06-27
Bandaid fix for GF's crying animation

Hundrec:
Addresses this bug report: https://github.com/FunkinCrew/Funkin/issues/2822

Temporarily reimplements Girlfriend's crying animation in `PlayState.hx`.
When a combo of 70 or more is broken, GF's 'drop70' animation plays.

This isn't an ideal long-term solution, so I've added some comments to suggest a way to fix the code in `BaseCharacter.hx`.
The main problem is that the GF combo drop animation code only runs when `onNoteMiss` is called. This means that combo breaks caused by Bad or Shit ratings won't cause the animation to play.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2673,6 +2673,11 @@ class PlayState extends MusicBeatSubState
     health += healthChange;
     if (isComboBreak)
     {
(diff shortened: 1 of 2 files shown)


Hundrec:
PR #2927 better addresses this issue.

PR #2972, opened 2024-07-06
[BUGFIX] Proper AnimOffsets for scaled Characters/Boppers

gamerbross:
Simply fixes applying animation offsets for scaled boppers/characters to not fly away. No side-effects, pixel week works perfectly.
(i would prefer #2310 solution, but if this is easier to approve im fine with that)


anysad (approved):

EliteMasterEric:
Superceded by a fix I made internally when working on #2310.

PR #2976, opened 2024-07-07
[BUGFIX] Fix manifest stuff messing up with example_mods folder

swordcube:
I was encountering this issue in another project using Polymod, and setting the mods folder type to "template" fixed the issue, So i thought i'd put the change here!

If i need to test with the actual Funkin repo, then I will (if i can get it to compile for the life of me)
It not compiling is why i made the commit on the GitHub website, I will try to test properly later


EliteMasterEric:
This seems great to have, although I think it makes sense to leave this in the Project.xml but commented out, because:

1) We probably don't want to include the example mods in release builds.
2) If you're building the game in `debug` mode without `-DGITHUB_BUILD` specified, it will automatically attempt to load its mods from `../../../../example_mods` instead of the `./mods` folder!

AbnormalPoof:

Hundrec:
Closing because Funkin' no longer uses `project.xml`

PR #2981, opened 2024-07-08
[BUGFIX] SustainTrail forced alpha

gamerbross:
This did nothing but screw scripts
(what was the decision behind changing the alpha of the sustain trails to 1? they looked better at 0.6 imo)


Hundrec:
I'm no coder, but I'm interested in this issue

How does this solution compare to #2893?

gamerbross:
> I'm no coder, but I'm interested in this issue
> 
> How does this solution compare to #2893 (and the comments under it)?

that makes it even worse, by always forcing the alpha to 0.6, not respecting the current design of alpha to 1

gamerbross:
> If you are removing the forced alpha, don't forget to also remove it from `buildHoldNoteSprite()` function from the same class. This is what handles the alpha pre-notehit.

if you remove it, press a sustain trail and drop it before it finishes, any next sustain trail which is recycled from that one, will be transparent causing problems. Even tho that didnt happen, doesnt really matter because there is no way to change the alpha before of that line

EliteMasterEric:
Quote from #2893:

> I messaged Dave about it but my recollection is that opaque holds were an intentional change. I will definitely merge the other PR that fixes the issue with mods changing opacity regardless.

EliteMasterEric:
I will approve this one though so the modders can configure things how they like.

PR #2986, opened 2024-07-09
[BUGFIX] Fixed `cancelMenu` sound not playing after switching state

AppleHair:
So in `StoryMenuState` and `OptionsState`, the `cancelMenu` sound doesn't play when coming back to `MainMenuState`, and in `MainMenuState`, the `cancelMenu` sound doesn't play when coming back to `TitleState`. This happens because `MusicBeatState`'s `startOutro` runs `FunkinSound.stopAllAudio`, which means every sound played before the call to `FlxG.switchState`, which calls `startOutro`, will be stopped. To fix this, instead of calling `FunkinSound.playOnce` BEFORE `FlxG.switchState`, we call it AFTER `FlxG.switchState`. Easy!


EliteMasterEric:
👍

Issue #2969, opened 2024-07-06
Bug Report: Hit window is offset from strumline


Dorentix:
basically when you hit what's on the left you get a "good" but when you hit what's on right you get "sick!!" (image recreated from memory)
![left good right sick](https://github.com/FunkinCrew/Funkin/assets/88308663/4b4b2be5-beb3-48bd-84b4-9df95cd5c5ce)

JackXson-Real:
I reported this in #2877 , they thought I was insane, but it’s real

JackXson-Real:
Also thanks for the visualization, it was hard to describe with only words

Hundrec:
Good to have confirmation on this issue. 
Can someone post a slowed down video that rigorously tests the Sick offset?

JackXson-Real:
> Good to have confirmation on this issue. Can someone post a slowed down video that rigorously tests the Sick offset?

I’ll make one tomorrow

DoReNtOs:
when I was editing the video I have noticed that this is a thing that not always happens,

https://github.com/FunkinCrew/Funkin/assets/88308663/6d3126af-6189-4b07-9fe5-45a1c093f705

game and video recorded in 144 fps

Hundrec:
Great demonstration, thank you!

Can you send the same video with the score counter showing? I'm curious how many points you're getting per note.

DoReNtOs:
here's fullscreen
https://github.com/FunkinCrew/Funkin/assets/88308663/b13a0fd0-feac-4e6d-8fd4-f245847ca9b0

DoReNtOs:
also the first image is not very accurate but it explains the point

Hundrec:
Thanks for the videos!

I just tested this and replicated the issue.

It looks to me like a visually perfect hit (arrow on arrow) is scored perfectly (500 points), but the window is tighter above the arrow and looser below the arrow. 

I must not have noticed this before due to my own input delay, but this should definitely be addressed for better input accuracy.

JackXson-Real:
I wonder if this happens because the strum line is visually positioned higher than the code for the hit windows checks for?

Hundrec:
> I wonder if this happens because the strum line is visually positioned higher than the code for the hit windows checks for?

This is probably the cause, but I have no idea what part of the code handles the placement of the hit window

If someone finds it, please send it here!

DoReNtOs:
[...]

PR #2987, opened 2024-07-09
Fix Judging For Notes When Using Input Offset

lemz1:
This pr fixes an issue where the notes are wrongly judged when using an input offset.

This #2969 describes also another the problem in detail, which is that seemingly that above the strum is a shorter hit window than below. This is not yet fixed. (Im not sure if we are all just crazy)

To demonstrate the issue more clearly im using extremely large offsets.
Using:
Input Offset: 160ms
Visual Offset: 0ms

### Before

https://github.com/FunkinCrew/Funkin/assets/87707926/1a378317-e739-4947-ba71-36868366e17c

### After

https://github.com/FunkinCrew/Funkin/assets/87707926/c246b8f7-a974-4e18-ac8d-ff8062ecd809

### DIFF
--- a/docs/COMPILING.md
+++ b/docs/COMPILING.md
@@ -5,8 +5,9 @@
     - Download Git from [git-scm.com](https://www.git-scm.com)
     - Do NOT download the repository using the Download ZIP button on GitHub or you may run into errors!
     - Instead, open a command prompt and do the following steps...
-1. Run `git clone --recurse-submodules https://github.com/FunkinCrew/funkin.git` to clone the repository with the necessary assets submodule
-    - _If you accidentally cloned without the `assets` submodule (aka didn't follow the step above), you can run `git submodule update --init --recursive` to get the assets in a foolproof way._
+1. Run `git clone https://github.com/FunkinCrew/funkin.git` to clone the base repository.
+2. Run `git submodule update --init --recursive` to download the game's assets.
+    - NOTE: By performing this operation, you are downloading Content which is proprietary and protected by national and international copyright and trademark laws. See [the LICENSE.md file for the Funkin.assets](https://github.com/FunkinCrew/funkin.assets/blob/main/LICENSE.md) repo for more information.
 2. Run `haxelib --global install hmm` and then `haxelib --global run hmm setup` to install hmm.json
 3. Run `hmm install` to install all haxelibs of the current branch
 4. Run `haxelib run lime setup` to set up lime
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2437,7 +2437,7 @@ class PlayState extends MusicBeatSubState
 
     // Get the offset and compensate for input latency.
     // Round inward (trim remainder) for consistency.
-    var noteDiff:Int = Std.int(Conductor.instance.songPosition - note.noteData.time - inputLatencyMs + Conductor.instance.inputOffset);
+    var noteDiff:Int = Std.int(Math.abs(Conductor.instance.songPosition - note.noteData.time) - inputLatencyMs + Conductor.instance.inputOffset);
 
     var score = Scoring.scoreNote(noteDiff, PBOT1);
     var daRating = Scoring.judgeNote(noteDiff, PBOT1);
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2437,7 +2437,7 @@ class PlayState extends MusicBeatSubState
 
     // Get the offset and compensate for input latency.
     // Round inward (trim remainder) for consistency.
-    var noteDiff:Int = Std.int(Math.abs(Conductor.instance.songPosition - note.noteData.time) - inputLatencyMs + Conductor.instance.inputOffset);
+    var noteDiff:Int = Std.int(Conductor.instance.songPosition - note.noteData.time - inputLatencyMs + Conductor.instance.inputOffset);
 
     var score = Scoring.scoreNote(noteDiff, PBOT1);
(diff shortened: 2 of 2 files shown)


Hundrec:
Nice and easy fix!

Does this provide more insight into the offset-by-default hit window described in the issue?

lemz1:
> Does this provide more insight into the offset-by-default hit window described in the issue?

you mean the issue where, it seemed that the hit window above was shorter than below?

Hundrec:
Yep!

lemz1:
i haven't tested it out

lemz1:
no, i don't think it fixes that issue, still seems like below has a larger hit window than above

Hundrec:
That makes sense, I expected the issue to remain.

I think a similar change could be made to address the offset window somewhere in the same file!

lemz1:
It might be fixed now, i cant really tell.
The reason for using the abs of the difference, is that the inputLatencyMs seems to be always positive, no matter if you hit it early or later. However because it is like that we need to ensure consistency with the signs.

anysad (approved):

lemz1:
Also another reason why it seems as if the hit window is shorter above, is because when `songPosition - strumTime` is equal to 0, the notes are a tiny bit below the strum, like 10 pixels

lemz1:
Here's a screenshot of that

screenshot-2024-07-09-23-19-27


> This might be more appropriate in a separate PR, but I think we could probably just bump the notes up by the difference (if we can confirm it's there)

We could, but maybe the system for note positioning needs to be reworked in general. It seems overly complicated with all the weird variables like INITIAL_OFFSET, NUDGE, etc.

Hundrec:
Good screenshot and good point!

I also noticed the notes seem to be a tiny bit left of the strumline as well.

lemz1:
Here is a demonstration of the the sick hit window
45 is the sick threshold

`songPosition - strumTime = -45`
screenshot-2024-07-09-23-35-27

`songPosition - strumTime = 45`
screenshot-2024-07-09-23-38-01

lemz1:
> Out of curiosity, how are you pinpointing these values? Are you hitting the notes and getting a perfect score or something else?

i replace strumTime:
```haxe
note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(note.strumTime, vwoosh);
```

with songPosition:
```haxe
note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(Conductor.instance.songPosition, vwoosh);
```

in the case of the sick threshold i did:

```haxe
note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(Conductor.instance.songPosition + 45, vwoosh);
```
and
```haxe
note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(Conductor.instance.songPosition - 45, vwoosh);
```

lemz1:
Or did you mean, how i know what time they are. If so in Scoring.hx there is this part:
```haxe
  /**
   * The time within which a note is considered to have been hit with the Sick judgement.
   * `~25% of the hit window, or 45ms`
   */
  public static final PBOT1_SICK_THRESHOLD:Float = 45.0;

  /**
   * The time within which a note is considered to have been hit with the Good judgement.
   * `~55% of the hit window, or 90ms`
   */
  public static final PBOT1_GOOD_THRESHOLD:Float = 90.0;

  /**
   * The time within which a note is considered to have been hit with the Bad judgement.
   * `~85% of the hit window, or 135ms`
   */
  public static final PBOT1_BAD_THRESHOLD:Float = 135.0;

  /**
   * The time within which a note is considered to have been hit with the Shit judgement.
   * `100% of the hit window, or 160ms`
   */
  public static final PBOT1_SHIT_THRESHOLD:Float = 160.0;
```

Hundrec:
> > Out of curiosity, how are you pinpointing these values? Are you hitting the notes and getting a perfect score or something else?
> 
> 
> 
> i replace strumTime:
> 
> ```haxe
> 
> note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(note.strumTime, vwoosh);
> 
> ```
> 
> 
> 
> with songPosition:
> 
> ```haxe
> 
> note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(Conductor.instance.songPosition, vwoosh);
> 
> ```
> 
> 
> 
> 
> 
> in the case of the sick threshold i did:
> 
> 
> 
> ```haxe
> 
> note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(Conductor.instance.songPosition + 45, vwoosh);
> 
> ```
> 
> and
> 
> ```haxe
> 
> note.y = this.y - INITIAL_OFFSET + calculateNoteYPos(Conductor.instance.songPosition - 45, vwoosh);
> 
> ```
> 
> 
> 
> This works because for the calculation it does `songPositon - time`. Meaning when time is set to songPosition, it will always equal 0, which is why i can add specific millisecond offsets that i want to test

This is a very clever debugging method! Is it possible for you to use this method to manually adjust the notes to perfect alignment?

lemz1:
> Is it possible for you to use this method to manually adjust the notes to perfect alignment?

what do you mean with perfect alignment?

Hundrec:
> Also another reason why it seems as if the hit window is shorter above, is because when `songPosition - strumTime` is equal to 0, the notes are a tiny bit below the strum, like 5-10 pixels

I mean changing the positions of the colored notes to match the gray notes at the center of the hit window. This might resolve the offset hit window issue!

lemz1:
this is something that probably shouldnt be handled by adjusting the time calculation. 
it would be better to use width and height calculations to perfectly center them.
maybe it has something to do with `NUDGE`, but idk

Hundrec:
I'll tinker around to see if I can move the arrows. Thanks for the info!

PR #2994, opened 2024-07-10
 [ENHANCEMENT + BUGFIX] soft codable visualizers + polymod download fix

MidyGamy:
This make visualize soft modable so nene.hxc now use it (making ABotVis.hx useless)

It also bring some other bug fixes related to the visualizer during the countdown, the shader un phillyStreets and missplaced offset for nene

Should be paired with https://github.com/FunkinCrew/funkVis/pull/7 for better results (and it will fix that last bar on Abot)


MidyGamy:
What does the "large" label means?

MidyGamy:
> What does the "large" label means?

Nvm i know now

MidyGamy:
Btw, you might wanna see 
https://github.com/FunkinCrew/funkVis/pull/7
It complement well with this PR

MidyGamy:
yeah, also with [that pull request](https://github.com/FunkinCrew/funkVis/pull/8) but you are already reviewing it...

AbnormalPoof (commented):
This PR looks good and works well, but my attention is caught in `nene.hxc`, this is something I encountered during testing:

Line 385:
```haxe
    #if desktop
    // On desktop it uses FFT stuff that isn't as optimized as the direct browser stuff we use on HTML5
    // So we want to manually change it!
    analyzer.fftN = 256;
    #end
```
This code will likely never run since `#if desktop` is a condition used at compilation. HScript interprets code instead of actually compiling it. This doesn't throw a syntax error, but it doesn't actually seem to detect if it's running on desktop, so the visualizer in WeekEnd 1 is different as a result.

A solution I figured out is using `PlatformUtil` to get the host platform:
```haxe
import funkin.util.PlatformUtil;
import funkin.util.HostPlatform;

var platform:HostPlatform = PlatformUtil.detectHostPlatform();

if (platform != HostPlatform.HTML5) {
	// On desktop it uses FFT stuff that isn't as optimized as the direct browser stuff we use on HTML5
	// So we want to manually change it!
	analyzer.fftN = 256;
}
```
This makes the visualizer look how it was before!

(Side note: Since ABot is now softcoded, `ABotVis.hx` is redundant and should probably be deleted.)

MidyGamy:
ok let me do it

MidyGamy:
Ok done

MidyGamy:
Well the 2 mentioned pull requests of [funkVis](https://github.com/FunkinCrew/funkVis/) have been merged... now waiting to see if this one will be accepted

MidyGamy:
help, how do I transfer it to another branch of mine?

moxie-coder:
> help, how do I transfer it to another branch of mine?

simply hit the edit button, and then you can change branches
IMG_2556

MidyGamy:
> > help, how do I transfer it to another branch of mine?
> 
> simply hit the edit button, and then you can change branches IMG_2556

image
No I only can change the branch I want to merge into, not the branch I want to use

moxie-coder:
> > > help, how do I transfer it to another branch of mine?
> > 
> > 
> > simply hit the edit button, and then you can change branches IMG_2556
> 
> image No I only can change the branch I want to merge into, not the branch I want to use

you have to make an new PR then

MidyGamy:
Or Ill just wait until it's merged or denied...

moxie-coder:
> Or Ill just wait until it's merged or denied...

you can easily close it yourself
IMG_2576

MidyGamy:
Yeah I tried to edit nene.hxc to render the visualizers with cutscenes musics... And for some reasons I get a null object reference when calling the analyzer.getLevels()...

AbnormalPoof:
I think it's because FlxG.sound.music doesn't exist (yet) during the cutscene.

MidyGamy:
Im not using FlxG.sound.music + I codes a safe state in case the sound is inexistent...

MidyGamy:
image
image
Yeah verry helpful...

MidyGamy:
I found the problem, the visualizer only works with FlxG.sound.music
image

MidyGamy:
I DID IT!!!
https://x.com/i/status/1828892194180092169

moxie-coder:
> I DID IT!!! https://x.com/i/status/1828892194180092169

that's nice

MidyGamy:
time to make another push request in the funkin viz repo

MidyGamy:
Well waiting for https://github.com/FunkinCrew/funkVis/pull/9 to be accepted before pushing the change

MidyGamy:
Working on separating the character sprite from the speaker sprite for 3 reasons:
- reduce sprite sizes (especially gf, gf-car and pico in week 7)
- make the speakers play the idle anim on beat regardless of the character anim
- being able to swap easily between speakers (like I applied abot to gf very easily) of put the speakers without a character, or even put a custom character without having to re-export speakers

AbnormalPoof:
In my opinion, I think it’d be better to have something like that be a part of a separate PR so it’s easier for the maintainers to review.

I think Eric explains it better:
7B45B947-8488-46FC-BD19-4AEB33CE221D

MidyGamy:
That was the idea

MidyGamy:
> In my opinion, I think it’d be better to have something like that be a part of a separate PR so it’s easier for the maintainers to review.
> 
> I think Eric explains it better:
> 7B45B947-8488-46FC-BD19-4AEB33CE221D
> 
> 
> Side note: I recommend re-adding this line in `nene.hxc` since it’s a feature added in 0.4.1:
> ```haxe
> animFrame = Math.round(animFrame * FlxG.sound.volume);
> ```
> 

About the 0.4.1 I actually disabled it cause it doesn't make really sense to me that this thing react to the volume... 1st in fictional universe, they aren't affected by the volume, 2nd, (I don't know if it's the case for everyone) I often reduce the volume of the game cause it's way too strong, and I don't want the visualizers to be affected by it...

I think the best solution would be to add an option for it

EliteMasterEric:
@MidyGamy The reason it actually exists is because the audio data provided to the game on web scales with game volume, and since we wanted to maintain parity, we just applied a multiplier on desktop which approximates the effect. In practice, I've found that people tend to turn down the in-game volume a lot, and with that the visualizer look significantly worse on both platforms.

The proper solution is to figure out the root cause and make the web visualizer look more like desktop. Simply multiplying the levels didn't seem to work for me.

MidyGamy:
Can't you just divide?

EliteMasterEric:
> Can't you just divide? (I mean if it's automatically multiplied by the volume, and if you divide back, if like just multiplying by 1... It's just mathematics...)

As mentioned, I tried simply dividing, but that didn't look accurate either. It'd take more experimenting to figure it out.

MidyGamy:
ok, and about the revision, what does it concern?

PR #3005, opened 2024-07-11
[BUGFIX] Default health icon (face) not used

gamerbross:
If the icon file wasn't found, the icon shows as HaxeFlixel logo instead of `face`. It was supposed to call the loadCharacter in the setter of `characterId` but was later changed and not working as intended.
image
I changed some other code that was using BF icon as default instead of face. Also in the previous way of checking if the icon exists, it was did twice, leading into a possible stack overflow (which don't show a log) if also the `face` icon image doesn't exist.


ninjamuffin99:
gamerbross ur so goated

EliteMasterEric:
LGTM

Issue #2837, opened 2024-06-17
Bug Report: Opponent stops singing on miss when bf has no vocals file


IDiMooo:
## Describe the bug

When there is only one vocal file for opponent and not for boyfriend, this audio will stop play on miss even if it's not bf's vocal file.

## To Reproduce

Create a song with only one vocal file for opponent and miss a note before opponent's turn.

## Expected behavior

The audio should still play if I understand why the vocals were seperated.

## Screenshots/Video

![image](https://github.com/FunkinCrew/Funkin/assets/163159763/762eb5e5-6b59-4996-abbe-d8db988ba77f)

https://github.com/FunkinCrew/Funkin/assets/163159763/60069027-27c1-4983-82a7-02741da4d09a

## Desktop
 - OS: Windows 10
 
 - Browser: nope
 
 - Version: 0.4.1
 

## Additional context

I was porting some mods and they have the old Voices files so I just wanted to give it to opponent so it won't mute and the miss sounds will be the only thing but I noticed that even if the voices are on the opponent's side they still get muted after miss. If both vocal files are there it mutes just bf's but if there is one for opponent it mutes it too. I don't know if it is intented but ye.

M7theguy:
This is the reason why they seperated the vocals. The mod you are playing only has one file for vocals of the opponent AND Boyfriend. A fix for this might be to check how many vocal files there are and shorten the miss effect

IDiMooo:
Yeah I get why they seperated the files it's just the fact that bf can mute opponents side in this case which i think may not be intended

PR #3012, opened 2024-07-12
[BUGFIX] Remove unnecessary player vocals muting

gamerbross:
ghost missing muted the player vocals, theres no point

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2525,7 +2525,6 @@ class PlayState extends MusicBeatSubState
           });
       }
     }
-    vocals.playerVolume = 0;
 
     if (Highscore.tallies.combo != 0) if (Highscore.tallies.combo >= 10) comboPopUps.displayCombo(0);
 
@@ -2588,7 +2587,6 @@ class PlayState extends MusicBeatSubState
 
     if (event.playSound)
     {
-      vocals.playerVolume = 0;
       FunkinSound.playOnce(Paths.soundRandom('missnote', 1, 3), FlxG.random.float(0.1, 0.2));
     }
   }


tposejank (approved):
this also broke songs with only bf vocal track merged.

AbnormalPoof (approved):
Hoping this gets merged since it gets very awkward on songs that do not have split vocals (Basically every mod)

https://github.com/user-attachments/assets/5e13d3a5-c875-4271-b7fb-2eeec62aba7e

anysad (approved):

KutikiPlayz:
what if, instead of fully removing this, you just unmute the vocals when the opponent sings for songs that have the vocals in one file, like how the game used to work

tposejank:
> what if, instead of fully removing this, you just unmute the vocals when the opponent sings for songs that have the vocals in one file, like how the game used to work

See the problem with that is the game expects people use separate voice tracks as it's more convenient, and isn't likely they want ot do that  instead of just easily not muting the bf vocal track when a ghost miss occurs.

AbnormalPoof:
> See the problem with that is the game expects people use separate voice tracks as it's more convenient
The game actually DOES look for `Voices.ogg` as a last resort. The vocal muting most likely wasn't accounted for that though.
Screenshot 2024-07-12 at 2 34 53 PM

KutikiPlayz:
this is my point, I don't mean do it all the time I mean specifically if the vocals are combined
better yet, just put your combined vocals as the opponent vocals and leave the bf vocals empty or missing
solves the problem ez pz

tposejank:
Indeed it does look, but it may soon feel like a deprecated feature

Hundrec:
> this is my point, I don't mean do it all the time I mean specifically if the vocals are combined
> 
> better yet, just put your combined vocals as the opponent vocals and leave the bf vocals empty or missing
> 
> solves the problem ez pz

#2837 reports that this doesn't solve the problem

EliteMasterEric:
Getting this PR to work such that it functions if the player misses on a single vocal track would be great for people looking to port mods.

tposejank:
> Getting this PR to work such that it functions if the player misses on a single vocal track would be great for people looking to port mods.

FNF Porter reference!!!

gamerbross:
> Getting this PR to work such that it functions if the player misses on a single vocal track would be great for people looking to port mods.

I've been working on a true solution for that. I have 2 solution proposals:
 - Unmute voices when the opponent hits a note (just when the game detects only player vocal track).
 Cons:
   - You will still hear the player voices in a duet part and will just sound with stutter.
   - Will keep the sound stutter in songs with a true player only vocals, like Camellia.

 - (i think its the best) An internal timer in the VoiceGroup object that resets to 0 everytime the player misses, when the timer is done (duration will be hold note lenghth, or for single notes 0.5s) unmute the player vocal track.
 Pros:
   - Less stutter than the other solution in duet parts.
   - Lyrics parts without notes don't get muted, like TwiddleFinger. (most of the PRs I've done lately it's because I've been porting it lol)
 Cons:
   - uhh nothing i think
   
---
@EliteMasterEric plz i need your opinion on these so i can make a cool pr

tposejank:
I dunnoooooo new property time??

EliteMasterEric:
I've been thinking about it, and maybe vocal muting should just straight up be disabled if there is no opponent vocal track.

**Pros:**
- This ensures the opponent's voice is never muted even during duets.

**Cons:**
- The player not muting on hit is inconsistent with the normal behavior, but with only one vocal track something will have to be inconsistent from normal anyway.

tposejank:
Exactly.

Hundrec:
Hey @gamerbross, did you mean to delete this branch?

gamerbross:
ummm not really i was just updating my fork and deleting old branches and I thought this PR was closed, I hadn't been following the development so I thought something was done about this. I will take a look later and maybe create another pull request with Eric's suggestions

PR #3032, opened 2024-07-15
[ENHANCEMENT] FileUtil overhaul + file and url sandboxing

cyn0x8:
solves (partially) #4737 

---

- `openUrl` in `WindowUtil` now performs sanitization
  - `http` or `https` only, or does nothing at all
- moved `openFolder` and `openSelectFile` from `WindowUtil` into `FileUtil`
  - `openFolder` now only opens to directories that exist on the filesystem, and `openSelectFile` now only opens to paths that exist on the filesystem
- adds the following new functions to `FileUtil`:
  <br>
  ```haxe
  moveFile(path:String, destination:String):Void
  deleteFile(path:String):Void
  getFileSize(path:String):Int
  pathExists(path:String):Bool
  fileExists(path:String):Bool
  directoryExists(path:String):Bool
  readDir(path:String):Array<String>
  moveDir(path:String, destination:String, ?ignore:Array<String>, strict:Bool = true):Void
  deleteDir(path:String, recursive:Bool = false, ?ignore:Array<String>):Void
  getDirSize(path:String):Int
  rename(path:String, newName:String, keepExtension:Bool = true):Void
  ```
- some changes to existing functions in `FileUtil`
  - "browse" functions now require the `onSelect` callback (why would you open a browse prompt and do nothing with it...?)
  - `doesFileExist` renamed to `pathExists` (not to be confused with `fileExists`)
  - reorganized a few functions
- sandboxed `FileUtil` by creating a separate `FileUtilSandboxed` class that `FileUtil` is aliased to for scripts to use, sanitizing the paths first and preventing them from leaving the game folder
  - also, functions in `FileUtilSandboxed` that have the capability to modify/delete are prevented from messing with certain game files and game folders and their contents
- both `FileUtil` and `FileUtilSandboxed` obey `nullSafety`
- file inputs and outputs are now properly closed in `FileUtil`
- various small changes for readability in `FileUtil`

> [!warning]
> this pr likely has quite a few breaking changes

---

example of sanitization:

```haxe
package testing123;

import StringTools;

import funkin.modding.module.ModuleHandler;
import funkin.modding.module.ScriptedModule;
import funkin.util.FileUtil;

class Test extends ScriptedModule {
    override public function new() {
        super("testing123.Test");
    }

    override public function onStateChangeEnd(event:StateChangeScriptEvent):Void {
        var workingDir:String = FileUtil.sanitizePath(null);
        for (input in [
            "./",                                     // "."
            "/absolute/path",                         // "./absolute/path"
            "relative/path",                          // "./relative/path"
            "test/..",                                // "."
            "test/../path",                           // "./path"
            "C:\\absolute\\path",                     // "./absolute/path"
            "C:\\test\\..",                           // "./"
            "test/////path",                          // "./test/path"
            "../../../../_etc/password.txt",          // "./_etc/password"
            "c:/test/c:\\path"                        // "./path"
            // "./_folder1/symlink_good/file.txt",    // "./_folder2/file.txt"
            // "./_folder1/symlink_EVIL/password.txt" // "."
        ]) {
            trace(StringTools.replace(FileUtil.sanitizePath(input), workingDir, "."));
        }
    }
}
```

---

todo in the future: figure out how to get "real" path of symlinked paths (`sys.FileSystem.fullPath` doesnt seem to resolve them?)


AbnormalPoof:
Wouldn't this theoretically allow mods to do malicious things? Unless somehow directories that are not Funkin's are restricted.

cyn0x8:
> Wouldn't this theoretically allow mods to do malicious things? Unless somehow directories that are not Funkin's are restricted.

mods can already do that with the functions already in `FileUtil` like going `../` repeatedly upwards out of the game root directory and then back down to some important file and overwriting it or something

though sandboxing to only the game root directory is definitely needed i think thats an issue for another pr

AbnormalPoof:
Fair enough! Sandboxing would be very nice.

AbnormalPoof (approved):

cyn0x8:
> Fair enough! Sandboxing would be very nice.

itd be an easy addition like using this in all the functions that have a path input

```haxe
/**
 * Prevent paths from exiting the root.
 * This is useful for preventing directory traversal attacks.
 *
 * @param path The path to sanitize.
 * @return The sanitized path.
 */
public static function sanitizePath(path:String):String
{
  path = path.replace('\\', '/');

  var parts:Array<String> = path.split('/');
  var sanitized:Array<String> = [];
  for (part in parts)
  {
    if (part == '.' || part == '')
    {
      continue;
    }
    else if (part == '..')
    {
      if (sanitized.length > 0) sanitized.pop();
    }
    else
    {
      sanitized.push(part);
    }
  }

  return sanitized.join('/');
}
```

however the problem is that theres quite a few places internally where they use the functions to do stuff outside of the root folder (unit tests/polymod asset redirect/etc)

source/funkin/util/FileUtil.hx
```diff
@@ -522,6 +522,90 @@ class FileUtil
...
+  }
+
+  /**
+   * Get a file's size in bytes. Max representable size is ~2.147 GB.
+   * Only works on desktop.
+   *
+   * @param path The path to the file.
+   * @return The size of the file in bytes.
+   */
+  public static function getFileSize(path:String):Int
```
lemz1, line 584: Couldn't we just return a `Float` which is a 64-bit floating point number, which leaves us with much bigger values than 2.147 GB

tposejank (approved):

source/funkin/util/FileUtil.hx
```diff
@@ -538,6 +622,145 @@ class FileUtil
...
+
+  /**
+   * Delete a directory, optionally including its contents, and optionally ignoring some paths.
+   * Only works on desktop.
+   *
+   * @param path The path to the directory.
+   * @param recursive Whether to delete all contents of the directory.
+   * @param ignore A list of paths to ignore.
+   */
+  public static function deleteDir(path:String, recursive:Bool = false, ?ignore:Array<String>):Void
```
tposejank, line 692: me when I `deleteDir('../')`

anysad: that's what we call... **a little trolling** 🤑🤑

cyn0x8: > me when I `deleteDir('../')`

yeahhh tbh its way too easy... might hold off on this pr until i also figure out how to sandbox file utils

cyn0x8: image

win

cyn0x8:
image
image

it works

tposejank:
Shouldn't it be better to extend FileUtilBase and override all functions including those that cause trouble or am i dumb ?

cyn0x8:
> Shouldn't it be better to extend FileUtilBase and override all functions including those that cause trouble or am i dumb ?

image

will do that rn lol

cyn0x8:
wait nvm static doesnt have inheritance

EliteMasterEric:
Pretty interesting!

Of note, you could change the names so that the one that gets used by the core game is still `FileUtil`, and the one that the scripts uses is called something like `FileUtilSandboxed`, and then you can add a Polymod import alias which replaces imports of `FileUtil` with `FileUtilSandboxed`.

MidyGamy (approved):

tposejank:
why isn't this merged yet..... grrrrr.....

cyn0x8:
> why isn't this merged yet..... grrrrr.....

this is a sizeable pr and also related to security stuff so im not surprised if it takes a while to review and merge

Keoiki:
Still hoping this gets merged into 0.6.0.

cyn0x8:
made a bunchh of changes; still need to test them all...
but its open for review if anyone else wants to as well

BlueColorsin (approved):

EliteMasterEric:
I plan to review and merge this as-is if it's functional, but one change to make for the future (to help with #4737) is to disallow URLs from `FileUtilSanitized.openFolder`, because this function can be used to access any URL (`openFolder("https://www.google.com/")` worked for me on Desktop)

cyn0x8:
> I plan to review and merge this as-is if it's functional, but one change to make for the future (to help with #4737) is to disallow URLs from `FileUtilSanitized.openFolder`, because this function can be used to access any URL (`openFolder("https://www.google.com/")` worked for me on Desktop)

noted
adding to pr asap

cyn0x8:
alsoo we need to figure out how to reliably resolve symlinks because currently the root can be exited via a relative symlink... 

`sys.FileSystem.fullPath` despite being documented with "symlinks will be followed and the path will be normalized" it wasnt able to resolve their true paths on my machine (windows 11)

EliteMasterEric:
> alsoo we need to figure out how to reliably resolve symlinks because currently the root can be exited via a relative symlink...

I think this is fine as long as the game doesn't provide a way to create a symlink. If it doesn't, exploiting this would require an existing symlink inside the Funkin' game directory.

Lasercar:
This was partially merged in 0.6.3, it's just missing the `fixes + stronger sandboxing` commit

cyn0x8:
idk why html5 was even trying to create a directory in the first place but this commit shouldd fix the issue

Hundrec:
This puts the PR into sort of a weird place since it's already in the game, but I'll make sure this commit gets reviewed!

Lasercar:
> This puts the PR into sort of a weird place since it's already in the game, but I'll make sure this commit gets reviewed!

_re_-reviewed

Hundrec:
The commit was made after the update, so it will require reviewing and merging for 0.6.4

Kade-github:
the new commits look fine, just gotta fix the conflicts and we should be good on this one.

PR #3035, opened 2024-07-16
FIXES CHARACTER PATH ONLY CHECKING IN SHARED

7oltan:
so im trying to make it so it will load characters from a week dir but its not letting me so i checked and saw this was the problem so i fixed it :3


moxie-coder:
I don't think this is a good idea since I don't think it checks in shared by default

anysad:
why not just place character assets in `shared/images/characters`?

anysad:
> I don't think this is a good idea since I don't think it checks in shared by default

it doesn't check in `shared` by default, it initially checks in `preload`

moxie-coder:
> > I don't think this is a good idea since I don't think it checks in shared by default
> 
> it doesn't check in `shared` by default, it initially checks in `preload`

I know, that's kinda why I think this might cause issues

7oltan:
> > > I don't think this is a good idea since I don't think it checks in shared by default
> > 
> > 
> > it doesn't check in `shared` by default, it initially checks in `preload`
> 
> I know, that's kinda why I think this might cause issues

why tho?

7oltan:
> why not just place character assets in `shared/images/characters`?

i want to add the characters in a week type dir to save memory and just to make images look cleaner

mayo78:
> > > I don't think this is a good idea since I don't think it checks in shared by default
> > 
> > 
> > it doesn't check in `shared` by default, it initially checks in `preload`
> 
> I know, that's kinda why I think this might cause issues

how would it be an issue? it checks first for preload, then for the level path, and then shared

EliteMasterEric:
Paths automatically checks which library has the file before using it, so this should be good.

I believe characters were originally IN the level folders before moving them out due to issues when characters were shared between levels.

source/funkin/play/character/AnimateAtlasCharacter.hx
```diff
@@ -131,7 +131,7 @@ class AnimateAtlasCharacter extends BaseCharacter
   {
     trace('[ATLASCHAR] Loading sprite atlas for ${characterId}.');
 
-    var sprite:FlxAtlasSprite = new FlxAtlasSprite(0, 0, Paths.animateAtlas(_data.assetPath, 'shared'));
```
PurSnake, line 134: For this move you should rewrite Paths atlas thing since it takes the atlas without checking the shared library

PurSnake (changes requested):

PurSnake (approved):

7oltan:
@EliteMasterEric so this may or may not have broken how animate atlas characters work, since paths for animate atlases uses getLib path and now that's it's null it ONLY checks in preload, and before that it only checked in shared, you should make it so the paths.animateAtlas takes the path and checks for its animate json file by normal getPath and then make it go back a few to the folder

EliteMasterEric:
> @EliteMasterEric so this may or may not have broken how animate atlas characters work, since paths for animate atlases uses getLib path and now that's it's null it ONLY checks in preload, and before that it only checked in shared, you should make it so the paths.animateAtlas takes the path and checks for its animate json file by normal getPath and then make it go back a few to the folder

I don't understand what you mean, if this breaks things then Blazin' would be unplayable.

7oltan:
> > @EliteMasterEric so this may or may not have broken how animate atlas characters work, since paths for animate atlases uses getLib path and now that's it's null it ONLY checks in preload, and before that it only checked in shared, you should make it so the paths.animateAtlas takes the path and checks for its animate json file by normal getPath and then make it go back a few to the folder
> 
> 
> 
> I don't understand what you mean, if this breaks things then Blazin' would be unplayable.

The characters in blazin' put "shared:" in the start of the path; but mods that don't before it would work for them and now their characters are broken image

mayo78:
kinda not a problem though? in normal flixel projects you already get assets from libraries in that way

PR #3037, opened 2024-07-17
[BUGFIX] Ensure the variation used for the next song is valid.

AppleHair:
When playing in story mode, `PlayState` assumes all the songs use the same variation, which is usually true, but not guaranteed. This PR makes `PlayState` use the `getFirstValidVariation` method to find the valid variation instead of using the `currentVariation` variable, which might not be valid.


EliteMasterEric:
Aside from the fact that Story Mode generally isn't used for variations, the issue here is that if you DO use a variation, then `getFirstValidVariation` will not be correct as it will try to use the default variation first.

The correct variation should be determined by first checking if the variation+difficulty is valid for that song, and using `getFirstValidVariation` only if variation+difficulty isn't valid.

Issue #2888, opened 2024-06-23
[BUG] Instrumentals and vocals desync when the song starts


CryptoLockedFry:
## Describe the bug

When starting a song on the desktop version (not the website) occasionally the instrumental and vocals can de-sync. This can be fixed by pausing and unpausing the game.

## To Reproduce

Start a song on desktop until it happens

## Screenshots/Video

https://github.com/FunkinCrew/Funkin/assets/173584212/46388c3d-01e8-4654-965a-a69bd72dad42

## Desktop
 - OS:
Windows 11
 - Version:
0.4.1

JackXson-Real:
this is fire

CryptoLockedFry:
> this is fire

fr

Hundrec:
This seems to be caused by poor performance.

Every time I start a song for the first time after launching the game on my laptop, there's a lag spike after "Go!" and the vocals desync.

I've developed muscle memory to restart the first song every time at this point!

AbnormalPoof:
This issue is addressed in PR #2706

Benefonde:
> This seems to be caused by poor performance.
> 
> Every time I start a song for the first time after launching the game on my laptop, there's a lag spike after "Go!" and the vocals desync.
> 
> I've developed muscle memory to restart the first song every time at this point!

I literally also just pause unpause right as a song starts

CryptoLockedFry:
> This seems to be caused by poor performance.
> 
> Every time I start a song for the first time after launching the game on my laptop, there's a lag spike after "Go!" and the vocals desync.
> 
> I've developed muscle memory to restart the first song every time at this point!

I don't think it's poor performance, it happens on my laptop with an i7 and a 4070 even if I'm running nothing else in the background

EliteMasterEric:
Should have been resolved in v0.5.

PR #2706, opened 2024-06-08
[BUGFIX] Fixed Song Synchronization

VirtuGuy:
Probably should've explained this when I first made the pull request, but this person explained it well.

image

PR #3058, opened 2024-07-23
Use chart inst offset for song resync + other resync fixes

Burgerballs:
resyncVocals() function in PlayState now makes use of instrumentalOffset.

This PR also halves the requirement for a resync, because 100ms is a big enough difference to greatly change timings in my personal opinion.

The resyncVocals() function now also resyncs the instrumental track the same way as the vocals.

Resyncing is checked every beat instead of every step.

This PR eliminates any and all resyncing issues players are plagued with, below is a before and after;

Before:

https://github.com/user-attachments/assets/7b822328-8787-425f-bdce-b942a3cc3156

After:

https://github.com/user-attachments/assets/379890b9-6d02-4738-8dac-90af3284a50f


Hundrec:
This finally fixes https://github.com/FunkinCrew/Funkin/issues/2888. Yay!

NotHyper-474:
Doesn't #2706 already do the desync fix?

Cartridge-Man (approved):

cyn0x8 (approved):

doggogit (approved):

VirtuGuy (approved):

EliteMasterEric:
Worked well for me when I tried it, and since it works for a bunch of other reviewers I'd say it's good.

Sword352:
Any reasons the tracks are being paused though? You can just set the sound's time and move on

VirtuGuy:
> Any reasons the tracks are being paused though? You can just set the sound's time and move on

You can't just set the sound's time and move on without pausing as that would cause issues getting both the instrumental and vocals synced.

tposejank (approved):

AbnormalPoof (approved):

NotHyper-474 (approved):

Burgerballs:
> Any reasons the tracks are being paused though? You can just set the sound's time and move on

Pausing both tracks at the same time stops the time variable from increasing while you're setting it, making it easier to resync.

EliteMasterEric:
> Pausing both tracks at the same time stops the time variable from increasing while you're setting it, making it easier to resync, and makes it more accurate.

I was under the impression `FlxSound.time` only increases during the `update()` function.

Burgerballs:
> > Pausing both tracks at the same time stops the time variable from increasing while you're setting it, making it easier to resync, and makes it more accurate.
> 
> I was under the impression `FlxSound.time` only increases during the `update()` function.

The time variable does *change* during the update function but it doesn't really *increase* if paused.

Issue #3066, opened 2024-07-27
Bug Report: Time Signatures, excluding 4/4, are wrong.


Keoiki:
### Issue Checklist

- [X] I have properly named the issue
- [X] I have checked the issues/discussions pages to see if the issue has been previously reported

### What platform are you using?

Itch.io (Downloadable Build) - Windows

### If you are playing on a browser, which one are you using?

None

### Version

0.4.1

## The amount of beats is incorrect outside of x/4 time.
This is an issue I've been trying to fix myself ever since I saw how weird the 3/8 grid looks like in the chart editor. 2 beats with 6 steps, divided into a beat with 4 steps and the second one with only 2 steps, seriously?
With the game's "4 steps per beat" constant, that's only 1.5 beats. The issue with this, is that **the time signature numerator always defines how many beats there are in a measure** (or a bar, if you'd like). In a 3/8 time, there should be 3 beats, not 1.5 beats.

There's a line in the Conductor class, which states that a 7/8 time has 3.5 beats and 14 steps:
https://github.com/FunkinCrew/Funkin/blob/9b483de1da9c1b72b02ec30f4d13923470cf32b9/source/funkin/Conductor.hx#L32
But, this is only half the amount of beats there should be! 14 steps is indeed the "correct" amount using the game's math to calculate the amount of steps: `7 / 8 * 4 * 4 = 14` (The math produces the correct amount of steps for x/4 times)
https://github.com/FunkinCrew/Funkin/blob/9b483de1da9c1b72b02ec30f4d13923470cf32b9/source/funkin/Conductor.hx#L298

This creates a problem, as the numerator states that there should be 7 beats, a step count of 14 would result in 2 steps per beat. However the game forces 4 steps into each beat, which causes the beat count to end at 3.5.
https://github.com/FunkinCrew/Funkin/blob/9b483de1da9c1b72b02ec30f4d13923470cf32b9/source/funkin/Conductor.hx#L286

## What actually counts as a beat?
As a result of the math in the Conductor class, the definition of a beat is a quarter note (as they are in x/4 time). However, stepping outside of x/4 time this changes, as beats are counted as defined by the denominator. Beats are:
* eighth notes in x/8,
* sixteenth notes in x/16 and
* half notes in x/2.

This is not taken into account when calculating beat and step lengths. Should steps always remain as sixteenths, this would cause the charting grid to look awful in x/16 time, as every step would be a beat.

[...]

PR #3067, opened 2024-07-27
[BUGFIX] Correct step lengths in x/4 time signatures.

Keoiki:
Without changing the BPM, `stepLengthMs` has a different a different value depending on the x/4 signature.
- Example: at 100 BPM, in x/4 time a beat is 600ms, so each step is 150ms, in 2/4 that was 300ms instead, 100ms in 6/4 and so on.
- This affected, for example, how often a Character plays their idle animation.

This PR simply fixes the wrong math.

This addresses the 3rd section of #3066

### DIFF
--- a/source/funkin/Conductor.hx
+++ b/source/funkin/Conductor.hx
@@ -152,13 +152,13 @@ class Conductor
   }
 
   /**
-   * Duration of a step (sixtennth note) in milliseconds. Calculated based on bpm.
+   * Duration of a step (sixteenth note) in milliseconds. Calculated based on bpm.
    */
   public var stepLengthMs(get, never):Float;
 
   function get_stepLengthMs():Float
   {
-    return beatLengthMs / timeSignatureNumerator;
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
How does this affect things on time signatures with other denominators (8/16)?

Keoiki:
This shouldn't have any issues with those denominators, here's actually a comparison with the 3/8 beat I've made and been using:
(Notice how the metronome doesn't align with Up (measure beat hits) and Left (other beat hits) notes in the Before clip.)

### Before:

https://github.com/user-attachments/assets/3ee13b2d-612e-4b9b-a666-27ace348483d

### After:

https://github.com/user-attachments/assets/974ec8b4-52d7-4b4a-a276-7245a273f8c1

Cartridge-Man (approved):
Dude thank god for this change.

Keoiki:
I see no reason to keep this open since #4770 exists.

Issue #2375, opened 2024-05-09
Bug Report: Hitsounds and metronome are delayed in the Chart Editor


OrbyyOrbinaut:
[weed]: <> (FILL THIS ISSUE THING OUT AS MUCH AS POSSIBLE)
[weed]: <> (OR ELSE YOUR ISSUE WILL BE LESS LIKELY TO BE SOLVED!)
[weed]: <> (DO NOT POST ABOUT ISSUES FROM OTHER FNF MOD ENGINES! I CANNOT AND PROBABLY WON'T SOLVE THOSE!)
[weed]: <> (GO TO THEIR RESPECTIVE GITHUB ISSUES AND REPORT THEM THERE LOL!)

[weed]: <> (ALSO MAKE SURE THAT YOU USE PROPER LABELS, IF YOU'RE RUNNING INTO COMPILER ISSUES, USE THE compiler issue LABEL!!!)

### If you are playing the game in a browser, what site are you playing it from?

[weed]: <> (Put an X in the [ ] thingies to fill out checkbox!)
[weed]: <> (something like [x] pretty much, don't screw up or you will look stupid)

- [ ] [Newgrounds](https://www.newgrounds.com/portal/view/770371)
- [ ] [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below
- - [x] Windows
- - [ ] Mac
- - [ ] Linux

### If you are playing the game in a browser, what browser are you using?

[weed]: <> (Again, put an x in the [ ] box!)

- [ ] Google Chrome (or chomium based like Brave, vivaldi, MS Edge)
- [ ] Firefox
- [ ] Safari

## What version of the game are you using? Look in the bottom left corner of the main menu. (ex: 0.2.7, 0.2.1, shit like that)

0.3.2

## Have you identified any steps to reproduce the bug? If so, please describe them below in as much detail as possible. Use images if possible.

Just seems like a general bug with the chart editor.

## Please describe your issue. Provide extensive detail and images if possible.

The hitsounds and metronome sound very delayed in the chart editor, see videos below:

Hitsounds:

https://github.com/FunkinCrew/Funkin/assets/77822077/b73f7c25-065a-463b-b272-6fb861f0e61f

Metronome:

https://github.com/FunkinCrew/Funkin/assets/77822077/7f99d29d-65e0-44ac-b418-ce3693e86c1c

EliteMasterEric:
[...]

PR #3081, opened 2024-08-01
[BUGFIX] Fix for strange ChartingState hitsounds offset, i hope

PurSnake:
Hitsounds had wierd offset, not they dont.
(Flixel thing, i guess)

_(Yes, I had to use clideo to compress the video, 10mb for video is CRAZY)_

**Before:**
https://github.com/user-attachments/assets/8fbe0f58-075d-42fe-94e4-0fd8d7f06f3c

**After:**
https://github.com/user-attachments/assets/9e70898f-08e2-40a5-a60e-327bcfda2373

Aannd DadBattle Erect with fixed hitsounds offsets, because it's charted normally:
https://github.com/user-attachments/assets/afe3a551-a7d4-435b-bb2e-bb232955308e

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -283,6 +283,11 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
   public static final WELCOME_MUSIC_FADE_IN_DURATION:Float = 10.0;
 
   /**
+	 * Hitsounds are played with a strange offset and this should not happen.
+	 */
+	public static final HITSOUND_OFFSET:Float = -13;
+
+  /**
    * INSTANCE DATA
    */
   // ==============================
@@ -6290,10 +6295,10 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
     {
       // Check for notes between the old and new song positions.
 
-      if (noteData.time < oldSongPosition) // Note is in the past.
+     if (noteData.time < (oldSongPosition - HITSOUND_OFFSET)) // Note is in the past.
         continue;
 
-      if (noteData.time > newSongPosition) // Note is in the future.
+    if (noteData.time > (newSongPosition - HITSOUND_OFFSET)) // Note is in the future.
         return; // Assume all notes are also in the future.
 
       // Note was just hit.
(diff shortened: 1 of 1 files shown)


Hundrec:
Big if true!

biomseed:
This just adds another offset of -13 to the sound. I don't think Eric will approve this but it's the right way either way if the sound effect itself is delayed without the team knowing.

lemz1:
When I look at the sound in audacity, it doesn't seem like the sound has an offset. Maybe this is an issue with how sounds are played in haxeflixel?

A picture of audacity:
Capture

PurSnake:
Okay, now OBS works so i can show difference
(
And yes, I had to use clideo to compress the video)

**Before:**
https://github.com/user-attachments/assets/8fbe0f58-075d-42fe-94e4-0fd8d7f06f3c

**After:**
https://github.com/user-attachments/assets/9e70898f-08e2-40a5-a60e-327bcfda2373

Aannd DadBattle Erect with fixed hitsounds offsets, because it's charted normally:
https://github.com/user-attachments/assets/afe3a551-a7d4-435b-bb2e-bb232955308e

lemz1:
> > A picture of audacity: Capture
> 
> The opponent one DEFINITELY has an offset that the player one does not, and it's caused by the sound effect starting up it seems, which the player one doesn't do because it may have been edited out of that one. This should not happen unless the devs accidentally slowed down the sound effect (which they can easily fix by just pitching the other sound to the same pitch without slowing down and using that instead) or they edited it out of the player one and forgot to do it for the opponent one. (For Eric if you read this, the delay is definitely an issue and DOES exist)

I guess the opponent one does have a small offset but if you look at the time, it is below 3ms. Im pretty sure that, this is not noticable.

biomseed:
> I guess the opponent one does have a small offset but if you look at the time, it is below 3ms. Im pretty sure that, this is not noticable.

Even if it's not the source of the problem, it should be fixed to begin with.

lemz1:
Maybe this is an issue with ogg/mp3 files. I can't really say for certain, but I remember reading something about ogg/mp3 files containing delay, because of the compression algorithm.

wav files shouldn't have that problem because they are uncompressed.

AbnormalPoof:
Quote from a comment in #2375:
> It seems to be an issue with the OpenAL Soft time change on Funkin's Lime fork. I tried the fork on a different non-FNF related project and it seemed to suffer the same issue

AbnormalPoof:

PurSnake:
Its shitty

PR #3092, opened 2024-08-05
[ENHANCEMENT/BUGFIX] Bopper auto speed check + BPM Change song event type

MidyGamy:
This fixes the bopper desyncing from each other, imitating the
`if(curBeat % 2 == 0)
dad.dance();`
from the 0.3.X of the game

### DIFF
--- a/source/funkin/Conductor.hx
+++ b/source/funkin/Conductor.hx
@@ -76,6 +76,17 @@ class Conductor
   public var onStepHit(default, null):FlxSignal = new FlxSignal();
 
   /**
+   * Signal fired when the current Conductor instance changes BPM.
+   */
+  public static var bpmChange(default, null):FlxSignal = new FlxSignal();
+
+  /**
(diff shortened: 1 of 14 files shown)


PurSnake (approved):

MidyGamy:
For the moment, I have to continue it...
It crashes with Week 7 and Weekend 1

MidyGamy:
submodule are pain

MidyGamy:
Now should work correctly

MidyGamy:
Yeah, I just made a branch combining my 2 pull requests... I never thought that would work so well.

MidyGamy:
So yeah, 2 bug fixes in one:
- the fact it plays too fast the idle twice (forced) after a sing pose
- sometimes the idle plays on odd beats instead of even ones
And it should find the optimal beat to dance on to be consistent if the idle last more than 2 beats

MidyGamy:
You should notice it here:
https://www.youtube.com/watch?v=o2e4R2bUNPk

PR #3112, opened 2024-08-19
[ENHANCEMENT] Window titles

IdealistCat:

No

Almost Every state has a Window title (not an issue but a QOL thing)

### DIFF
--- a/source/funkin/ui/MusicBeatState.hx
+++ b/source/funkin/ui/MusicBeatState.hx
@@ -14,6 +14,7 @@ import funkin.modding.events.ScriptEvent;
 import funkin.modding.module.ModuleHandler;
 import funkin.util.SortUtil;
 import funkin.input.Controls;
+import funkin.util.WindowUtil;
 
 /**
  * MusicBeatState actually represents the core utility FlxState of the game.
@@ -21,6 +22,7 @@ import funkin.input.Controls;
  */
 class MusicBeatState extends FlxTransitionableState implements IEventHandler
 {
+  var windowTitle = "Friday Night Funkin'";
   var controls(get, never):Controls;
 
   inline function get_controls():Controls
@@ -59,6 +61,7 @@ class MusicBeatState extends FlxTransitionableState implements IEventHandler
 
   override function create()
   {
+    WindowUtil.setWindowTitle(windowTitle);
     super.create();
 
     createWatermarkText();
--- a/source/funkin/ui/MusicBeatState.hx
+++ b/source/funkin/ui/MusicBeatState.hx
@@ -22,7 +22,7 @@ import funkin.util.WindowUtil;
  */
 class MusicBeatState extends FlxTransitionableState implements IEventHandler
 {
-  var windowTitle = "Friday Night Funkin'";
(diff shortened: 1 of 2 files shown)


JVNpixels (approved):

JVNpixels:
May be a bit unnecessary to the devs, but I think this is great!

IdealistCat:
I thought it would be cool

IdealistCat:
Thought it would be neat

Cartridge-Man:
ngl this reminds me of leather engine (i literally KNOW that nobody will know what this is at all lol)

source/Main.hx
```diff
@@ -38,6 +39,7 @@ class Main extends Sprite
     // We need to make the crash handler LITERALLY FIRST so nothing EVER gets past it.
     CrashHandler.initialize();
     CrashHandler.queryStatus();
+    WindowUtil.setWindowTitle("Friday Night Funkin'");
```
NotHyper-474, line 42: I don't think this is needed as this only ever runs once and game already sets the title to that by default.

AbnormalPoof (commented):
I suggest also adding Window Titles for the Animation Editor (`DebugBoundingState`) and the Input Offsets menu (`LatencyState`).

IdealistCat:
> ngl this reminds me of leather engine (i literally KNOW that nobody will know what this is at all lol)

I do

source/Main.hx
IdealistCat: ah

IdealistCat:
> I suggest also adding Window Titles for the Animation Editor (`DebugBoundingState`) and the Input Offsets menu (`LatencyState`).

good idea

Cartridge-Man:
>  I do

Good job then ig

IdealistCat:
> > I do
> 
> Good job then ig

:)

source/Main.hx
ayed-jake-Offical: is it like 
`Application.current.window.title = "Friday Night Funkin'";`
or they make a `public function setWindowTitle(titleWindow:String){Application.current.window.title = titleWindow;}` on class WindowUtil
well cool a idea !!!

ninjamuffin99:
i thought about something like this, however can you make sure this is compatible with OBS? I know there's a setting that can match a source to the window title (or same exe type), and I've had troubles with that for Spotify where it forgets and OBS isn't able to detect the window correctly!

IdealistCat:
> i thought about something like this, however can you make sure this is compatible with OBS? I know there's a setting that can match a source to the window title (or same exe type), and I've had troubles with that for Spotify where it forgets and OBS isn't able to detect the window correctly!

No Idea Honestly, never fucked around with OBS Scripting.

ninjamuffin99:
its not so much scripting, but adding the window as a capture source. essentially if someone is playing the game in windowed mode and spitting it out through obs

ninjamuffin99:
I think I will close this one, it's a cool idea but I think *stylistically* we'd just keep the window title nice and simple!

IdealistCat:
> I think I will close this one, it's a cool idea but I think _stylistically_ we'd just keep the window title nice and simple!

understandable

PR #3114, opened 2024-08-19
[ENHANCEMENT] Legacy Revert

IdealistCat:

Nope

QOL: reverts input to lgact for if its not detected. don't ask why I picked legacy

beep


Cartridge-Man:
I mean, I see the idea of what it does, like using legacy as a fallback but at that point I would just make the game just send an error message and then give a window notification saying it didn’t load the thing properly

HOWEVER, I would probably just change this to add an if statement asking if the currently not working input system is NOT the weekend 1 input, then it will switch to that

EliteMasterEric:
If you're switching the scoring function, it should spit an error in the console and return a value of 0 so you can immediately investigate the issue (i.e. the current behavior). If you set the scoring function, and it seems to work when you're in-game, it may take a while to notice the problem.

PR #3119, opened 2024-08-24
[ENHANCEMENT] small scriptable class additions / reorganization

cyn0x8:
- added scriptable blank `Object` class for custom objects (more convenient than having a module to set up anonymous objects)
- added scriptable `FlxBasic`s and `FlxObject`s
- moved `funkin` package scriptable classes out of `funkin.modding.base` and into their base class packages (and added import aliases for backwards compatibility)

### DIFF
--- a/source/funkin/graphics/ScriptedFunkinSprite.hx
+++ b/source/funkin/graphics/ScriptedFunkinSprite.hx
@@ -1,0 +1,8 @@
+package funkin.graphics;
+
+/**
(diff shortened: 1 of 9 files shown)


biomseed:
make sure to list deprecations in changelog @EliteMasterEric

cyn0x8:
the `Object` script class is not necessary if #4181 is implemented, which its planned to be...

should it stay in the pr?

AbnormalPoof (approved):
Thanks!

PR #3131, opened 2024-09-05
[ENHANCEMENT] let modules interact with each other during onDestroy

cyn0x8:

currently, interaction between modules in `onDestroy` is unreliable as they are currently removed from the module list *as* they are iterated through instead of *after*

the module priority currently cannot be used to fix this since map iteration is unordered

i have a use case right now where id like to retrieve a value in a module from another module (a bool used to conditionally manage a set of sprites), but trying to retrieve the value will randomly error because `ModuleHandler` will attempt to grab the module after it has been removed from the cache

while in this case i can make a workaround by setting a variable to that bool before the reload starts, there arent any workarounds for functions....

this pr simply just removes that removal during the iteration (`clear()` functionally does the same thing anyways)


AbnormalPoof (approved):

EliteMasterEric:
> (clear() functionally does the same thing anyways)

👍

AppleHair (approved):

PR #3139, opened 2024-09-08
chore: Update ScriptedFunkinSprite.hx

AbnormalPoof:
`ScriptedFunkinSprite.hx` erroneously references `FlxSprite` in the comments. This PR changes it so it properly references `FunkinSprite` instead.


Keoiki:
we love copy pasting and then forgetting to change comments

Issue #3323, opened 2024-09-16
Inconsistencies with notes visual


MidyGamy:
### Issue Checklist

- [X] I have properly named the issue
- [X] I have checked the issues/discussions pages to see if the issue has been previously reported

### What platform are you using?

Itch.io (Downloadable Build) - Windows

### If you are playing on a browser, which one are you using?

None

### Version

0.3.0-0.5.0

### Context (Provide images, videos, etc.)

(I might add screenshots later)

I noticed inconsistencies with notes and strum lines while playing and editing the source code of the game:
- Strum lines are offset (but I think it's intentional)
- Hold notes are 2 pixels offsets (i think it's due by the fact it's rendered using vertexes)
- Pixel hold notes are way too small and aren't pixel perfect (btw I noticed the previous thing by fixing the scale)
- Already mentioned in an other issues report but the lack of splashes for pixel notes (I know it's planned for addition)
- The space between pixel strum notes aren't pixel perfect (it's at 4 when it should be at 6 to be pixel perfect)

Lasercar:
This issue has been partially fixed, but still has to remain up because it's not completely fixed, _sigh_.

Hundrec:
whatever
go my `status: resolved`

PR #3351, opened 2024-09-17
Fix note and receptor scaling and note tail offsets

dombomb64:
~~#3323~~ (Actually wait only some of the things mentioned there are fixed, only the third bullet and maybe the second one, if you count my duct-tape fix of putting offsets for the hold notes to counteract the offset that I couldn’t find the cause of as an actual fix)

- Made note asset scaling easier to understand. Before, there was some math going on behind the scenes (the hardcoded scaling stuff was probably added before the JSON scaling was implemented), but now, it just uses the scale in the JSON as the scale to apply to the sprites.
- Added offset functionality to the note tail sprites. The attached JSON for the pixel art notes uses this to pretty much fix the tails' offsets.
- In turn, I was able to make the pixel art notes have a scale factor that doesn't have mixels. The tails' scale has also been updated to keep a consistent pixel size.

Note tails before adding offset functionality (but after fixing receptor and note scaling)
Note tails before adding offset functionality (but after fixing receptor and note scaling)

## NOTE:
If this pull request is accepted, these are the new JSON files for the note styles.
[funkin.json](https://github.com/user-attachments/files/17034954/funkin.json)
[pixel.json](https://github.com/user-attachments/files/17034973/pixel.json)

### DIFF
--- a/source/funkin/play/notes/StrumlineNote.hx
+++ b/source/funkin/play/notes/StrumlineNote.hx
@@ -85,7 +85,8 @@ class StrumlineNote extends FlxSprite
     noteStyle.applyStrumlineFrames(this);
     noteStyle.applyStrumlineAnimations(this, this.direction);
 
-    this.scale.set(noteStyle.getStrumlineScale());
+    var scale = noteStyle.getStrumlineScale();
+    this.scale.set(scale, scale);
     this.updateHitbox();
     noteStyle.applyStrumlineOffsets(this);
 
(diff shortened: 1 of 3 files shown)


dombomb64:
Forgot to mention that most of this code was written on the `main` branch and then transferred over to the correct one because I couldn't figure out how to get the `develop` branch to compile properly.

dombomb64:
I realize now that most of the stuff mentioned in the issue I linked isn’t fixed by this pull request except for the scale of the note tails, but I still think it should be merged because it fixes other things. Also, on the topic of the note tails, I’m starting to think that there should be a better fix to the offsets rather than just counteracting them in the json (although I do still want the functionality for the json offsets kept), but I can’t begin to wrap my head around how the rendering code for them works right now, so I don’t really know what’s causing them to be slightly offset.

PR #3356, opened 2024-09-17
[BUGFIX] Fix bug where FPS counter was always displaying on startup even when disabled

Lethrial:
N/A
Fixes a minor bug where the FPS counter object was added regardless of what the Debug Display option was set to.
https://github.com/user-attachments/assets/28119ba4-c710-43c2-870e-96fa8f1271e3


EliteMasterEric:
if the debug display option is ON, does it correctly display on startup?

PR #3357, opened 2024-09-18
[BUGFIX] Modding (Polymod): Make addCharacter actually set the characters type

Kade-github:
N/A
Fixes a call to addCharacter in where if you were to create a custom character in an hscript and then try to add it to the stage, it wouldn't have it's character type get set. Meaning it would not react to any notes being played.
N/A


Kade-github:
reproduced by doing
```haxe
character = CharacterDataParser.fetchCharacter("bf");
PlayState.instance.currentStage.addCharacter(character, CharacterType.BF);
```

PR #3364, opened 2024-09-18
[BUGFIX] Fixed VSCode Lime/Haxe Extensions Failing To Build Cache

KarimAkra:
recently Project.xml has been replaced with project.hxp
but the Lime extension still tries to build the cache using a Project.xml file so it results with this error
Screenshot 2024-09-18 092217

This config makes lime use project.hxp instead of Project.xml so it builds the cache just fine


KarimAkra:
WHY DID IT PULL A WHOLE DIFFERENT REPO

PR #3384, opened 2024-09-19
[BUGFIX] Chart Editor/Gameplay: Fixes countdown and audio artifacts with offsets

Kade-github:
Don't think anyone uses these so no.
Fixes an issue with offsets that would cause it to mass resync vocals, or go forward in time and back really quickly. Causing a lot of artifacts, lagging the game, and being unpleasant.

This also fixes the song starting before the countdown (due to the offset).

This shows blammed at +1 second of instrumental offset.

https://github.com/user-attachments/assets/021e5b06-f67c-491a-8e17-d77bddf9c8be
(2nd clip is accidently the normal version of blammed instead of hard, but the point stands)

### DIFF
--- a/source/funkin/Conductor.hx
+++ b/source/funkin/Conductor.hx
@@ -397,9 +397,12 @@ class Conductor
    */
   public function update(?songPos:Float, applyOffsets:Bool = true, forceDispatch:Bool = false)
   {
+    var currentTime:Float = (FlxG.sound.music != null) ? FlxG.sound.music.time : 0.0;
+    var currentLength:Float = (FlxG.sound.music != null) ? FlxG.sound.music.length : 0.0;
+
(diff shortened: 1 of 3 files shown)


Kade-github:
Draft rn as i'm still fixing some weird stuff

Kade-github:
squashed a bunch of commits to make it less spammy on commits

Kade-github:
The last commit was to kind of fix a nasty bug with offsets that causes vocals to be desynced, to fix this i made the tolerance for resyncVocals 5 instead of 100. Didn't seem to affect me that harshly, but it might on lower end devices. I tested on HTML5 and it went well so be aware.

Burgerballs:
i find it funny how my sync fix also fucks with the audio, epic pr tho!

PR #3397, opened 2024-09-20
[ENHANCEMENT] Move the week directory to be loaded from PlayStatePlaylist to StageData, alongside some other changes to StageDataProp

JustKolosaki:
N/A

Adds a new field to [StageData](./source/funkin/data/stage/StageData.hx), "directory", which is read by [LoadingState](./source/funkin/ui/transition/LoadingState.hx) and sets the current level of Paths to the field's value and then loads it. The default value is "shared".

Alongside it, it adds three new fields to StageDataProp, "angle" (Float, default value 0.0), "blend" (String, default value "") and "color" (String, default value "#FFFFFF"), which are read in [Stage](./source/funkin/play/stage/Stage.hx) and set to the corresponding StageProps. 

N/A

## Note
If this gets accepted, stage files would need to be updated to at least contain the "directory" field.

### DIFF
--- a/source/funkin/data/stage/StageData.hx
+++ b/source/funkin/data/stage/StageData.hx
@@ -20,6 +20,10 @@ class StageData
   @:optional
   public var cameraZoom:Null<Float>;
 
+  @:default("shared")
+  @:optional
+  public var directory:Null<String>;
+
   public function new()
   {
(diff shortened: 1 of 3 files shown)


EliteMasterEric:
These seem great! Will look into it later.

PR #3355, opened 2024-09-17
[BUGFIX] Disable input in CharSelect until stayFunky plays

ActualMandM:
## Does this PR close any issues? If so, link them below.
#3156

## Briefly describe the issue(s) fixed.
This PR introduces a new allowInput boolean to CharSelectSubState, false by default and is only set to true when we start playing stayFunky. This prevents the game from being able to run/cancel tweens on a currently null object, being FlxG.sound.music.

### Before

https://github.com/user-attachments/assets/faf4bbc3-e1a6-4e28-bd7d-7e002523de97

### After

https://github.com/user-attachments/assets/3d7587ef-781f-4b91-a5eb-93d6ba9cb5d9

PR #3439, opened 2024-09-23
CharSelect polish

MaybeMaru:
Fixed "duplicate" bf's when opening the char select screen.
Fixed memory leak at the unlock character sequence.
Fixed inputs being allowed at the unlock character sequence.
Fixed spam causing errors displaying the current character / nametag.
Made it remember your last selected character.
Optimized a bit by checking for changes only on key presses.

### Before
https://github.com/user-attachments/assets/47657dc5-25a0-4523-b6ae-a69f56683542

### After
https://github.com/user-attachments/assets/64a76a86-9786-4f6b-8b53-af36d392d33e

### DIFF
--- a/source/funkin/ui/charSelect/CharSelectPlayer.hx
+++ b/source/funkin/ui/charSelect/CharSelectPlayer.hx
@@ -10,11 +10,13 @@ class CharSelectPlayer extends FlxAtlasSprite implements IBPMSyncedScriptedClass
 {
   public function new(x:Float, y:Float)
   {
-    super(x, y, Paths.animateAtlas("charSelect/bfChill"));
+    super(x, y, Paths.animateAtlas("charSelect/lockedChill"));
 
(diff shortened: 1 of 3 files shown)


doggogit (approved):

AbnormalPoof:
> Fixed inputs being allowed at the unlock character sequence.

#3355 already handles this, but everything else seems good.

JVNpixels (approved):

ninjamuffin99:
some changes may have been fixed with #3457 although what was the specifics with the memory leak?

MaybeMaru:
> some changes may have been fixed with #3457 although what was the specifics with the memory leak?

Was refering to the slowdown when switching to a character for the first time since it has to load the full character atlas mid-frame, precaching characters on create fixes the issue
Ive seen a lotta prs fixing these issues tho so i should prob close this one

PR #3447, opened 2024-09-24
Show Outdated Mods

lemz1:
This pr adds a `OutdatedModsState`, only mods that have been detected as outdated the first time, will be shown.
If there are no outdated mods `InitState` will immediately switch to `TitleState`

screenshot-2024-09-24-20-25-24

### DIFF
--- a/.github/actions/setup-haxe/action.yml
+++ b/.github/actions/setup-haxe/action.yml
@@ -82,7 +82,8 @@ runs:
     with:
       run: |
         git config --global --name-only --get-regexp 'url\.https\:\/\/x-access-token:.+@github\.com\/\.insteadOf' \
-        | xargs git config --global --unset
+        | xargs -I {} git config --global --unset {}
+
         git config -l --show-scope --show-origin
         git config --global 'url.https://x-access-token:${{ inputs.gh-token }}@github.com/.insteadOf' https://github.com/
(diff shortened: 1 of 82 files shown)


charlesisfeline (approved):

AbnormalPoof:
I feel like for something like this, it might be better to use `PolymodErrorHandler`'s `showAlert` function. Something like this maybe:
Screenshot 2024-09-24 at 4 17 33 PM
I'm not against the current implementation, though.

lemz1:
> I feel like for something like this, it might be better to use `PolymodErrorHandler`'s `showAlert` function. Something like this maybe:
> 
> Screenshot 2024-09-24 at 4 17 33 PM
> 
> I'm not against the current implementation, though.

Yeah its probably better with the showAlert thing, but i wasn't sure if it would look good if alot of mods are outdated, so i just immediately went with haxe-ui, since it implements a scrollable view

EliteMasterEric:
I definitely prefer the `showAlert` method, and having it display every time rather than the first time so we don't need to write anything to save data.

lemz1:
> I definitely prefer the `showAlert` method, and having it display every time rather than the first time so we don't need to write anything to save data.

Updated the pr to use `showAlert`, as well as showing outdated mods always

AbnormalPoof (approved):
works

Hundrec:
Finally, Eric added popups for outdated mods!

Issue #3501, opened 2024-09-30
Bug Report: Submodules are broken.


AbnormalPoof:
### Issue Checklist

- [X] I have properly named my issue
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Compiled from GitHub Source Code

### Browser

None

### Version

Not applicable.

### Description (include any images, videos, errors, or crash logs)

Self-explanatory.
![Screenshot 2024-09-30 at 6 25 56 PM](https://github.com/user-attachments/assets/69da1281-6e5d-4ae3-a99c-357a667f3306)
![Screenshot 2024-09-30 at 6 25 44 PM](https://github.com/user-attachments/assets/9d2fe1e5-2238-44b5-a21d-d8cd53a4005a)

NotHyper-474:
Oops https://github.com/FunkinCrew/Funkin/commit/4a9ad536955b2c3b524b605b25bccfd6ebf1c158

Seems like although the names are pointing to the private repos the commit hashes still are from the public repos

EliteMasterEric:
Fixed

PR #3503, opened 2024-10-01
submod link fix

CrusherNotDrip:
you guys had it to the public sub modules for literally less than an hour :sob:

#3501


EliteMasterEric:
Fixed it already

PR #3506, opened 2024-10-01
Fix Pico's songs not appearing on Freeplay

ninjamuffin99:
Freeplay would default it's difficulty string to `normal` but pico variations are `normal-pico`, so it wouldn't show you any Pico remix songs until you went to Random and changed difficulties.

This just initializes the suffixed difficulty string with `-pico` (or any other non-bf character id) when we load up Freeplay, so the initial generateSongs() filters correctly


EliteMasterEric (approved):

7oltan:
thanks ninja muffin, you da goat

PR #3507, opened 2024-10-01
Fix the character select cursor lerp for lower framerates

### LINKED ISSUE #3505: Bug Report: Character select square completely breaks at framerates below 60FPS
### Issue Checklist

### Platform

### Browser

None

### Version

0.5.1

When playing at framerates below 60, the "aftereffect" of the character select UI gets extremely volatile.

https://github.com/user-attachments/assets/2d9f3b4b-83f1-4a21-8852-3613df85974e

### Steps to Reproduce

See above.

ninjamuffin99:
fixes #3505 

Updating the main cursor to use the `smoothLerp()` instead of `coolLerp()` seems to fix it, as the rest of the squares lerp towards the main yellow cursor.

I believe this bug is caused from the lerp overshooting and needing to recorrect itself using the old lerp method

### DIFF
--- a/source/funkin/ui/charSelect/CharSelectSubState.hx
+++ b/source/funkin/ui/charSelect/CharSelectSubState.hx
@@ -885,8 +885,8 @@ class CharSelectSubState extends MusicBeatSubState
     cursorLocIntended.x += cursorOffsetX;
     cursorLocIntended.y += cursorOffsetY;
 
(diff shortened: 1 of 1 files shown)


trayfellow:
Maybe all instances of the deprecated coolLerp function in the game should just be replaced with smoothLerp? Or would that still cause a bunch of issues

ninjamuffin99:
coolLerp interpolates in a different way than smoothLerp, smoothLerp interpolates in a given time, so for each instance you'd need to tweak some values to be able to match the lerp visually

PR #3513, opened 2024-10-01
Known bugs

IdealistCat:
Nope

I mean for one the players can know a little bit of some of the bugs they might find and won’t have to worry about reporting cause it can link to an issue on the repo

also the devs can kinda use it as a checklist lol


EliteMasterEric:
We already have the Issues tab and an internal tool to use as a checklist.

PR #3527, opened 2024-10-01
Quick-fix for hanging when exiting game

### LINKED ISSUE #3196: Bug Report: Title Screen Sys.exit(0) Halting Issue
### Issue Checklist

- [X] I have properly named the issue
- [X] I have checked the issues/discussions pages to see if the issue has been previously reported

### What platform are you using?

### If you are playing on a browser, which one are you using?

None

### Version

0.5.0

### Context (Provide images, videos, etc.)

The Title Screen itself can halt after pressing the `ESCAPE` Key to exit the game by calling the `Sys.exit(0)` method, but it instead just proceeds to halt the entire game itself; without exiting/closing the window instance.

image

### Steps to reproduce (or crash logs, errors, etc.)

TitleState.hx => Press ESCAPE: The re-production method.

NotHyper-474:
Fixes #3196
Lime 8.2.0 has an [issue](https://github.com/openfl/lime/issues/1803) where `Sys.exit` calls hang the game on Windows, however `lime.system.System.exit` doesn't suffer from the same problem while achieving the same result. This PR replaces the instances of the function as a temporary solution for the issue.


Hundrec:
Yes please! I crash every 5 minutes while bug-testing

NotHyper-474:
Finally had a chance to actually test this and this only worked for the call in TitleState LOL, on crash it still hangs.

ninjamuffin99:
one thing i do know, i believe on chart editor if you crash / close the game, it will run code to save your current chart backup, which could cause a delay when crashing / closing the game on chart editor specifically.

EliteMasterEric:
> Finally had a chance to actually test this and it only worked for the call in TitleState LOL, on crash it still hangs.

Looking into this branch and the Crash Handler was calling Sys.exit() lol

NotHyper-474:
> EDIT: Sys.exit, System.exit(), and no call all cause the game to freeze. I think we should hold off on merging this until we find a proper fix or workaround that resolves the issue in all cases.

Yeah I actually had also replaced it on Crash Handler but then I reverted it cuz that didn't work.

Ironically enough System.exit [also calls Sys.exit](https://github.com/openfl/lime/blob/dd79e5852a17eee424b0951ff9400c881a357e33/src/lime/system/System.hx#L212) so I'm not sure why only in that case it doesn't hang.

NotHyper-474:
Seems like someone came up with a better implementation so I'll close this

PR #2629, opened 2024-05-29
[BUGFIX?] Reset CWD before Preloader

NotHyper-474:
When running the game from the command line and the current working directory is not the executable's directory the Preloader will fail to find the files it needs, this PR aims to fix this problem. This could perhaps also fix #2444 although it seems to be kinda unrelated.

I know that ideally you'd use `lime run PLATFORM` to run the game without recompiling, but this would not be the case for when downloading the game, specially for Linux players that for whatever reason would like to run it through a terminal without having to change the CWD.

From what I tested it doesn't break anything. Here's a before and after of the problem:
 (Before was done in Windows cuz I'm too lazy to recompile, but it should be consistent regardless of the platform)

https://github.com/FunkinCrew/Funkin/assets/40342021/c0c9e1c3-9295-440d-8e5b-30c4c97d82d7

https://github.com/FunkinCrew/Funkin/assets/40342021/d3235fcb-ce9a-4b10-9b56-9fbc929d50ed

PR #3538, opened 2024-10-02
Revert "[BUGFIX] Reset CWD before Preloader"

NotHyper-474:
Reverts FunkinCrew/Funkin#2629 as it's not needed anymore due to Lime having patched [this bug internally](https://github.com/openfl/lime/blob/9f369b7637f6f682ffbfb1257ac37435f6dcb820/CHANGELOG.md?plain=1#L73) since 8.1.0


Hundrec:
This is a `large` change.

NotHyper-474:
More like `smarge`

This bot should probably also remove other tags as well lol and also only consider changes made by the author of the PR.

Issue #3495, opened 2024-09-30
Unknown issue is causing gamplay to skip, or desync with the song, resulting in gameplay stutter, lag, etc


TudioS20:
### Issue Checklist

- [ ] I have properly named the issue
- [ ] I have checked the issues/discussions pages to see if the issue has been previously reported

### What platform are you using?

Itch.io (Downloadable Build) - Windows

### If you are playing on a browser, which one are you using?

None

### Version

0.5.1

### Context (Provide images, videos, etc.)

Upon loading any given song, sometimes the game's audio will stutter, or lag. When this occurs, the general gameplay + visuals will stutter as well.

I have a somewhat low-end computer, so bugs will happen to me that aren't actually bugs, but from what I've heard elsewhere, this bug has happened before to others.

### Steps to reproduce (or crash logs, errors, etc.)

Just...play a song and see what happens. It's kinda random, so I don't know if it happens with EVERYONE.

biomseed:
This happens to me too. It's a bug that happened in the latest patch. I think it happens to everyone

Hundrec:
Can you provide a video of this happening?

TudioS20:
Sadly, no, I can't. However, I CAN tell you that it has been reported on Twitter. Multiple people have talked about lag spikes and things similar to it.

If you check Ninjamuffin's Twitter, he made a post where he talked about 0.5.1 being released. There's some replies that talk about it.

Hundrec:
Thanks for the info!
This tweet demonstrates the bug quite well: https://x.com/MajidyMajidy_/status/1840811401683636307

TudioS20:
What I saw wasn't as...harsh. It was somewhat more...subtle, but yes, that IS in fact, the bug.

NotHyper-474:
Apparently it's the input offset causing this. Another video demonstrating the issue:

https://github.com/user-attachments/assets/79468dd8-5e68-44ea-9555-333de758c4f4

TudioS20:
Yeah, that's the bug. But my offsets aren't changed. I don't even know HOW to change it.

What I had was VERY mild, only appearing like, a couple times. It wasn't...THIS bad.

biomseed:
> Apparently it's the input offset causing this. Another video demonstrating the issue:
> 
>  8mb.video-dWi-yFMMw8Ns.mp4

It is NOT the input offset. I have not changed it once in my entire life

NotHyper-474:
[...]

PR #3546, opened 2024-10-03
[BUGFIX] Changing song/input offsets no longer causes the game to go all weird

JustKolosaki:
[3495](https://github.com/FunkinCrew/Funkin/issues/3495)

Changes a bit of `Conductor`'s code as well as a bit of code in `PlayState` to ensure that the math for offset calculation actually works and doesn't update the game on every beat, thus making the game unplayable with higher offset differences.

https://github.com/user-attachments/assets/1918a35d-0569-430e-8757-299966065048

The video shows command prompt box as well as desired song offsets. Command prompt is shown because AbnormalPoof noticed that the bug is caused by a large amounts of updates to resync vocals, which is then traced.


JustKolosaki:
oops accidentally left in a bit of the code from eric's character offset and animation flipping thingy my bad !!!

charlesisfeline (approved):

PR #3556, opened 2024-10-03
[feature/automatic-offset-flipping] Attribute Is Facing Right

lemz1:
- #3539 Automatic character offset flipping isn't 100% working yet

## FIXES / CHANGES
- Adds `isFacingRight` attribute to character data
- Reverts tankman.json

By using `isFacingRight`, we can fix the problem of wrong animations and broken offsets

Default is `true`

if a character seems to have broken offsets, set `isFacingRight` to false, and see if that fixes it.

### DIFF
--- a/source/funkin/data/character/CharacterRegistry.hx
+++ b/source/funkin/data/character/CharacterRegistry.hx
@@ -443,13 +443,13 @@ class CharacterRegistry
     }
 
     // If `flipXOffsets` is enabled, we need to account for the default value of `flipX`.
-    if (result.flipX && result.flipXOffsets)
-    {
-      for (anim in result.animations)
-      {
-        anim.offsets[0] *= -1;
-      }
-    }
+    // if (result.flipX && result.flipXOffsets)
+    // {
+    //   for (anim in result.animations)
+    //   {
+    //     anim.offsets[0] *= -1;
+    //   }
+    // }
 
     return result;
   }
--- a/source/funkin/data/character/CharacterData.hx
+++ b/source/funkin/data/character/CharacterData.hx
@@ -125,6 +125,13 @@ typedef CharacterData =
   var flipSingAnimations:Bool;
 
   /**
+   * Whether the character's animations are looking to the right or left.
+   */
+  @:optional
+  @:default(true)
+  var isFacingRight:Bool;
+
+  /**
    * The amount to offset the camera by while focusing on this character.
    * Default value focuses on the character directly.
    * @default `[0, 0]`
--- a/source/funkin/play/character/BaseCharacter.hx
+++ b/source/funkin/play/character/BaseCharacter.hx
@@ -46,6 +46,11 @@ class BaseCharacter extends Bopper
   public var flipSingAnimations:Bool = false;
 
   /**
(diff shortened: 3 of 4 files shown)


EliteMasterEric:
Why can't you just use the default value of `flipX`?

lemz1:
lets look at tankman as an example:
_data.flipX is true,
and his offsets are made for when he is the opponent

problem:
when he is the opponent BaseCharacter.flipX is true,
we implemented it that when flipX is true we need to flip offsets and anims, 
however this shouldn't happen here

note:
im pretty sure this is what that animOffsets[0] *= -1
CharacterRegistry was for, however that is an incorrect calculation

EliteMasterEric:
> we implemented it that when flipX is true we need to flip offsets and anims,

But the `funkin.play.stage.Stage#addCharacter()` function data applies either the flipX value, or the inverse of that value!

So if a character has `flipX = true` in their data, then `BaseCharacter.flipX` will become false when they switch sides.

`animOffsets[0] *= -1` is supposed to account for this, but if that calculation is incorrect, then we just need to apply the correct one.

lemz1:
> But the `funkin.play.stage.Stage#addCharacter()` function data applies either the flipX value, or the inverse of that value!
> 
> So if a character has `flipX = true` in their data, then `BaseCharacter.flipX` will become false when they switch sides.

grafik

it only swaps when its the character is on the player side,
but thats a problem for tankman since he will have BaseCharacter.flipX set to `false`, meaning it won't flip the offsets because of we check if BaseCharacter.flipX is `true`

source/funkin/data/character/CharacterRegistry.hx
```diff
@@ -443,13 +443,13 @@ class CharacterRegistry
...
-      for (anim in result.animations)
-      {
-        anim.offsets[0] *= -1;
-      }
-    }
+    // if (result.flipX && result.flipXOffsets)
+    // {
+    //   for (anim in result.animations)
+    //   {
+    //     anim.offsets[0] *= -1;
```
lemz1, line 450: the calculation we actually need is one similar to this:
https://github.com/FunkinCrew/Funkin/blob/fb4bbba3c1160be8fc5ba8b6f6b5aebad69bed72/source/funkin/play/stage/Bopper.hx#L373

lemz1: but im pretty sure even if we have the correct calculation, there will still be edge cases.
so i might just write a python script to automatically set `isFacingRight` to the correct value (just need to hope there is some library that can help me do that)

EliteMasterEric: > the calculation we actually need is one similar to this:
> 
> https://github.com/FunkinCrew/Funkin/blob/fb4bbba3c1160be8fc5ba8b6f6b5aebad69bed72/source/funkin/play/stage/Bopper.hx#L373

I see, I should test this change out later.

lemz1:
> That's what the anim.offsets flip in the `CharacterRegistry` is for!

ooh, i thought that was for something else, my bad
let me do something then real quick

PR #3567, opened 2024-10-04
Center Box Text in Preloader

Burgerballs:
Just a simple quick and dirty fix to center the "DSP" and "FNF" text in the preloader cuz it bugged me a little...

image


ninjamuffin99:
common goated burgerballs commit

Issue #2555, opened 2024-05-19
Bug Report: [Notes stacked on top of each other for no reason]


TylerCKB:
[weed]: <> (FILL THIS ISSUE THING OUT AS MUCH AS POSSIBLE)
[weed]: <> (OR ELSE YOUR ISSUE WILL BE LESS LIKELY TO BE SOLVED!)
[weed]: <> (DO NOT POST ABOUT ISSUES FROM OTHER FNF MOD ENGINES! I CANNOT AND PROBABLY WON'T SOLVE THOSE!)
[weed]: <> (GO TO THEIR RESPECTIVE GITHUB ISSUES AND REPORT THEM THERE LOL!)

[weed]: <> (ALSO MAKE SURE THAT YOU USE PROPER LABELS, IF YOU'RE RUNNING INTO COMPILER ISSUES, USE THE compiler issue LABEL!!!)

### If you are playing the game in a browser, what site are you playing it from?

[weed]: <> (Put an X in the [ ] thingies to fill out checkbox!)
[weed]: <> (something like [x] pretty much, don't screw up or you will look stupid)

- [ ] [Newgrounds](https://www.newgrounds.com/portal/view/770371)
- [X] [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below
- - [X] Windows
- - [ ] Mac
- - [ ] Linux

### If you are playing the game in a browser, what browser are you using?

[weed]: <> (Again, put an x in the [ ] box!)

- [ ] Google Chrome (or chomium based like Brave, vivaldi, MS Edge)
- [ ] Firefox
- [ ] Safari

## What version of the game are you using? Look in the bottom left corner of the main menu. (ex: 0.2.7, 0.2.1, shit like that)
0.3.1

## Please describe your issue. Provide extensive detail and images if possible.
For some reason, after charting a song, I tried to play it, but it was like my inputs weren't registering because even when I pressed the arrows, they just kept going up. I looked at the chart and realized that the ENTIRE chart was duplicated over itself. No clue how it happened, but now I have to go through it and delete every duplicate. Anyone know why that happened?

## If you're game is FROZEN and you're playing a web version, press F12 to open up browser dev window, and go to console, and copy-paste whatever red error you're getting

https://github.com/FunkinCrew/Funkin/assets/109717234/06980291-5698-4052-99bf-cf5e3b292823

JackXson-Real:
1) Why are you still on 0.3.1
[...]

PR #3574, opened 2024-10-04
[FEATURE] Visual Indication of Stacked Notes

NotHyper-474:
## What is this?
Inspired by a [comment by Hundrec](https://github.com/FunkinCrew/Funkin/issues/2555#issuecomment-2391193813) and the many issues about notes stacked on top of each other, this PR intends to add a way to see these pesky notes.

## To Do
- [ ] A warning when you attempt to save the chart file with existing stacked notes.
- [ ] _To be added_

Instead of an indicator maybe these notes should just not be able to be created (e.g when pasting or dragging notes) but what do you think?

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -6467,6 +6467,11 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
     audioVocalTrackGroup.clear();
   }
 
+  function getStackedNotes()
+  {
+    for (note in renderedNotes) {}
+  }
+
   function isNoteSelected(note:Null<SongNoteData>):Bool
   {
     return note != null && currentNoteSelection.indexOf(note) != -1;
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -6467,11 +6467,6 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
     audioVocalTrackGroup.clear();
   }
 
-  function getStackedNotes()
-  {
-    for (note in renderedNotes) {}
-  }
-
   function isNoteSelected(note:Null<SongNoteData>):Bool
   {
     return note != null && currentNoteSelection.indexOf(note) != -1;
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -848,6 +848,8 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
    */
   var stretchySounds:Bool = false;
 
+  var stackedNotes:Array<SongNoteData> = [];
+
   // Selection
 
   /**
@@ -2010,9 +2012,10 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
 
   /**
    * The IMAGE used for the selection squares. Updated by ChartEditorThemeHandler.
-   * Used two ways:
+   * Used three ways:
    * 1. A sprite is given this bitmap and placed over selected notes.
-   * 2. The image is split and used for a 9-slice sprite for the selection box.
+   * 2. Same as above but for notes that are overlapped by another.
+   * 3. The image is split and used for a 9-slice sprite for the selection box.
    */
   var selectionSquareBitmap:Null<BitmapData> = null;
 
@@ -3656,6 +3659,34 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
         }
       }
 
+      // Retrieve notes stacked on top of others (TODO: Is there a non-O(n^2) way of doing this?)
+      // Another TODO: Maybe this can be merged into another existing loop
+      for (i in 0...displayedNoteData.length)
+      {
+        final noteData = displayedNoteData[i];
+
+        if (noteData == null || stackedNotes.contains(noteData))
+        {
+          continue;
+        }
+
+        for (j in 0...displayedNoteData.length)
+        {
+          final otherNote = displayedNoteData[j];
+          if (i == j || noteData == otherNote || stackedNotes.contains(otherNote)) continue;
+
+          if (noteData.getStrumlineIndex() == otherNote.getStrumlineIndex() && noteData.getDirection() == otherNote.getDirection())
+          {
+            // If the notes are close enough in time, consider them stacked
+            if (Math.abs(otherNote.time - noteData.time) < 5)
+            {
+              trace('Found two stacked notes ${noteData}, ${otherNote}');
+              stackedNotes.append(noteData);
+            }
+          }
+        }
+      }
+
       // Add events that are now visible.
       for (eventData in currentSongChartEventData)
       {
@@ -3785,11 +3816,31 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
           selectionSquare.x = noteSprite.x;
           selectionSquare.y = noteSprite.y;
           selectionSquare.width = GRID_SIZE;
+          selectionSquare.color = FlxColor.WHITE;
 
           var stepLength = noteSprite.noteData.getStepLength();
           selectionSquare.height = (stepLength <= 0) ? GRID_SIZE : ((stepLength + 1) * GRID_SIZE);
         }
+        else
+        {
+          if (noteSprite.noteData != null && stackedNotes.contains(noteSprite.noteData))
+          {
+            // TODO: Maybe use another way to display these notes
+            var selectionSquare:ChartEditorSelectionSquareSprite = renderedSelectionSquares.recycle(buildSelectionSquare);
+
+            // Set the position and size (because we might be recycling one with bad values).
+            selectionSquare.noteData = noteSprite.noteData;
+            selectionSquare.eventData = null;
+            selectionSquare.x = noteSprite.x;
+            selectionSquare.y = noteSprite.y;
+            selectionSquare.width = selectionSquare.height = GRID_SIZE;
+            selectionSquare.color = FlxColor.RED;
+          }
+        }
       }
+
+      // TODO: Maybe stackedNotes can just stay as a local variable?
+      stackedNotes.clear();
 
       for (eventSprite in renderedEvents.members)
       {
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -1646,7 +1646,9 @@ class FreeplayState extends MusicBeatSubState
           FunkinSound.playMusic('freakyMenu',
             {
               overrideExisting: true,
-              restartTrack: false
+              restartTrack: false,
+              // Continue playing this music between states, until a different music track gets played.
+              persist: true
             });
           FlxG.sound.music.fadeIn(4.0, 0.0, 1.0);
           close();
--- a/source/funkin/ui/mainmenu/MainMenuState.hx
+++ b/source/funkin/ui/mainmenu/MainMenuState.hx
@@ -183,7 +183,9 @@ class MainMenuState extends MusicBeatState
     FunkinSound.playMusic('freakyMenu',
       {
         overrideExisting: true,
-        restartTrack: false
+        restartTrack: false,
+        // Continue playing this music between states, until a different music track gets played.
+        persist: true
       });
   }
 
--- a/source/funkin/ui/story/StoryMenuState.hx
+++ b/source/funkin/ui/story/StoryMenuState.hx
@@ -242,7 +242,9 @@ class StoryMenuState extends MusicBeatState
     FunkinSound.playMusic('freakyMenu',
       {
         overrideExisting: true,
-        restartTrack: false
+        restartTrack: false,
+        // Continue playing this music between states, until a different music track gets played.
+        persist: true
       });
   }
 
--- a/source/funkin/ui/title/TitleState.hx
+++ b/source/funkin/ui/title/TitleState.hx
@@ -233,6 +233,7 @@ class TitleState extends MusicBeatState
         startingVolume: 0.0,
         overrideExisting: true,
         restartTrack: false,
+        // Continue playing this music between states, until a different music track gets played.
         persist: true
       });
     // Fade from 0.0 to 1 over 4 seconds


AbnormalPoof:
I can see a warning being added, maybe with a button that allows you to see all instances of them? Something like CTRL+F in a web browser:
Screenshot 2024-10-04 at 11 37 22 AM
Though that's probably a *bit* out of scope...

lemz1:
I'd personally just either have a check that removes all stacked notes, or if you want to a button that does it for you.
I don't see the reason for showing where stacked notes are, since you are going to remove them either way, right?

source/funkin/ui/debug/charting/ChartEditorState.hx
```diff
@@ -3656,6 +3659,34 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
         }
       }
 
+      // Retrieve notes stacked on top of others (TODO: Is there a non-O(n^2) way of doing this?)
```
lemz1, line 3662: we could "cut" the notes into chunks an then do the check.
it would basically be O(n*m), (worst case O(n^2)

we could maybe also parallelize it

but i don't think it really matters, since usually there aren't that many notes.

Hundrec:
I'd say giving the charter the freedom to create stacked notes is important. Having an adjustable threshold for what constitutes as "stacked" (0-100ms, for example) could go a long way to enable that freedom

AbnormalPoof:
> since you are going to remove them either way, right?

Unless you're like making a spam chart, then yeah maybe?

lemz1:
> I'd say giving the charter the freedom to create stacked notes is important. Having an adjustable threshold for what constitutes as "stacked" (0-100ms, for example) could go a long way to enable that freedom

I thought stacked meant exactly 0ms, but i guess it might also be good to have a threshold for more control

Hundrec:
I thought of having that customization feature because I've found that Funkin's chart editor would often place notes in an imprecise manner back in the legacy versions.
I haven't really trusted note placements to be millisecond-precise since then, so a slider would ensure any stacked notes not visible to the naked eye would be shown.

NotHyper-474:
I've been thinking of the behaviour when stacking notes when pasting/dragging and came up with these, using the criteria that hold notes have priority over tap notes, and I want opinions on it.

### When Pasting
- If a tap note is pasted on top of another note
  - Tap note: We just override the placed note.
  - Hold note: In this case I think we allow it to be placed and a visual indicator is necessary, or we just don't allow for the pasted note to be placed at all.
- Hold note on top of another note
  - Tap note: Override placed note.
  - Hold note: If they've the same length we just override the placed note too, else we do like with the tap note on top of hold note.

### When dragging
- Tap note
  - On top of tap note: We do like with pasting and replace the overlapped note, but only if the dragged note is deselected, that way if many notes are being dragged the already placed notes don't just get deleted because a dragged note was on top of it.
   - On top of hold note: Maybe similarly to the other but we instead don't allow the dragged note to be placed instead.
- Hold note
  - On top of tap note: Just like for tap note on tap note
  - On top of hold note: I don't know what could be done here, leave it up to the charter's decision or perhaps keep hold note with greater length?

We might need to take note kinds into account as well.

NotHyper-474:
Finally, the messy commit history is no more. ~~Thank god for rebase~~

Hundrec:
Just wanted to say thanks for working on this much-needed feature for months, Hyper!

NotHyper-474:
Technically it's been finished for a while now, I just need to iron out a bug.

NotHyper-474:
Alright, it should be good to merge now. I've decided to discard the idea of preventing stacked notes when moving a note on top of another because it would be a bit complex to do and unlikely to happen when charting normally, and even if it happens there's the visual indicator.

NotHyper-474:
I was experimenting with being able to visualize stacked/overlapping notes in the preview and found a bunch of them at section 75 in the opponent side of 2hot Easy difficulty lol
image

But should this be added too?

AbnormalPoof:
It'd be pretty useful since the user wouldn't have to scroll through the notes slowly to check for stacked notes methinks

Hundrec:
Oh gosh, what a great example of the utility of this feature!
image

You should report this to funkin.assets when you get a chance!

EliteMasterEric:
This is pretty great! Some notes:

- I think the main critique is that the stack notes might be confusing if you don't know what it is you're looking at. It's not intutive/it might need some tooltips.
- Stack note threshold should be based on note snaps rather than on milliseconds. Say, anything closer than a 128th note is a stacked note, for example.
- Is stack detection for hold notes implemented?
- It'd be cool to have a "remove all stack notes" action/keybind. It should filter to just selected notes if there is a selection to match with other actions.

NotHyper-474:
> * I think the main critique is that the stack notes might be confusing if you don't know what it is you're looking at. It's not intutive/it might need some tooltips. **EDIT:** I got some feedback from other people and they said it's obvious what it's for so I think this point might be moot? Something to think about in the future though.

I see. The idea of reusing the select box for that might've not be the best idea, but honestly I can't think of a better one atm. If anyone has suggestions I would appreciate that.

> * Stack note threshold should be based on note snaps rather than on milliseconds. Say, anything closer than a 128th note is a stacked note, for example.

In that case should I keep it on a separate threshold field or should it follow the chart editor's note snap?

> * Is stack detection for hold notes implemented?

I think so. It basically works for any note as a hold note is just a tap note with a non-zero length. The only time the length is considered is when you paste a note (a longer length note overrides a shorter length note but not the other way around).
The way a stacked hold note is rendered is the same as the other stacked notes though.

Hundrec:
Merge conflicts? True...

NotHyper-474:
> Merge conflicts? True...

NotHyper-474:
Merged conflicts solved? Also true...

Hundrec (changes requested):
Thank you Hyper, very cool.
Just a few things before this is ready to be put in game:
1. The undo function can be a bit buggy right now. Not sure how it exactly works, but it wasn't behaving the way I expected it to.
2. The note snap threshold behaves strangely. Sometimes increasing the threshold will un-highlight notes, which shouldn't be happening.
3. When the note snap threshold works as intended, the HaxeUI snap for the setting should be reduced to 0.5 or 1. Changing it by 2 at a time might not be precise enough for very close together notes.

VioletSnowLeopard:
I love that commit message: `Assets submodule (for easier PR testing, remove in rebase before merging, if that ever happens)`

source/funkin/ui/debug/charting/commands/PasteItemsCommand.hx
```diff
@@ -52,29 +59,44 @@ class PasteItemsCommand implements ChartEditorCommand
 
     state.sortChartData();
 
-    state.success('Paste Successful', 'Successfully pasted clipboard contents.');
+    var title = isRedo ? 'Redone Paste Succesfully' : 'Paste Successful';
```
Hundrec, line 62:
```suggestion
    var title = isRedo ? 'Redone Paste Successfully' : 'Paste Successful';
```
Minor spelling mistake 💥

source/funkin/ui/debug/charting/ChartEditorState.hx
```diff
@@ -779,6 +782,13 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
    */
   var currentLiveInputPlaceNoteData:Array<SongNoteData> = [];
 
+  /**
+   * Defines how "close" two notes must be to be considered stacked, based on steps.
+   * For example, setting this to `0.5` (16/32) will highlight notes half a step apart.
+   * Setting it to `0` only highlights notes that are perfectly aligned.
+   */
```
Hundrec, line 789: This comment appears to be outdated. It should say something like:

```suggestion
  /**
   * Defines how "close" two notes must be to be considered stacked, based on steps.
   * For example, setting this to `32` (1/32) will highlight notes half a step apart.
   * Setting it to `192` will highlight notes that are very close together, but not perfectly aligned.
   */
```

source/funkin/ui/debug/charting/commands/RemoveStackedNotesCommand.hx
```diff
@@ -0,0 +1,79 @@
+package funkin.ui.debug.charting.commands;
+
+import funkin.data.song.SongData.SongNoteData;
+import funkin.data.song.SongDataUtils;
+import funkin.data.song.SongNoteDataUtils;
+
+/**
+ * Deletes the given notes from the current chart in the chart editor if any overlap another.
+ * Use only when ONLY notes are being deleted.
```
Hundrec, line 9:
```suggestion
 * Use when ONLY notes are being deleted.
```

source/funkin/ui/debug/charting/commands/PasteItemsCommand.hx
```diff
@@ -52,29 +59,44 @@ class PasteItemsCommand implements ChartEditorCommand
...
+    var msgType = removedNotes.length > 0 ? 'warning' : 'success';
+    var msg = if (removedNotes.length > 0)
+    {
+      'However overlapped notes were overwritten.';
```
Hundrec, line 66:
```suggestion
      'Any overlapping notes were overwritten.';
```
If possible, it would be useful for this notification to indicate the number of notes that were deleted in the process.
Is there a variable you can use to fetch this?

source/funkin/data/song/SongNoteDataUtils.hx
```diff
@@ -0,0 +1,134 @@
...
+    return result;
+  }
+
+  /**
+   * @param noteA First note.
+   * @param noteB Second note.
+   * @param threshold The note stack threshold, in steps.
+   * @return Returns `true` if both notes are on the same strumline, have the same direction
+   * and their time difference in steps is less than the step-based threshold.
+   * A threshold of 0 will return `True...` if notes are exactly aligned.
```
Hundrec, line 124: True...

Hundrec (changes requested):
Good progress! This is close to the stacked note threshold feature I have in mind.

When I opened a build to test this, my threshold was automatically set to `64` (not sure if this happens on clean saves)
I would prefer that a toggle checkbox is added above the threshold labeled "Use Note Stack Threshold".
This would eliminate the need for a "0" option in the number stepper.

I'd also like to see a "1/" to the left of the field, and the math reversed.
Currently, all the values are numbers based on 256, with larger values indicating larger distances. It would be best for 1/`32` to mean half a step, but right now that value is `8`.

Alternatively, you could replace the number stepper with a dropdown that includes 11 options listing all the note steps and "Exact".
It would look like this:
* Exact
* 1/192
* 1/96
...
* 1/8
* 1/4

Maybe this is the simpler solution! If a HaxeUI number stepper can't be locked to a specific set of values, then a dropdown might be a better fit (just like time signatures).

Great job Hyper! We're almost there :)

source/funkin/ui/debug/charting/commands/RemoveStackedNotesCommand.hx
NotHyper-474: Oh it is also like that in `RemoveNotesCommand` where I shamelessly copied from. Free chore PR I guess

source/funkin/ui/debug/charting/commands/PasteItemsCommand.hx
NotHyper-474: The `removedNotes` variable referenced above should work for this :)

Hundrec:
Connection terminated. I'm sorry to interrupt you, Elizabeth, if you still even remember that name, But I'm afraid you've been misinformed. You are not here to receive a gift, nor have you been called here by the individual you assume, although, you have indeed been called. You have all been called here, into a labyrinth of sounds and smells, misdirection and misfortune. A labyrinth with no exit, a maze with no prize. You don't even realize that you are trapped. Your lust for blood has driven you in endless circles, chasing the cries of children in some unseen chamber, always seeming so near, yet somehow out of reach, but you will never find them. None of you will. This is where your story ends. And to you, my brave volunteer, who somehow found this job listing not intended for you, although there was a way out planned for you, I have a feeling that's not what you want. I have a feeling that you are right where you want to be. I am remaining as well. I am nearby. This place will not be remembered, and the memory of everything that started this can finally begin to fade away. As the agony of every tragedy should. And to you monsters trapped in the corridors, be still and give up your spirits. They don't belong to you. For most of you, I believe there is peace and perhaps more waiting for you after the smoke clears. Although, for one of you, the darkest pit of Hell has opened to swallow you whole, so don't keep the devil waiting, old friend. My daughter, if you can hear me, I knew you would return as well. It's in your nature to protect the innocent. I'm sorry that on that day, the day you were shut out and left to die, no one was there to lift you up into their arms the way you lifted others into yours, and then, what became of you. I should have known you wouldn't be content to disappear, not my daughter. I couldn't save you then, so let me save you now. It's time to rest - for you, and for those you have carried in your arms. This ends for all of us. End communication.

source/funkin/ui/debug/charting/ChartEditorState.hx
```diff
@@ -779,6 +782,13 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
    */
   var currentLiveInputPlaceNoteData:Array<SongNoteData> = [];
 
+  /**
+   * Defines how "close" two notes must be to be considered stacked, based on steps.
+   * For example, setting this to `0.5` (16/32) will highlight notes half a step apart.
+   * Setting it to `0` only highlights notes that are nearly perfectly aligned.
```
Hundrec, line 788:
```suggestion
   * Defines how "close" two notes must be to be considered stacked, based on note snaps.
   * For example, setting this to `1/32` will highlight notes half a step apart.
   * Setting it to `Exact` will only highlight notes that are (nearly) perfectly aligned.
```

Hundrec (approved):
Bravo! This PR is now absolutely Perfect 💯

Hundrec:
Huge!

source/funkin/ui/debug/charting/ChartEditorState.hx
```diff
@@ -2936,6 +2985,17 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
       }
     };
 
+    menubarItemDeleteStacked.onClick = _ -> {
```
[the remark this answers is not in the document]
Hundrec: Ooh good reminder! I'll go get those and try again

NotHyper-474:
RIP to whoever tries to go to the chart editor on `develop` without this PR's assets changes

AbnormalPoof:
> RIP to whoever tries to go to the chart editor on `develop` without this PR's asset changes

i'll get those in as well just need to figure out how to

Hundrec:
Easy! Just gotta update the assets submod commit 😁

⛄️

Issue #3631, opened 2024-10-10
Bug Report: Game crashes when a score is larger than 9,999,999 (seven digits).


Shaunyman:
### Issue Checklist

- [X] I have properly named my issue
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.4.1

### Description (include any images, videos, errors, or crash logs)

Note: This issue has not been tested in v0.5.0 or v0.5.1. However, I believe the issue will still occur as the freeplay score only reflects 7 digits in all current versions of the game (v0.5.1 being the newest). 

This is related to my last issue [https://github.com/FunkinCrew/Funkin/issues/3070](url) where I made a song super long that it broke the chart editor. The temporary issue there was to lower the audio bitrate from 41kHz (41,000 Hz), to something lower until the chart editor would work. I ended up getting the full 192 minute chart to work. The new issue that arises from this is that the score that I accumulated was 13,000,000 and change. I believe the issue is that the game's score at the top right of the freeplay menu is only 7 digits and my actual score is 8 digits, which is too many for the freeplay menu to show. The results screen however, can represent a score that is 10 digits (9,999,999,999 max). I believe that the crash arises when the game tries to show the results of a score being more than 7 digits. I have attached pictures of the error messages as well as a video of the crash occurring. Let me know if there's anything I can do to help!
![Scoreboard ok 3](https://github.com/user-attachments/assets/0d5a6b12-d952-41bd-91ed-e442d803f8c5)
![Scoreboard ok error 2](https://github.com/user-attachments/assets/2a7a33fc-c565-4817-a3b6-a1f3a64fd25c)
![Scoreboard ok error](https://github.com/user-attachments/assets/842287d7-2d22-410c-8fd6-d946432e7cbf)
![Scoreboard ok](https://github.com/user-attachments/assets/3f8af834-ad28-46ed-8a29-f035b1e3dc63)

https://github.com/user-attachments/assets/da7a709d-bf4e-4acf-9c67-e358007c00e1

### Steps to Reproduce

[...]

PR #3634, opened 2024-10-10
[FIX] Cap Score Displays to their biggest possible values

KoloInDaCrib:
[3631](https://github.com/FunkinCrew/Funkin/issues/3631)
No, this does not cap the score itself as well. Just wanted to clarify that.
Caps the score displays in `ResultState` and `FreeplayState` to their maximum possible value. Adding more numbers instead would just make it look ugly and yucky :[

(this is not as massive of a change as it shows lol it makes it look like the automatically made spaces from vscode settings are one of the changes lol)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "assets"]
 	path = assets
-	url = https://github.com/FunkinCrew/Funkin-assets-secret
+	url = https://github.com/FunkinCrew/Funkin.assets
 [submodule "art"]
 	path = art
-	url = https://github.com/FunkinCrew/Funkin-art-secret
+	url = https://github.com/FunkinCrew/Funkin.art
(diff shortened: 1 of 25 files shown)


Hundrec:
mmyes
Thank you for preventing another crash!

JustKolosaki:
> mmyes Thank you for preventing another crash!

anytime

Shaunyman:
The people deserve to have their hundred-million scores displayed! Lol, I'm just messing around. While I would like that to be the case, I'm honestly just glad somebody looked into the issue. If both the freeplay score and the results screen score were capped to 9,999,999, that would fix the crash that happens when trying to select the chart difficulty again, right? Because that's really the biggest issue in my humble opinion.

JustKolosaki:
> The people deserve to have their hundred-million scores displayed! Lol, I'm just messing around. While I would like that to be the case, I'm honestly just glad somebody looked into the issue. If both the freeplay score and the results screen score were capped to 9,999,999, that would fix the crash that happens when trying to select the chart difficulty again, right? Because that's really the biggest issue in my humble opinion.

Yup, that should be why the crash happens - the game tries to reference a null object when the score is over the cap. ResultState is in a similar boat

AbnormalPoof (approved):
peak

EliteMasterEric (approved):

PR #3643, opened 2024-10-11
docs: add contributor's guide

amyspark-ng:
is this good


Lasercar:
oof, and that's why you use separate branches for each set of changes you make, rather than trying to use one branch for everything.

PR #3646, opened 2024-10-11
Support Shader on SustainTrail

afeefflox:
idk whoever people want add shader on sustain note (like RGB shader or HSV shader) this is kinda useless I think

no?

well without it the shader won't support it

with shader 
image

without shader 
image


charlesisfeline:
somebody already did a pr of this (#2537) tho

Issue #3641, opened 2024-10-10
Enhancement: add deselect button to stage editor


Burgerballs:
### Issue Checklist

- [X] I have properly named my enhancement
- [X] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested

### What is your suggestion, and why should it be implemented?

In the experimental stage editor, could you add some sort of deselect button, the dotted select filter gets kinda annoying cuz its always there

Lasercar:
This issue has a pending PR.

PR #3650, opened 2024-10-11
[ENHANCEMENT] Prototype Stage Editor De-Select Keybind

KoloInDaCrib:
[3641](https://github.com/FunkinCrew/Funkin/issues/3641)
Adds the keybind for `BACKSPACE` that de-selects the current sprite. Also I accidentally make images get cleared after the stage file is saved, oops! So yeah fixed that as well

ty burgerballs

shoulda pulled debug branch first the change isnt that large


ninjamuffin99:
hmmm personally I feel like ESCAPE is a bit more intuitive "deselect" key, and backspace should delete something! That's off the top of my head, I think another alternative could be CTRL+D which I believe is the photoshop shortcut for deselect. That is, only if we don't use CTRL+D for a non-deselect function in chart editor!

Burgerballs:
Probably could be better if backspace wasnt uses, backspace seems more like a delete key

JustKolosaki:
> hmmm personally I feel like ESCAPE is a bit more intuitive "deselect" key, and backspace should delete something! That's off the top of my head, I think another alternative could be CTRL+D which I believe is the photoshop shortcut for deselect. That is, only if we don't use CTRL+D for a non-deselect function in chart editor!

Fair,
I haven't used ESCAPE because I had already binded ESCAPE to exiting the editor, and I couldnt use DELETE because I binded it to delete the object. I could theoretically work a bit more on this to figure out a way to use a mouse to deselect, like when you click in an empty area,,,,,,,,

Issue #3679, opened 2024-10-12
Game's save data reset in some circumstances (when opening old versions of the game?)


TylerCKB:
### Issue Checklist

- [ ] I have properly named my issue
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.2

### Description (include any images, videos, errors, or crash logs)

I was playing the new patch. Everything was fine. My settings and scores were saved from previous versions of the game no problem. However, I closed out of the game and accidentally booted up an older version of the game (0.5.0). It for some reason was completely silent, as the volume was set to 0. All of my key binds were reset. All of my scores were reset. The new erect songs said the 'New' text on them. I thought nothing of it at first. Maybe if was just an error with the newer data mixing with old versions. But when I went back to the current patch (0.5.2) everything was reset there too. It wiped my save file entirely. I had to replay weekend 1 to access the Pico songs again, and even their scores were gone too. I did a test, and this happened every time I opened the old version. It just resets everything. I don't think there is any way of getting my scores back, but if anyone has any ideas, I'm all ears. 

### Steps to Reproduce

1. Open an older version of the game (pre 0.5.2)
2. Discover that your saves are now fucked

TylerCKB:
Yes, I know that someone reported a similar issue, but they just closed their game, and it happened. My problem seems to be reproduceable.

NotHyper-474:
This bug was also mentioned in a comment https://github.com/FunkinCrew/Funkin/issues/3524#issuecomment-2408637532

GoombarioFan:
I had this happen to me too while testing something, but I didn't think too much of it since I usually don't downgrade and just assumed this was a normal consequence of downgrading versions. I was thinking of suggesting an enhancement of adding a warning that playing on older versions could corrupt save files by detecting if the save file is from a newer version or not, but this also seems like a good place to mention it.

Hundrec:
New versions should start telling users to back up .sol files...

Fused47:
Yep, this happened to me too

speedlorenzo12:
to prevent all data from being deleted due to an update there would be 2 ways:
the game creates 2 separate saves, if the first one breaks the second one must not be modified and the save remains intact.
prevent the file itself from breaking even if I think it could be complicated

EliteMasterEric:
> New versions should start telling users to back up .sol files...

[...]

PR #3727, opened 2024-10-18
[BUGFIX] Fix the Save File being overriden in certain occassions, such as downgrading to versions between 0.5.2 and 0.3.0

JustKolosaki:
[3679](https://github.com/FunkinCrew/Funkin/issues/3679)

Modifies the Save File so that it uses Enum Abstracts instead of Enum fields. Also adds a proper backup system, in case of something going horrible.

https://github.com/user-attachments/assets/85d2c631-0b63-4087-ac22-622ec8e72c14

## Note
Considering Enum fields are bugged, they are the only ones that get reset when changing from the current version to a version in the title. A fix to this would require a fix to haxe's Serializer classes, which is unlikely to happen (both flixel uses openfl to store things and openfl uses haxe's Serializer classes to store the data).

But also like who the hell would play on an older version after 0.3.0 like I get for previous versions but why this range lol

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,40 +5,45 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.5.2] - 2024-10-11
+### Added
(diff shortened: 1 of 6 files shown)


Hundrec:
Emergency save file hotfix yes please!

EliteMasterEric:
These particular changes do not resolve the problem as mentioned in #3679. Replacing enums with enum abstracts without proper workarounds will actually exacerbate the issue, and there is no case in which `loadScoreBackups` will EVER succeed because if the main save slot has bad data written to it, that same invalid data will be written to the backup slot.

Closed in favor of #3728.

Issue #3142, opened 2024-09-13
Bug Report: Song may restart at the end, meanwhile missed notes pass by


NotHyper-474:
### Issue Checklist

- [X] I have properly named the issue
- [X] I have checked the issues/discussions pages to see if the issue has been previously reported

### What platform are you using?

Itch.io (Web)

### If you are playing on a browser, which one are you using?

Firefox

### Version

0.5.0

### Context (Provide images, videos, etc.)

After a song ends, it might repeat once more. While this is happening "ghost" notes pass by.
Edit: In certain cases it seems like it'll keep repeating from where you last paused and basically getting you stuck unless you don't keep pausing and unpausing.

https://github.com/user-attachments/assets/76d26b1a-3972-4be3-a7dd-3e5ba4559eb5

### Steps to reproduce (or crash logs, errors, etc.)

* Just play any song until the end. (I was testing this on Freeplay - Pico character, might need more testing.)
  This might not always happen and seems to occur more frequently on web builds when using Unlocked Framerate, but it is not necessarily caused by it.

AbnormalPoof:
I managed to reproduce this bug back in v0.4.1, but I simply thought it was an issue with one of my mods.

EliteMasterEric:
Definitely looking for help diagnosing/reproducing this.

saturn-volv:
I was encountering the same issue but can only reproduce it on web. I also use firefox.

NotHyper-474:
> I was encountering the same issue but can only reproduce it on web. I also use firefox.

Yeah I was trying it on the standalone build but couldn't reproduce it. Could be a browser specific issue.

NotHyper-474:
> I managed to reproduce this bug back in v0.4.1, but I simply thought it was an issue with one of my mods.

I also had managed to reproduce it in that version while I was testing https://github.com/FunkinCrew/Funkin/pull/2706, but I was on WSL and couldn't replicate it again afterward.

Hundrec:
Encountered this one at the end of Cocoa Erect.

NotHyper-474:
> Encountered this one at the end of Cocoa Erect on itch.io Web.

Were you perchance in Firefox as well?

Hundrec:
I was using Brave when this happened

KATBOI667:
Had this exact same thing happen to me on both DadBattle Pico & Blammed Pico

Opera GX, btw

PlazmaZX:
[...]

Issue #2921, opened 2024-06-27
Voices start to glitch if the other character's voice audio file ends early


sword0115:
## Describe the bug
When playing a custom song, if one character's voice file ends early, the other character's voice will start to glitch. the audio glitch does not happen in the chart editor

## To Reproduce
Create a custom song where one character's voices audio file ends before the other's, for instance, if one character has a voice file that is only 2 minutes long, while the second character's is 2 minutes 30 seconds, the last 30 seconds of the second character's voices will be glitched

## Expected behavior
Following the above example, the last 30 seconds of the second character's voices should play normally.

## Desktop
 - OS: Windows 10
 
 - Version: 0.4.1
 

## Additional context
I know this can be fixed by making the audio files the same length, but FL Studio insists on cutting out any silence at the end of the song by default, I often forget to change that setting and it constantly reverts back to default so I usually have to recompile the voices of my songs

Hundrec:
I've also run into this issue many times! 

This should be pretty easy to fix by someone who knows their way around PlayState.hx

sword0115:
Small update to this issue, it has gotten worse in the 0.5.0 update. Now when one character's vocals end the entire game starts to get choppy, the audio and gameplay seem to slow down and "lag"

Hundrec:
New Swing Mode bug just dropped (reproduced in 0.5.3)

https://github.com/user-attachments/assets/5544e3cd-8c8b-433a-add2-6b6ba0003f4e

https://github.com/user-attachments/assets/ad7cd39b-7d79-4a04-9e3a-a3139b5b68d1

speedlorenzo12:
the same thing happens to me but much less frequently and it also lags in the arrows making me make mistakes

Hundrec:
Does this still happen in 0.7.5?

sword0115:
Sorry for not replying before, I wasn't playing funkin' at the time, but this issue is still present as of version 0.8.1

PR #3775, opened 2024-10-26
[BUGFIX] Fixes the Song repeating infinitely due to the Vocals being longer than the Instrumental

KoloInDaCrib:
[3142](https://github.com/FunkinCrew/Funkin/issues/3142) and at least 7 duplicate issues that I wont even bother to list

The dastardly `resyncVocals` is at it again! I've simply set the `endTime` variable of both Vocals and Inst to the same value, specifically to the lower of the two. That way no song should be repeating infinitely I hope! This was tested on Senpai Erect, which suffered from the issue, and Thorns Erect, which didnt but I wanted to check either way

https://github.com/user-attachments/assets/7e7f46b8-040e-4811-91aa-b5b5fefc1c84

(ignore me being dogshit at the game)
oh and also i would appreciate more testing on this! it will remain as a draft for a bit since I had only tested it on two songs but feel free to take a gander!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,53 +4,63 @@ All notable changes will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.3] - 2024-10-18
+This patch resolves a critical issue which could cause user's save data to become corrupted. It is recommended that users switch to this version immediately and avoid using version 0.5.2.
+### Fixed
+- Fixed a critical issue in which the Stage Editor theme value could not be parsed by older versions of the game, resulting in all save data being destroyed.
+  - Added a check that prevents save data from being loaded if it is corrupted rather than overriding it.
+- `optionsChartEditor.chartEditorLiveInputStyle` in the save data converted from an Enum to a String to fix save data compatibility issues.
+- `optionsStageEditor.theme` in the save data converted from an Enum to a String to fix save data compatibility issues.
+  - In the future, Enum values should not be used in order to prevent incompatibilities caused by introducing new types to the save data that older versions cannot parse.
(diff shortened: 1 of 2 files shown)


NotHyper-474:
I wonder if this affects issue https://github.com/FunkinCrew/Funkin/issues/2921 too

JustKolosaki:
> I wonder if this affects issue #2921 too

I guess? Since I made it so that the shortest audio is used to determine the end of the song, haven't even seen the issue let alone test it at all lol

KutikiPlayz:
it might be better if you did the longest audio instead of shortest, or just cap them all at the instrumental length, cuz I can imagine a case where a composer cuts one vocal file down in length cuz they stop singing before the end of the song, and then the song ends entirely because that one file is shorter

JustKolosaki:
> it might be better if you did the longest audio instead of shortest, or just cap them all at the instrumental length, cuz I can imagine a case where a composer cuts one vocal file down in length cuz they stop singing before the end of the song, and then the song ends entirely because that one file is shorter

Probably, though I am unsure as to how I would go about extending the timer of the audio files without outright making the bug happen again, since longer voice files were the reason the infinite song bug happened in the first place (courtest of NotHyper for finding it out). Flixel already handled the shorter audio files via 'endTime' variable in 'FlxSound'.
I am of the opinion that having to do a janky solution and manually extend vocals in programs like Audacity would be better than having a chance for the bug to occur again.

moxie-coder:
I remember the original solution was to stop the vocals or just turn the volume to 0 when they ended, like seen in the Week 7 source code, but I don’t know if it truely worked well

KutikiPlayz:
yeah if there's anything too short just stop it so it doesn't loop or anything

JackXson-Real:
Even though this issue is already closed, I did some testing to see if this would actually fix it since the bug has unfortunately started happening again on the latest develop build. This PR does not fix the issue, just wanted to mention that for documentation sake.

PR #3777, opened 2024-10-26
[FEATURE] Abstracts In Hscript

lemz1:
## DESCRIPTION
This pr aims to make it possible to import abstract classes or enums.
The macro creates "wrappers" around the actual abstract.

## NOTE
Currently these packages/classes are not supported:

- `thx.Set` (Template issue, specifically not being able to cast Dynamic to some other type)
- `cpp.*` (some issue with cpp.Function.fromStaticFunction)
- `lime.*` (Somehow duplicate fields, could not determinte type for parameter T)
- `openfl.*` (Somehow duplicate fields, could not determinte type for parameter T)

## EXPLANATION
For example `flixel.util.FlxColor`:
We create a class `polymod.abstracts.flixel.util.FlxColor_`.
This class has all static fields of `flixel.util.FlxColor`, and just calls the respective fields from the actual `flixel.util.FlxColor` class. We then create an alias for `flixel.util.FlxColor` and make it point to `polymod.abstracts.flixel.util.FlxColor_`.
We also create a public static function called `create`, which will just call the constructor of the `flixel.util.FlxColor`.

## EXAMPLES
```haxe
import funkin.play.song.Song;
import flixel.util.FlxColor;
import funkin.PathsFunction;
import flixel.util.FlxTypedSignal;

class ExampleSong extends Song
{
	function new()
	{
		super('exampleSong');
		
		trace(FlxColor.fromInt(0xffff0000)); // -65536
		trace(PathsFunction.INST); // INST
		var signal = FlxTypedSignal.create();
		signal.add(() -> { // the lambda can also take in a parameter, which is of type Dynamic
		    trace('DISPATCHED');
		});
		signal.dispatch();
	}
}

```

## TODO
- [x] Abstract functions
- [x] Abstract variables
- [x] Fix some issues when there is no from/to
- [x] Create aliases for all specified abstracts (we specify abstracts using an array containing packages and/or classnames)
- [x] Bit more testing (FlxSignal)
- [x] Try to make it work with pretty much all abstract. (Gonna need to place Dynamics everywhere)
- [ ] (Optional) Get operators to work implicitly if possible

### DIFF
--- a/project.hxp
+++ b/project.hxp
@@ -588,6 +588,15 @@ class Project extends HXProject {
 	function configureCustomMacros() {
 		// This macro allows addition of new functionality to existing Flixel. -->
 		addHaxeMacro("addMetadata('@:build(funkin.util.macro.FlxMacro.buildFlxBasic())', 'flixel.FlxBasic')");
+
+		var abstracts:Array<String> = [
+			'.*',
+			'!thx.Set',
+			'!cpp.*',
+			'!lime.*',
+			'!openfl.*'
+		];
+		addHaxeMacro("funkin.util.macro.PolymodMacro.buildPolymodAbstracts(['" + abstracts.join("', '") + "'])");
 	}
 
   function configureOutputDir() {
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -250,6 +250,11 @@ class PolymodHandler
     Polymod.addImportAlias('lime.utils.Assets', funkin.Assets);
     Polymod.addImportAlias('openfl.utils.Assets', funkin.Assets);
 
+    for (key => value in funkin.util.macro.PolymodMacro.aliases)
+    {
+      Polymod.addImportAlias(key, Type.resolveClass(value));
+    }
+
     // Add blacklisting for prohibited classes and packages.
 
     // `Sys`
@@ -335,8 +340,19 @@ class PolymodHandler
   {
     return {
       assetLibraryPaths: [
-        'default' => 'preload', 'shared' => 'shared', 'songs' => 'songs', 'videos' => 'videos', 'tutorial' => 'tutorial', 'week1' => 'week1',
-        'week2' => 'week2', 'week3' => 'week3', 'week4' => 'week4', 'week5' => 'week5', 'week6' => 'week6', 'week7' => 'week7', 'weekend1' => 'weekend1',
+        'default' => 'preload',
+        'shared' => 'shared',
+        'songs' => 'songs',
+        'videos' => 'videos',
+        'tutorial' => 'tutorial',
+        'week1' => 'week1',
+        'week2' => 'week2',
+        'week3' => 'week3',
+        'week4' => 'week4',
+        'week5' => 'week5',
+        'week6' => 'week6',
+        'week7' => 'week7',
+        'weekend1' => 'weekend1',
       ],
       coreAssetRedirect: CORE_FOLDER,
     }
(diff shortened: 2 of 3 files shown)


AbnormalPoof:
Does this work for abstract classes like `FlxSignal`? Would love to use that class in HScript!

charlesisfeline (approved):

lemz1:
> Does this work for abstract classes like `FlxSignal`? Would love to use that class in HScript!

For now it probably won't work, but i want to add support for it

charlesisfeline:
image

lemz1:
> image

Do you get this error using the latest version of this branch?

charlesisfeline:
> > image
> 
> if you have changed stuff yourself, can you tell what you changed?

not much to polymodmacro in terms of functionality

lemz1:
> > > image
> > 
> > 
> > if you have changed stuff yourself, can you tell what you changed?
> 
> not much to polymodmacro in terms of functionality

The error `Void should be Array<haxe.macro.Field>` gives me an assumption:
Does `PolymodHandler` have a @:build attribute? If yes, remove it. (or better yet, just remove my branch and re-add it, because maybe github made some mistakes when you updated the branch in your fork)

charlesisfeline:
> The error `Void should be Array<haxe.macro.Field>` gives me an assumption: Does `PolymodHandler` have a @:build attribute? If yes, remove it. 
i already removed it

lemz1:
> i already removed it

can you give me your source code?

charlesisfeline:
oh btw i got this now
 ``[ERROR]  (unknown position)``

     | Type name polymod.abstracts.funkin.data.song.SongTimeFormat is redefined from module polymod.abstracts.funkin.data.song.SongData (ed6580333f122e0f68c312851416b0fe)

lemz1:
> I already have a work in progress for this for Polymod.

Im kind of interested in the implementation, is it actual support for abstracts or is it something similar to mine, meaning something like the wrappers im using.

Having actual support for abstracts would be fire.

EliteMasterEric:
> Im kind of interested in the implementation, is it actual support for abstracts or is it something similar to mine, meaning something like the wrappers im using.

If you look at the compiled game code, the wrapper actually exists internally. For example, `flixel.util._FlxColor.FlxColor_Impl_`. I have code that is a work in progress that uses a macro to retrieve this internal class and access its functions. Sadly, this internal class does not provide access to static variables on C++ builds specifically due to a bug.

lemz1:
> > Im kind of interested in the implementation, is it actual support for abstracts or is it something similar to mine, meaning something like the wrappers im using.
> 
> 
> 
> If you look at the compiled game code, the wrapper actually exists internally. For example, `flixel.util._FlxColor.FlxColor_Impl_`. I have code that is a work in progress that uses a macro to retrieve this internal class and access its functions. Sadly, this internal class does not provide access to static variables on C++ builds specifically due to a bug.


Oh, so the static variables not existing were a bug. I was going crazy when I tried using the "Impl" classes. 
But if I think about it, there still might be a way to get the statics.
Maybe using the class expression or something like that?

lemz1:
Do you have a branch of polymod implementing the abstracts. I believe it makes more sense if I try out things, using your implementation.
So I could create a pull request. Or are you already finished?

EliteMasterEric:
> > Do you have a branch of polymod implementing the abstracts. I believe it makes more sense if I try out things, using your implementation. So I could create a pull request. Or are you already finished?
> 
> yeah its right in the [experiment/static-fields](https://github.com/larsiusprime/polymod/tree/experiment/static-fields) branch, ig

Correct. This branch also has some other experimental changes such as allowing static fields and methods which can be accessed from other modules.

lemz1:
I've opened a pull request in the polymod repository, which adds support for reading static member variables.

https://github.com/larsiusprime/polymod/pull/189

PR #3799, opened 2024-10-28
[ENHANCEMENT] ResultState Accept Keybinds

JVNpixels:
# Fix:
This PR allows users to press the accept keybinds when coming from the ResultsState, before you could only press the pause keybinds which I thought was strange, but I kept that in there just in case if there is a reason for it, and also allowing the player to press the accept keybinds as well.
# Media:
Uhh, I made this just so that KreekCraft and his 8 year old audience don't freak out again (lol), along with saving other streamers. Sure, it doesn't get rid of the cartoon sex, but since people assume that space (which is usually an accept keybind set by default) is the way to close the menus, it will for sure help them right before the you know happens.

https://www.youtube.com/live/CeQQamE9kJs?si=y77S6iEvmtZN_mMA

### DIFF
--- a/source/funkin/play/ResultState.hx
+++ b/source/funkin/play/ResultState.hx
@@ -22,6 +22,7 @@ import funkin.ui.freeplay.charselect.PlayableCharacter;
 import flixel.util.FlxColor;
 import flixel.tweens.FlxEase;
 import funkin.graphics.FunkinCamera;
+import funkin.input.Controls;
 import funkin.ui.freeplay.FreeplayState;
 import flixel.tweens.FlxTween;
 import flixel.addons.display.FlxBackdrop;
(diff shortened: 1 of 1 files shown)


Lasercar:
> # Media:
> 
> Uhh, I made this just so that KreekCraft and his 8 year old audience don't freak out again (lol), along with saving other streamers. Sure, it doesn't get rid of the cartoon sex, but since people assume that space (which is usually an accept keybind set by default) is the way to close the menus, it will for sure help them right before the you know happens.
> 
> https://www.youtube.com/live/CeQQamE9kJs?si=y77S6iEvmtZN_mMA

Is there uh, already a suggestion to make a naughtiness enabled version of bf's 100% score? I'd check but I'm currently looking through all the PRs and finding linked issues that haven't been marked as having pending PRs.

Hundrec:
Actually, this PR should be based on `develop`

JVNpixels:
It is now based on develop.

Hundrec (approved):
Simple and clean!

PR #3809, opened 2024-10-30
[ENHANCEMENT] More `ReflectUtil` additions.

AbnormalPoof:
This PR is an extension of #3622 which adds more 3 more functions to `ReflectUtil`: `copyAnonymousFieldOf()`, `deleteAnonymousField()`, and `compareValues()`. These should be relatively self explanatory based on their function names!


Matriculaso:
image

PR #3708, opened 2024-10-15
[BUGFIX] Fix Input Offset interactions with Input Window (+ sustains)

xenkap:
Very odd that this was already done on the opponent's side, but not the Player's.

## Does this PR close any issues? If so, link them below.
https://github.com/FunkinCrew/Funkin/issues/3692, https://github.com/FunkinCrew/Funkin/issues/2382

## Briefly describe the issue(s) fixed.
Input Windows would not be offset by Input Offset, only Note Judgement would be. This pull request fixes that.

## Include any relevant screenshots or videos.
BEFORE: 

https://github.com/user-attachments/assets/0b646dc7-d023-4c25-b1c7-cbd233acb484

AFTER:

https://github.com/user-attachments/assets/8536df28-cfc1-4b3a-ba33-a4e816dcae5e

(Recorded on 200ms Input Offset)
~~(side note: can someone make events responsive to your offset ?? the hey animations are playing too early)~~ _ignore, use visual offset on negative side to achieve effect_

PR #3832, opened 2024-11-05
[BUGFIX] Sustains now give consistent scores [+ songScore as Float]

### LINKED ISSUE #4512: Bug Report:
### Issue Checklist

### Platform

Newgrounds (Web/HTML5)

### Browser

Firefox

### Version

0.6.2

I have noticed a rather important "bug" with the scoring system.

Sustain notes, of course, add onto your score by using the (i believe is the) `elapsed` value, however what I noticed is that this can be inaccurate between different frame rates.

For example, playing Bopeebo on the HARD diff:
On 60 FPS, I get a gold PFC score of around ~33.400, where as on 165 FPS, I get a gold PFC score of ~33.040.

This can be detrimental to your runs, as there is now a leaderboards system implemented with the 0.6.0 update.
Can someone give this a try and possibly report back on it, please?

### Steps to Reproduce

Play any song, on 60 FPS, aim for a consistent gold PFC.
After taking note of the score, play the same song on an even higher FPS.

Compare them both.

xenkap:
N/A

Sustain notes previously did not consistently give score-- it was all dependent on how long you held it for, regardless of when you started holding, disconnected from the actual length of the note. This caused sustains hit earlier to reward more points than sustains hit later, as well as other undesirable effects. This behavior is now fixed.

These demonstrations are done with a script that disables regular note scoring, leaving only sustain scores to count.

Before: 

https://github.com/user-attachments/assets/6882c412-0b71-43a2-9f9e-3b5911b04122

After (VIDEO SLIGHTLY OUTDATED, 100% FIXED NOW): 

https://github.com/user-attachments/assets/96338947-9574-4bd8-b216-824e5288b921

This PR places scoring behavior in Strumline, which may be undesirable.
An alternative approach would be to add a variable in SustainTrail.hx that tracks PlayState's last read on it, and then performing logic where it used to be done anyway. I tried this, but it didn't exactly work out...

This PR also changes songScore to be a Float, and all score-related scripting variables to Float as well.
All saving and displaying of the score still uses the integer using songScoreInt.
Written with the assumption that https://github.com/FunkinCrew/Funkin/pull/3708 will be pulled. If it isn't, removals of offset corrections will be required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,16 +181,6 @@ The Pit Stop 2 update!
 - Removed an unused class from Polymod blacklist. ([06c12e3](https://github.com/FunkinCrew/Funkin/commit/06c12e36c6bd6df4e2be32a3bec540172e79e162)) - by @AbnormalPoof in [#3729](https://github.com/FunkinCrew/Funkin/pull/3729)
 - Many additional small bug fixes.
 
-### Fixed
-
-* @PatoFlamejanteTV made their first contribution in [#3843](https://github.com/FunkinCrew/Funkin/pull/3843)
-* @sphis-Sinco made their first contribution in [#3881](https://github.com/FunkinCrew/Funkin/pull/3881)
-* @MrMadera made their first contribution in [#3934](https://github.com/FunkinCrew/Funkin/pull/3934)
-* @MidyGamy made their first contribution in [#4068](https://github.com/FunkinCrew/Funkin/pull/4068)
-* @Lasercar made their first contribution in [#4082](https://github.com/FunkinCrew/Funkin/pull/4082)
-* @MrScottyPieey made their first contribution in [#4085](https://github.com/FunkinCrew/Funkin/pull/4085)
-
-
 ## New Contributors for 0.6.0
 
 * @PatoFlamejanteTV made their first contribution in [#3843](https://github.com/FunkinCrew/Funkin/pull/3843)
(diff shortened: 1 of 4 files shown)


xenkap:
> An alternative approach would be to add a variable in SustainTrail.hx that tracks PlayState's last read on it, and then performing logic where it used to be done anyway. I'll might make a separate branch for that later-- if I don't, well.
On second thought, this seems like the simpler approach. Working on it now

Hundrec:
If done right, this could finally standardize scores across all attempts!
Please polish this up the best you can :)

xenkap:
> If done right, this could finally standardize scores across all attempts! Please polish this up the best you can :)

Thank you!! I'm not sure what much else I can do on the topic of sustains, since it seems to be 100% accurate now.
But talking about standardized scores, I'm not very sure about the 5ms input window for a perfect score, haha
Although that might be reasonable in other rhythm games. Haven't ever seen one with that strict of a window, though

xenkap:
I just realized the after video shows me taking a score of 88.
I've already fixed this, but since I recorded it before that, it doesn't appear so. Bwekh

xenkap:
Rebased, fixed merge conflicts, squashed it a bit to make it cleaner :fire:
Hoping for a review once more! I think the PR has what it needs, but if there's improvement or changes to be made, then :eyes:

Hundrec:
Thanks! Would still love to see this become a consistent standard for scoring

xenkap:
Nevermind, looks like it should still work fine
I'm not the biggest fan of having scoring be handled by Strumline still, ~but I can't see much of a way around it :sob:~
Scratch that, I do have an idea -- while it won't really fix it, changing out the PlayState.instance.sustainHit calls with a callback instead would allow it to be unspecified or not be strictly tied to PlayState. I think that'd be better !!

on it :boom:

Hundrec:
Aw darn, more conflicts

realvirtu:
Wake up from the dead already and bring back this PR.

FuroYT:
We need you for SDL3 vro
sentpajim

Hundrec:
Kade says this PR doesn't score sustains consistently. 
However, we made a change similar to this internally because the score should be stored as a Float.

PR #3834, opened 2024-11-05
[GITHUB] Create a dedicated charting issue template (and bump version to 0.5.3)

Hundrec:
I've seen a lot of charting issue reports over the past several months. 
One problem I've noticed is that reporters sometimes describe a charting error without providing directions to find it.

That's why I've decided to create a new issue template that:
* automatically applies the `type: charting issue` label
* includes a required "Location" field
* adapts the existing "Description" field to better suit charting issues

I've also bumped the bug report and crash report templates to 0.5.3

image


Hundrec:
Yay! Thanks, Eric!

Issue #3845, opened 2024-11-10
Bug Report: Results Text is moveable


JVNpixels:
### Issue Checklist

- [X] I have properly named my issue
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

v0.5.3

### Description (include any images, videos, errors, or crash logs)

I was spamming arrow keys in the results screen for fun, but then I saw that the text was moving?
(Hitsounds were included for proof of pressing arrow keys)

https://github.com/user-attachments/assets/55df9392-e25f-455c-a742-df63da41823d

Nothing really else to say, but I don't know if this was an intentional debug feature that was left in by mistake or something?

### Steps to Reproduce

1. Open the game...
2. Beat a week/song through Story Mode or Freeplay
3. Get to the results screen
4. Press arrow keys and (maybe) see error.

JVNpixels:
I am still able to reproduce without mods.

AbnormalPoof:
Looks like this is some leftover debug code!
https://github.com/FunkinCrew/Funkin/blob/9e13ea0b2c0799d9a7fae7db56eddb05898739d0/source/funkin/play/ResultState.hx#L724C1-L729C6

amyspark-ng:
i've known about this for months but didn't want people to find out so they don't remove it :pensive:

AbnormalPoof:
> i've known about this for months but didn't want people to find out so they don't remove it 😔

```haxe
import funkin.play.ResultState;
import funkin.modding.module.Module;
import flixel.FlxG;

class ResultsThing extends Module {
	public function new() {
		super('ResultsThing');
	}

	function onUpdate() {
	    if (FlxG.state.subState is ResultState) {
	      if (FlxG.keys.justPressed.RIGHT) FlxG.state.subState.speedOfTween.x += 0.1;
	      if (FlxG.keys.justPressed.LEFT) FlxG.state.subState.speedOfTween.x -= 0.1;
	    }
	}
}

JVNpixels:
no problem ninjamuffin, i love you bro. (kisses cutely)

PR #3848, opened 2024-11-10
[BUGFIX]: Results Screen Debug Feature Removal

JVNpixels:
#3845 
This PR removes the debug feature of moving the text left and right (most likely a test to see how the text functions) but it was still left in release builds, so this PR removes that feature.
IMG_7349

### DIFF
--- a/source/funkin/play/ResultState.hx
+++ b/source/funkin/play/ResultState.hx
@@ -721,12 +721,14 @@ class ResultState extends MusicBeatSubState
       }
     }
 
-    /*if (FlxG.keys.justPressed.RIGHT) speedOfTween.x += 0.1;
+    #if debug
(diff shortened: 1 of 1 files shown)


amyspark-ng:
why not leave it in debug builds instead of commenting?

JVNpixels:
> why not leave it in debug builds instead of commenting?

This is a better idea actually, I was gonna do that but wasn’t sure whether to leave it as a debug only feature or make it a comment, but now that I think of it, it would be better as a debug only feature

Lasercar:
This has been resolved internally?
https://github.com/FunkinCrew/Funkin/issues/3845

Hundrec:
I'll close this one then.

PR #3853, opened 2024-11-11
[ENHANCEMENT] Add `DEBUG_BUILD` to `Constants`.

AbnormalPoof:
Currently, there's no real way to detect whether or not the game is a debug build in HScript (`MacroUtil` doesn't work, so does basically anything that involves getting compiler defines), you COULD do something like this:
```haxe
function isDebugBuild():Bool {
  return Constants.versionSuffix == ' PROTOTYPE';
}
```
However, this is unreliable since V-Slice forks can change the version suffix.

This PR adds a new boolean to `Constants`, `DEBUG_BUILD`. Which is basically self-explanatory, `false` if it's a release build, and `true` if it's debug (has the `FEATURE_DEBUG_FUNCTIONS` compiler flag).

### DIFF
--- a/source/funkin/util/Constants.hx
+++ b/source/funkin/util/Constants.hx
@@ -56,6 +56,11 @@ class Constants
   #end
 
   /**
+   * Whether or not the game is a debug build.
+   */
+  public static final DEBUG_BUILD:Bool = #if FEATURE_DEBUG_FUNCTIONS true #else false #end;
+
+  /**
    * URL DATA
    */
   // ==============================


source/funkin/util/Constants.hx
```diff
@@ -55,6 +55,11 @@ class Constants
   }
   #end
 
+  /**
+   * Wether or not the game is a debug build.
```
anysad, line 59: `Wether` -> `Whether` :p

AbnormalPoof:
Merge conflicts should be fixed up!

PR #3897, opened 2024-11-26
[ENHANCEMENT] Put every Console Trace in the Crash Logs

KoloInDaCrib:
## Description
This Pull Request adds everything printed on the console via `trace` to crash logs in order to better determine the culprit of the crash.

## Potential Drawbacks
1. In one of the printed lines from the Discord RPC, the discord tag of the current user is mentioned, which could be considered an "attack on privacy" or whatevs - though considering the driver info is in the same log file I guess it doesn't hurt to have it be there
2. The Memory Size of the Log has been increased tremendously - from 4KB to at least 30KB, and would continue to increase the longer the user plays the game, which could fill up the user's used space on the drive. A potential fix to that would be deleting logs that are older than a certain time period

https://github.com/user-attachments/assets/0f8d9715-3ea6-4cd7-9378-fca764565f92

### DIFF
--- a/source/funkin/util/logging/AnsiTrace.hx
+++ b/source/funkin/util/logging/AnsiTrace.hx
@@ -2,6 +2,8 @@ package funkin.util.logging;
 
 class AnsiTrace
 {
+  public static var allTraces:Array<String> = [];
+
   // mostly a copy of haxe.Log.trace()
   // but adds nice cute ANSI things
   public static function trace(v:Dynamic, ?info:haxe.PosInfos)
@@ -10,6 +12,7 @@ class AnsiTrace
     return;
     #end
     var str = formatOutput(v, info);
+    allTraces.push(str);
     #if js
     if (js.Syntax.typeof(untyped console) != "undefined" && (untyped console).log != null) (untyped console).log(str);
     #elseif lua
--- a/source/funkin/util/logging/CrashHandler.hx
+++ b/source/funkin/util/logging/CrashHandler.hx
@@ -195,6 +195,19 @@ class CrashHandler
 
     fullContents += '\n';
 
+    fullContents += '=====================\n';
+
+    fullContents += '\n';
+
+    fullContents += 'Console Traces: \n';
+
+    for (line in AnsiTrace.allTraces)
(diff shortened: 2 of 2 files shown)


sphis-Sinco:
I feel like that would make it a bit more difficult to debug issues if you put EVERY one of the traces in there. But maybe like a select few?

MidyGamy:
I would suggest making more like a separate log folder, kinda like Minecraft, so you don't need to crash the game to see logs

KoloInDaCrib:
> I would suggest making more like a separate log folder, kinda like Minecraft, so you don't need to crash the game to see logs

I don't really see the need to do that, it would fill up the storage quickly - the traces in logs are just for better determining what caused the crash since most of the code does a trace.
Besides you could see the trace logs by launching the game from VSCode

AbnormalPoof:
I feel like putting ***every single*** trace in the log file will make it a lot harder to read through crash reports in Issues.
Maybe something like the last 10-20? That way there's still some traces, but it's closer to the crash.

KoloInDaCrib:
> I feel like putting _**every single**_ trace in the log file will make it a lot harder to read through crash reports in Issues. Maybe something like the last 10-20? That way there's still some traces, but it's closer to the crash.

I should probably have updated the title of this PR when I added a featureflag to blacklist some classes from having their traces appear in the logs (such as DiscordClient (trace every update) or FunkinPreloader (trace for everything loaded))
Though your idea does probably seem better, might add it along the blacklisted classes (without blacklisted classes your idea has a chance of having DiscordClient fill up all the log trace slots available since it runs on a different thread)

Hundrec:
Merge conflicts are in the air

sphis-Sinco:
> Merge conflicts are in the air

Its poison

Hundrec:
Merge... conflicts... are in the air...

KoloInDaCrib:
Ngl this PR's kinda old and dookie so imma execute it

Hundrec:
He executed it

macohi:
> Merge... conflicts... are in the air...

It's...
Poison...

Issue #3901, opened 2024-11-28
Bug Report: Scrolling restarts on page change (freeplay)


amyspark-ng:
### Issue Checklist

- [X] I have properly named my issue
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

If you go to the favorites page and then scroll back to the "ALL" songs page, it starts at the first song (tutorial) or random capsule, i think it should be in the capsule that holds the song you just were

Hundrec:
I like this suggestion!
Switching between categories could check whether the currently selected song exists in the new category, then select it.

amyspark-ng:
currently checking if i can fix it in a pr

ninjamuffin99:
when this issue gets closed, it means its been merged internally, thanks for the suggestion!

Hundrec:
Hey @ninjamuffin99, can we get a `status: fixed internally` label or use the existing `status: resolved` label?

Hundrec:
Oops... 
maybe we will need a new label that doesn't auto-close the issue!

amyspark-ng:
wait if this was solved internally shouldn't it be closed?

I'll also close my pr for it i guess

Hundrec:
It looks like Cam wants to keep internally resolved issues open until they're merged in their secret repo

ninjamuffin99:
mostly since we use Linear to keep track of our internal tasks / PRs and whatnot, so if its closed in this Github issue, I believe it will be marked as done internally when it's not merged yet

PR #3902, opened 2024-11-28
fix: now changing lists sends you to the same song

amyspark-ng:
example: if you are in fav songs and you scroll to the "ALL" section it will now send you to the song you were hovering the favs section, opposed to sending you to the start of the list, which it did before
fixes: https://github.com/FunkinCrew/Funkin/issues/3901

i was also trying to make it so it didn't restart the song preview but couldn't figure out how

also i don't know why it marks submodules as updated! i tried to discard and it wouldn't work sorry

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -523,7 +523,18 @@ class FreeplayState extends MusicBeatSubState
       if (grpCapsules.members.length > 0)
       {
         FunkinSound.playOnce(Paths.sound('scrollMenu'), 0.4);
-        curSelected = 1;
+
+        // stuff here
+
+        if (grpCapsules.members[curSelected].freeplayData != null)
+        {
+          curSelected = grpCapsules.members.indexOf(grpCapsules.members[curSelected]);
+        }
+        else
+        {
+          curSelected = 1;
+        }
+
         changeSelection();
       }
     };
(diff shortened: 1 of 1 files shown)


Hundrec:
Inspecting the existing comments, it looks like the devs intended for the song to land on the first song.
I'd treat your change as an enhancement that wouldn't necessarily be better in all cases.
While staying on the same song would be nice when accidentally switching letter sorts, it may confuse users when landing in the middle of the **AB** list, for example.  
I think it would be worthwhile to give this change some consideration before merging.

Also, the submodules are different due to new commits since the last update.

amyspark-ng:
> While staying on the same song would be nice when accidentally switching letter sorts, it may confuse users when landing in the middle of the **AB** list, for example.  
I see what you mean, i could make changes so it doesn't work on letters sorts???? idk would have to see what devs think

Also i created branch from develop, how are there changes? wanted to see a way to revert but couldn't, do you know how?

Hundrec:
It's just GitHub being silly
I think leaving it this way would be fine since Eric has magical merging powers

source/funkin/ui/freeplay/FreeplayState.hx
```diff
@@ -523,7 +523,17 @@ class FreeplayState extends MusicBeatSubState
...
-        curSelected = 1;
+
+        // This makes it so it sends you to the same song in the new list (if it's there).
+        if (grpCapsules.members[curSelected].freeplayData != null)
+        {
+          curSelected = grpCapsules.members.indexOf(grpCapsules.members[curSelected]);
+        }
+        else
+        {
+          curSelected = 1;
```
KoloInDaCrib, line 534: Personally I would've put `curSelected = 0;` here so that if the list doesn't contain any songs, it takes you to the Random button instead

Great PR nonetheless!

KoloInDaCrib (changes requested):

PR #3907, opened 2024-12-01
[CODE REWORK] Replace ControlsMenu's Header Hard-code with a Soft-code

KoloInDaCrib:
Technically it helps answer the discussion [3906](https://github.com/FunkinCrew/Funkin/discussions/3906) because custom headers aren't otherwise possible (even though custom controls aren't possible either)

Replaces the somewhat big hardcode in `ControlsMenu` with softcode. For comparison, this is how the game handles the name of the headers currently: 
image
(insert that one yanderedev tweet)

## Tiny Issue
The header for Note controls is now named "Note" instead of "Notes" - to fix it I would have to change the names of the values for the enum Control, which would break a lot of things, or to add a hardcode for this scenario which I was trying to avoid this entire time

### DIFF
--- a/source/funkin/ui/options/ControlsMenu.hx
+++ b/source/funkin/ui/options/ControlsMenu.hx
@@ -96,46 +96,12 @@ class ControlsMenu extends funkin.ui.options.OptionsState.Page
     {
       var control = controlList[i];
       var name = control.getName();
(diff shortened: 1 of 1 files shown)


charlesisfeline:
well it got rid of the headers


huh

Issue #3925, opened 2024-12-10
Bug Report: Chart Editor Crashing


KadePleaseHelpMe:
### Issue Checklist

- [X] I have properly named my issue
- [ ] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

Whenever I open my mod song (with Darnell-Payable as the player), this message shows up before a crash
![image](https://github.com/user-attachments/assets/dc49b9ff-5c77-44ff-9f3a-a43457fd061d)
Note that this _doesn't_ happen when I load the song from .fnfc

### Steps to Reproduce

1. Open Chart Editor from the ~ menu
2. Open a mod-inserted chart from this dinky little menu
![image](https://github.com/user-attachments/assets/8ed7c9be-3b63-442a-bf4c-585c7e1f4907)
4. Die

KadePleaseHelpMe:
> Potential duplicates:
> 
> * [[Bug Report: Animation Editor Crashing #3924](https://github.com/FunkinCrew/Funkin/issues/3924)] Bug Report: Animation Editor Crashing (86.67%)

Not even close, buddy. A) that's my bug report, B) that's not even the same editor

Hundrec:
Does your song have a Voices-bf.ogg file?
If not, can you add one and try again?

KadePleaseHelpMe:
> Does your song have a Voices-bf.ogg file? If not, can you add one and try again? Otherwise, can you rename Voices-darnell to Voices-bf?

I would be fine with doing that, but that wouldn't explain how the playable stuff gets added
Like it's only when I click it from Create From Song. Otherwise, if I load it from the fnfc, it's fine

Hundrec:
The Chart Editor currently struggles with handling difficulties, variations, and playable characters. 
The Create From Song section often breaks under some circumstances, so you should opt in for a workaround for now.

KadePleaseHelpMe:
aight will do
I'll always save to fnfc 👍

Issue #3857, opened 2024-11-13
Bug Report: Chart editor crash when loading Instrumental-only songs


sphis-sinco:
### Issue Checklist

- [X] I have properly named my issue
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

When loading a instrumental-only song from the Chart Editor from the "Create from song" section, the game will give a crash that looks just about like this

![image](https://github.com/user-attachments/assets/34af385a-2f4f-4c82-872f-1002c000ddd1)

### Steps to Reproduce

1) Make or download a mod with an instrumental-only song
2) Open chart editor
3) Click the song on the "Create from song" Section
4) Error popup

MAZ12211:
I'm charting an Inst only song and I didn't encounter this probably by clicking on a pre existing song like High
Just do that with a song Blazin' as your current workaround for now

MAZ12211:
I clicked on create from a pre existing song because it's a High remix (from what I remember)

IdealistCat:
oh

Lasercar:
This issue has a pending PR (I'm lost track of how many times I've made this comment now too).

sphis-Sinco:
> This issue has a pending PR (I'm lost track of how many times I've made this comment now too).

lol

Kade-github:
I could not reproduce this issue, and it's pretty old. Please re-open this if you get something familiar!

https://github.com/user-attachments/assets/9efaa807-e1eb-4f62-ae16-924dee70a860

PR #3927, opened 2024-12-10
[BUGFIX] Fix the Crash when loading an Instrumental-Only song in the Chart Editor

KoloInDaCrib:
[3925](https://github.com/FunkinCrew/Funkin/issues/3925) [3857](https://github.com/FunkinCrew/Funkin/issues/3857) (let me know if i missed one)

use the `Assets.exists` before retriveing openfl bytes !!!!! otherwise it throws an error and it sucks ass !!!
yeah that's the fix

https://github.com/user-attachments/assets/d023a93b-bcaf-400b-a7d4-b5646fa08e34

### DIFF
--- a/source/funkin/ui/debug/charting/handlers/ChartEditorAudioHandler.hx
+++ b/source/funkin/ui/debug/charting/handlers/ChartEditorAudioHandler.hx
@@ -51,7 +51,7 @@ class ChartEditorAudioHandler
    */
   public static function loadVocalsFromAsset(state:ChartEditorState, path:String, charId:String, instId:String = '', wipeFirst:Bool = false):Bool
   {
-    var trackData:Null<Bytes> = Assets.getBytes(path);
+    var trackData:Null<Bytes> = Assets.exists(path) ? Assets.getBytes(path) : null;
     if (trackData != null)
     {
       return loadVocalsFromBytes(state, trackData, charId, instId, wipeFirst);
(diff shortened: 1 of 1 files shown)


Hundrec:
Very simple, very nice

AbnormalPoof (approved):
neat

EliteMasterEric:
Going to test by deleting the Tutorial vocals but is Blazin not affected by this issue somehow? Or is the issue when the game tries and fails to load vocals for a song that's supposed to have them?

EliteMasterEric:
I haven't been able to reproduce the original issue, does it require creating a custom song? Assets.getBytes shouldn't even be called for a custom song ever.

EliteMasterEric:
> If I recall correctly I think the OP behind the issue removed the vocals but the vocal path in the json was still kept, making the game load non-existent vocals

1) I wasn't able to reproduce this error by deleting the Tutorial vocals so I'm not sure why OP is experiencing this error.

2) If the user deletes the vocals from the song in the UI, the metadata needs to be updated appropriately. If the user deletes the vocals from the song externally (without updating the metadata), the user needs to be presented with an error, since they may not have intended to delete the vocals. Either way, better error handling needs to be presented (ideally via a dialog).

Kade-github:
> > If I recall correctly I think the OP behind the issue removed the vocals but the vocal path in the json was still kept, making the game load non-existent vocals
> 
> 1. I wasn't able to reproduce this error by deleting the Tutorial vocals so I'm not sure why OP is experiencing this error.
> 2. If the user deletes the vocals from the song in the UI, the metadata needs to be updated appropriately. If the user deletes the vocals from the song externally (without updating the metadata), the user needs to be presented with an error, since they may not have intended to delete the vocals. Either way, better error handling needs to be presented (ideally via a dialog).

I also cannot reproduce this issue, deleting the vocals and then loading the chart doesn't crash. It acts pretty normally. I will close this PR and the issue, as this doesn't seem to be the case anymore.

PR #3955, opened 2024-12-25
Fix Resync Buffer

TechnikTil:
This PR fixes the buffer that resync has while doing its job. It does this by keeping the song going while its resyncing. Probably a way better way to do this, but this will work for now.


TechnikTil:
whoops forgot to do that

trayfellow:
Wait, is this the fix for the superevil song stuttering bug of doom

Kade-github (approved):
Significantly reduces stutters for me on resync, good changes.

PR #3965, opened 2024-12-31
[ENHANCEMENT] [MODDING] Custom position and vertex data in Notes and Sustains.

Kade-github:
N/A
Simple changes to expose functions that can be later used in hscript to modify positional data of notes, and make it so the game won't automatically update those notes. This is with `customPositionData` in Strumline.hx. Which when it is true, it will not update note positions.

In SustainTrail.hx, I have added `customVertexData` which will do the same thing but with it's vertices, indices, and uv data.
I have also added, `setIndices(indices:Array<Int>)`, `setVertices(vertices:Array<Float>)`, and `setUVTData(uvtData:Array<Float>)` so you can set these values in hscript. 

The reason we need functions for this is because hscript can not initialize a DrawData as it is an abstract class.
image

### DIFF
--- a/source/funkin/play/notes/Strumline.hx
+++ b/source/funkin/play/notes/Strumline.hx
@@ -75,6 +75,11 @@ class Strumline extends FlxSpriteGroup
   }
 
   /**
+   * Whether the game should auto position notes.
+   */
+  public var customPositionData:Bool = false;
+
+  /**
    * The notes currently being rendered on the strumline.
    * This group iterates over this every frame to update note positions.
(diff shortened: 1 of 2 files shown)


charlesisfeline (approved):
modcharting in vslice is so cool

Keoiki:
Yes please, trying to move notes currently is a massive pain

Eviethecoder:
WAVY NOTES ARE NOW POSSIBLE OMG HECK YEAH

Kade-github:
Going to merge this into one pr

PR #3968, opened 2025-01-02
removed unnecessary mentions of MainMenuState in Freeplay

7oltan:
it closes issues to my mod

im trying to load freeplay on another menu other than mainMenuState and the code i removed with this pr before was fucking up everything

please include this in 0.6.#


JVNpixels (approved):
I think it makes sense for it to be the parent state instead of MainMenuState, especially for different mods.

CyndaReal (approved):
Frankly, I'm surprised this wasn't how it was to begin with

PR #3978, opened 2025-01-07
[ENHANCEMENT] [MODDING] Put a "showNotesplash" public variable in Strumline.hx

Kade-github:
N/A
Super helpful for modding because right now you can't disable note splashes at all unless the strumline isn't a player.

This can later be used as a setting or a playstate param too!


Kade-github:
Going to merge this into one pr

Issue #3896, opened 2024-11-26
Enhancement:  Translation


TriedWharf:
### Issue Checklist

- [X] I have properly named my enhancement
- [X] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested

### What is your suggestion, and why should it be implemented?

I want to translate Friday Night Funkin' into Portuguese.
I can do the translation myself, and all I want is my name in the credits.
Thanks!

JVNpixels:
A spanish translation is being in the works, but I think Portuguese would be a nice translation as well!

Hundrec:
Community-made translations sound awesome, especially considering how much slang is in this game.
I wonder how they'll manage translations, but you can still translate everything in the meantime!

Trofem:
It would be cool if we could add additional fonts (and, perhaps bit controversial: also using not only sprite sheets of characters, but .ttf files or similar types too), as an existing file in images/font path, translate text, pictures into any language, in vslice mod format.

EliteMasterEric:
Official support for translations is planned and will come by Steam release (or possibly before then).

I already have a system set up in Polymod in which you can specify a `locales` folder in which any assets will be replaced. For example, `assets/locales/pt-br/data/introText.txt` will automatically get loaded if the game requests the file `assets/data/introText.txt` and the `pt-br` locale is selected. This will make it possible (and relatively straightforward) for mods to implement custom locales for both the base game and for other mods.

The main obstacles to getting official translations into the main game involve, among other things, hooking up every area of the game to use text from data files rather than hard-coded text and ensuring that custom fonts either properly support localized text or utilize a fallback font when needed. It's lower on our priority list but it's definitely something we're thinking about on a regular basis.

PR #3985, opened 2025-01-10
[ENHANCEMENT] Prepare previously Hard-Coded Texts for Localization Support

KoloInDaCrib:
# Description
Inspired by [a comment by Eric](https://github.com/FunkinCrew/Funkin/issues/3896#issuecomment-2515602030) about the translations being lower in the priority list due to some obstacles. This PR aims to well overcome those obstacles
It introduces a new class, `FunkinText`, which is an extension of `FlxText`. Basically what it does is every time text (or font!) get updated, it checks whether or not the current font has glyphs for all the characters in the text. If not, the font `Arial` is used (subject to change maybe i dunno lol), as it is supposed to have most of the characters due to often being the default font
This PR also hooks up a lot of previously hard-coded texts to a single .json file, `texts.json`. It's placed exactly where `introText.txt` is, in the `assets/data` folder. Eric mentioned in the comment linked above that Polymod has been upgraded with a locales feature, so now every time a locale is loaded, it should also load the `texts.json` file for that locale, so voilla, translation support hell yeah!

# Contents of the JSON file
The `texts.json` file is split into two "groups" per se - "title" and the rest. 

## The "title" part
The "title" part refers to the sequence that happens at the start of the game - basically making it soft-codeable now. It has two fields, `"randomTextSplitter"` and `"texts"`. 

"randomTextSplitter" is a String Field that dictates what the wacky random texts are split by - defaulting to "--". This means that the texts in the introText.txt file are split by "--" to indicate a line break. 

The field "texts" is an array of anonymous structures, with required fields being "beat" and "command". "beat" represents at what beat should the command happen and is an integer value. "command" is basically what should happen when the "beat" happens. It's possible values are:
- "show-text": Shows the multilined text. For all the source code nerds out there it call the function "createCoolText" in TitleState
- "append-text": Appends the text to the current one (i.e. calls "addMoreText"), it's separate from the previous because addMoreText also makes the Haptic vibrate ??? No clue what that is lol
- "clear-text": Gets rid of the awesome texts rest in peace 
- "show-logo": Shows the NG logo if it ain't null
- "hide-logo": The antithesis of the previous entry
- "finish-intro": Also known as the default value if for some reason we don't succeed! Self explanatory methinks

The optional values to the anonymous structure are "texts", which is an array of strings representing what texts to display, if at all, and "randomText", which is an integer value indicating which part of the randomly chosen line (separated via randomTextSplitter!) to show, if at all. Random Lines have a higher priority than "texts"! 

## The rest 
Alright now for the rest part. This one is a little more tricky. Basically because I could not stand making three gajillion predetermined typedefs for text organisation, so the data from the texts.json file is actually a Dynamic value! That means that any organisation of texts is entirely dependant on the person and it won't toggle any errors! For example, this is an excerp of the .json file with previously hard-coded texts related to freeplay menu

```
"freeplay": {
    "title": "FREEPLAY",
    "ost": "OFFICIAL OST",
    "charSelectHint:": "Press [ ${value} ] to change characters"
  },
```

Lets say we want to get the text from the field "ost". To do that, we would simply need to use `Texts.instance.getText("freeplay/ost")`. The slash is indicative of the subgroups n shit. Though be aware! If the command doesn't  detect either of the fields "freeplay" or ost (nested inside "freeplay") or if "freeplay/ost" is not a string, it returns `null`! This system is great for null checks and keeping my sanity in place! 

A keen eye might have noticed "${value}". That is a macro more or less. So basically some texts require additional fields to operate, like the charSelectHint text with the control. The macro is there so that it could automatically be replaced with a replacer. To get a text that should have a replacer, simply run `Texts.instance.getText([daPath], [daArrayOfValues])`. In this case, if we run the aforementioned command with the values ["balls"] we would receive the text "Press [ balls ] to change characters"!

One may wonder why the replacer values are an array instead of a dynamic value. Simply put, during my research to put in as many hard-coded texts I came across one text that required TWO values, eeyikes! To combat it, I had to use arrays. Though what about the macros, one may wonder. Worry not because I have a solution!

Let's say we have a text `Pico Erect was played by pico-christmas` at the location "test/results". Now we could replace "Pico Erect" and "pico-christmas" with the aforementioned macro ${value}, though they both would be linked to one value. If we instead did `${value1} was played by ${value2}` we now have macros for two values! And if we want to get the text, we would run `Texts.instance.getText("test/results", ["Pico Erect", "pico-christmas"])`. By doing so, we get exactly what we would needed! Hooray!!

# Additional Things that I could do but I'm not sure whether or not I should do that 

- [ ] Hook up Controls to the json
- [ ] Hook up Difficulty Names to the json
- [ ] Hook up Week and Song Names to the json
- [ ] Hook up everything related to the Debug Menu (that being chart and stage editor texts) to the json
- [ ] Add a title command to reroll random text
- [ ] probably more but i dont remember

# Conclusion
tired af lol anyways if any of this was too confusing to you take a gander at the `texts.json` file and the code! Perhaps it would be more understandable
ill add footage later kinda busy atm


EliteMasterEric:
I like `FunkinText`, that's pretty useful, but for the actual localization handling, we plan to use [FireTongue](https://github.com/larsiusprime/firetongue), which has close integration with Polymod. It hasn't updated in a bit since I'm the maintainer and haven't worked on translation in a while, but we definitely want to use that instead of your solution. It has a lot of good features like advanced variable replacement.

KoloInDaCrib:
> I like `FunkinText`, that's pretty useful, but for the actual localization handling, we plan to use [FireTongue](https://github.com/larsiusprime/firetongue), which has close integration with Polymod. It hasn't updated in a bit since I'm the maintainer and haven't worked on translation in a while, but we definitely want to use that instead of your solution. It has a lot of good features like advanced variable replacement.

I see, looks quite interesting and pretty much covers everything from this pr. However I noticed the "needs revision" label and I'm curious as to whether you're referring to the checklist of things that I *could* do (which is mostly just more hooking up texts) or something else entirely.

Also should I keep this to only soft-coded intro sequence since FireTongue obviously cannot cover that and it would be somewhat useful for modders?

EliteMasterEric:
> Also should I keep this to only soft-coded intro sequence since FireTongue obviously cannot cover that and it would be somewhat useful for modders?

Firetongue + Polymod has a way of handling asset files (like `introText.txt`). I'll probably just reject this PR and implement localization myself when we're ready to do the work of localizing other assets (like the main menu text).

EDIT: I still like FunkinText, so I'll probably just cherry pick that when I get around to it.

AbnormalPoof:

MidyGamy:
WAIT!!! HAXEUI ALREADY HAVE A LOCALIZATION TOOL!!!

EliteMasterEric:
> WAIT!!! HAXEUI ALREADY HAVE A LOCALIZATION TOOL!!!

We'd probably use that for the chart editor and use Polymod's system for other stuff.

AbnormalPoof:
If this PR gets rejected, then I'd love to see `FunkinText` in a separate PR. It seems really useful!

Hundrec:
Got some merge conflicts here too

KoloInDaCrib:
Will scrap in favour of #4067 or whatever the funkin crew is planning on doing
as for `FunkinText`, maybe another time...

Issue #3991, opened 2025-01-12
Enhancement: more consistent mod dependency behavior


lunacynne:
### Issue Checklist

- [x] I have properly named my enhancement
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested

### What is your suggestion, and why should it be implemented?

2 suggestions:
- check mod dependency based on the mod `title` property (or better yet a new separate `id` property) in their `_polymod_meta.json` files and not the mod folder names (gamebanana adds hashes to filenames to prevent name collisions on their end, but because the dependency system currently checks based on folder name this can cause issues if the player doesnt know to rename it)
- only unload the problematic mod instead of all mods when a dependency error occurs

EliteMasterEric:
Only unloading the problematic mod is not straightforward without a user interface for enabling or disabling mods. People would get confused why their mod doesn't load.

As for mod dependencies I definitely want to add a dedicated `id` property which defaults to the mod folder name if not present in the metadata.

PR #3993, opened 2025-01-12
Set `skipDependencyErrors` to true

AbnormalPoof:
Partially fixes https://github.com/FunkinCrew/Funkin/issues/3991

Currently, Funkin' will refuse to load any mods at all if one mod has a missing dependency. This can cause a bit of confusion for someone just trying to mod the game.

This PR sets `skipDependencyErrors` to true, which only skips mods that have missing dependencies, instead of disabling every mod.
Screenshot 2025-01-11 at 11 25 08 PM

Side note: There should probably be a popup or something to notify the user if a dependency error has occurred. That's probably something for another PR.

### DIFF
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -156,6 +156,8 @@ class PolymodHandler
         // Parsing rules for various data formats.
         parseRules: buildParseRules(),
 
+        skipDependencyErrors: true,
+
         // Parse hxc files and register the scripted classes in them.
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
The dependency system is definitely half-baked. It works great on the Polymod side, but to work in an intuitive way, it requires that the `DEPENDENCY_UNMET` and `DEPENDENCY_VERSION_MISMATCH` error codes be detected and responded to (via popup or something similar).

I'll approve this change but it should be reverted at some point in the future.

PR #4019, opened 2025-01-22
[ENHANCEMENT] EVEN MORE `ReflectUtil` additions.

AbnormalPoof:
This PR adds the remaining missing functions to `ReflectUtil`. This should hopefully complete the class and give scripts access to every function in `Reflect` that isn't harmful or malicious.


EliteMasterEric:
A change I added onto this internally; `Reflect` and `Type` are no longer blacklisted! I instead aliased them to `ReflectUtil`, and added a couple error messages for when you try to call specific blacklisted functions.

PR #4021, opened 2025-01-22
[ENHANCEMENT] Smoother Health Bar and use OUTLINE_FAST for Score Text Outline [FIXED]

CyndaReal:
This is a redo of PR #3680 because I accidentally pressed force push and broke it, whoops! Same ordeal.

below is the original description:

The healthbar is currently choppy, and doesn't smoothly line up with the lerp and health icon positions. This fixes this by giving the healthbar the same number of divisions as its width in pixels (effectively the maximum that fit in the space of the healthbar).

This PR also switches the FlxTextBorderStyle of scoreText from OUTLINE to OUTLINE_FAST, which is not noticeable visually (especially with the pixel font that scoreText uses) and is slightly more optimized.

PR #3680, opened 2024-10-12
[ENHANCEMENT] Smoother Health Bar and use OUTLINE_FAST for Score Text Outline

CyndaquilDAC:
The healthbar is currently choppy, and doesn't smoothly line up with the lerp and health icon positions. This fixes this by giving the healthbar the same number of divisions as its width in pixels (effectively the maximum that fit in the space of the healthbar).

This PR also switches the FlxTextBorderStyle of scoreText from OUTLINE to OUTLINE_FAST, which is not noticeable visually (especially with the pixel font that scoreText uses) and is slightly more optimized.

PR #4031, opened 2025-01-24
[ENHANCEMENT] Smoother Health Bar and use OUTLINE_FAST for Score Text Outline [FIXED, AGAIN]

Jaduplex:
Redo, AGAIN, of PR #3680 and PR #4021 because I'm an absolute genius and don't know how to use git that well. Same ordeal as the last 2, just actually proper.

below is the original description:

The healthbar is currently choppy, and doesn't smoothly line up with the lerp and health icon positions. This fixes this by giving the healthbar the same number of divisions as its width in pixels (effectively the maximum that fit in the space of the healthbar).

This PR also switches the FlxTextBorderStyle of scoreText from OUTLINE to OUTLINE_FAST, which is not noticeable visually (especially with the pixel font that scoreText uses) and is slightly more optimized.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1584,11 +1584,15 @@ class PlayState extends MusicBeatSubState
     healthBar.scrollFactor.set();
     healthBar.createFilledBar(Constants.COLOR_HEALTH_BAR_RED, Constants.COLOR_HEALTH_BAR_GREEN);
     healthBar.zIndex = 801;
+    // One division for each pixel of the bar's width ensures maximum bar smoothness.
+    // This is better looking, and syncs with the lerped icon movement better.
+    // This can be slightly more heavy on the CPU, though, so if lower end device performance options are added, maybe make them affect this?
+    healthBar.numDivisions = Std.int(healthBarBG.width - 8);
     add(healthBar);
 
     // The score text below the health bar.
     scoreText = new FlxText(healthBarBG.x + healthBarBG.width - 190, healthBarBG.y + 30, 0, '', 20);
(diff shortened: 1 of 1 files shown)


source/funkin/play/PlayState.hx
```diff
@@ -1584,11 +1584,15 @@ class PlayState extends MusicBeatSubState
     healthBar.scrollFactor.set();
     healthBar.createFilledBar(Constants.COLOR_HEALTH_BAR_RED, Constants.COLOR_HEALTH_BAR_GREEN);
     healthBar.zIndex = 801;
+    // One division for each pixel of the bar's width ensures maximum bar smoothness.
+    // This is better looking, and syncs with the lerped icon movement better.
+    // This can be slightly more heavy on the CPU, though, so if lower end device performance options are added, maybe make them affect this?
```
NeeEoo, line 1589: The performance impact is super minimal since the name is confusing, the code where its used is actually
https://github.com/HaxeFlixel/flixel/blob/dev/flixel/ui/FlxBar.hx#L762-L766

```haxe
var fraction:Float = (value - min) / range;
var maxScale:Float = (_fillHorizontal) ? barWidth : barHeight;
var scaleInterval:Float = maxScale / numDivisions;
var interval:Float = Math.round(Std.int(fraction * maxScale / scaleInterval) * scaleInterval);
```

this code is basically just
```haxe
var fraction:Float = (value - min) / range;
var maxScale:Float = (_fillHorizontal) ? barWidth : barHeight;
var interval:Float = Math.round(Std.int(fraction * numDivisions) / numDivisions * maxScale);
```
All it does it make it more stepped the lower the value is, and more smooth the higher the value is.
So the performance impact isn't big

CyndaReal: My bad. Not gonna remove the comments right now though, because I don't want to add unnecessary comments and I don't feel like cloning just to squash.

Starexify:
Rip score text change

PR #4033, opened 2025-01-24
[ENHANCEMENT] Add `readDirectory` to FileUtil

KoloInDaCrib:
back when i was making a menu for my replay mod i had to use a dinky ass method of listing all replay files, which was saving the file paths to the save file and then later checking if they exist, yikes! 
this PR adds a wrapper for FileSystem's `readDirectory` in FileUtil to avoid having modders struggling and doing something similar to the method described above

### DIFF
--- a/source/funkin/util/FileUtil.hx
+++ b/source/funkin/util/FileUtil.hx
@@ -357,6 +357,23 @@ class FileUtil
     #end
   }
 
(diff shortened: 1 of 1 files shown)


Keoiki:
#3032 would already implement this and a lot more.

PR #4057, opened 2025-01-28
no dialogue music crash fix

Lasercar:
A one line fix that prevents a null reference error crash if a dialogue doesn't have any music set when the dialogue ends.


EliteMasterEric:
This already got done internally 👍

PR #4099, opened 2025-02-04
[ENHANCEMENT + BUGFIX] IntervalShake absolute coords

lunacynne:
- `IntervalShake` now uses absolute coordinates instead of being based on width/height
- `axes` replaced with `scale` point
- fixed current usage y axis using `width` instead of `height` for the scalar


Hundrec:
Merge conflicts - for documentation purposes

TechnikTil:
Any updates on fixing the merge conflicts?

Issue #4120, opened 2025-02-08
Bug Report: Note trails position are not updated when moving a note in editor


amyspark-ng:
### Issue Checklist

- [x] I have read the pinned [Issues Guide](https://github.com/FunkinCrew/Funkin/issues/4050)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly named my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

When you move notes in the editor the trail that follows them won't move with them

https://github.com/user-attachments/assets/2e84224f-856f-47ee-aaa3-aa91eafb5131

### Steps to Reproduce

1. Open chart editor
2. Select long notes
3. Move them

Lasercar:
dangit, I close one issue, and 2 more pop up! 

Seriously though, I'll investigate this. Maybe it's a simple fix? 

I'll edit this comment as I investigate.

amyspark-ng:
it do be like that

do update me how you fix it! i was looking through the code but couldn't find how to fix it, chart editor code is still an enigma for me

amyspark-ng:
https://github.com/FunkinCrew/Funkin/blob/develop/source%2Ffunkin%2Fui%2Fdebug%2Fcharting%2FChartEditorState.hx#L3763 This line seems to be the one that changes the position but being honest i don't really understand how to fix it

NotHyper-474:
> https://github.com/FunkinCrew/Funkin/blob/develop/source%2Ffunkin%2Fui%2Fdebug%2Fcharting%2FChartEditorState.hx#L3763 This line seems to be the one that changes the position but being honest i don't really understand how to fix it

Our goat Kolo already made a PR addressing the issue :) https://github.com/FunkinCrew/Funkin/pull/4127

Lasercar:
This issue now has a pending PR.

PR #4127, opened 2025-02-09
[BUGFIX] Move the Note Trails when moving a Held Note in the Chart Editor

KoloInDaCrib:
#4120 

Essentially copied some code from the regular notes and applied it to note trails

https://github.com/user-attachments/assets/3afc1ef0-1e50-4810-9620-9e337868f0e6

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -3649,6 +3649,8 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
           var noteLengthPixels:Float = noteSprite.noteData.getStepLength() * GRID_SIZE;
 
           holdNoteSprite.noteData = noteSprite.noteData;
+          holdNoteSprite.overrideStepTime = null;
+          holdNoteSprite.overrideData = null;
           holdNoteSprite.noteDirection = noteSprite.noteData.getDirection();
 
           holdNoteSprite.setHeightDirectly(noteLengthPixels);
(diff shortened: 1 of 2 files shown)


Lasercar:
Ah, I thought I might had to create a null hold note var and then pass a hold note when one is found. Nice work!

Also, you could probably remove that // TODO: Handle selection of hold notes. comment (unless it's for something else?).

amyspark-ng:
i wonder why the editor handles things this way, im making a similar editor and all notes (regular, long and events) are something called stamps, with this all the behaviour is the same

at least make long notes and regular notes the same class, can't long notes just be a regular note that draws differently if length is not null?

KoloInDaCrib:
> i wonder why the editor handles things this way, im making a similar editor and all notes (regular, long and events) are something called stamps, with this all the behaviour is the same
> 
> at least make long notes and regular notes the same class, can't long notes just be a regular note that draws differently if length is not null?

`SustainTrail` uses the rendering system similar to `FlxStrip`, which is using triangles, vertices and similar to render the notes' graphic instead of rendering it normally like a regular `FlxSprite`. Granted it is possible to mix them into one class but I think that would've been less readable. Like back before 0.3 every util was using `CoolUtil` which was executed (rest in peace) in favour of the utils folder simply due to the fact that it's more readable, despite some utils not even being a 100 lines long excluding comments and therefore could've been easily bundled with other utils

amyspark-ng:
so it's basically because the trail has different rendering method?

Hundrec (approved):
Brings a tear to my eye.
Thank you Kolo.

AbnormalPoof (approved):
peak

Issue #3923, opened 2024-12-10
Bug Report: Keyboard Buttons Still Work In Typing Fields


Average-FNF-Modder:
### Issue Checklist

- [X] I have properly named my issue
- [ ] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Newgrounds (Web/HTML5)

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

Many actions in FNF are tied to keys on the keyboard, such as fullscreening and changing volume. But currently, these keys remain functional when you're in a typing field, which is problematic because it leads to accidentally performing actions you didn't want to. For me, it's most annoying when I'm typing it song names or composers and it goes between fullscreen and not when I don't want it to.

### Steps to Reproduce

1. Open The Game
2. Go To Any Available Typing Field (All In Chart Editor)
3. Use Keys That Already Do Things Outside Of Typing Their Regular Character On The Screen

Hundrec:
I've encountered this inconvenience as well (zero, for example) 
It's worth seeing if you can disable keybinds while in text fields.

amyspark-ng:
pretty sure you could

Lasercar:
> I've encountered this inconvenience as well (zero, for example) It's worth seeing if you can disable keybinds while in text fields.

![Image](https://github.com/user-attachments/assets/55ddd2b6-bd32-4a89-bf87-07573d6e13d1)

Looks like this was accounted for, but it doesn't work for some reason? It's like the focus manager always returns false for some reason.

This issue would require a proper developer to investigate it seems like.

Average-FNF-Modder:
> > I've encountered this inconvenience as well (zero, for example) It's worth seeing if you can disable keybinds while in text fields.
> 
> ![Image](https://github.com/user-attachments/assets/55ddd2b6-bd32-4a89-bf87-07573d6e13d1)
> 
> Looks like this was accounted for, but it doesn't work for some reason? It's like the focus manager always returns false. I tried stuffing this check into the controls and it still didn't prevent any input when inputting into fields and the like.
> 
> It seems like this issue would require a proper developer to investigate it.
> 
> Also, loading up the focusmanager.cpp file caused my builds of the game to fail until I restarted (and deleted the obj folder that generated), which was weird. Did vsc for some reason try to start building funkin in c++ because I opened a single .cpp file?
> 
[...]

PR #4131, opened 2025-02-11
[FIX] Don't allow fullscreen keybind in type fields

lemz1:
## ISSUES
#3923

## DESCRIPTION
This pr adds a check to the event listener in `WindowUtil`. It checks whether haxe ui is focused or not.

Before when you were typing in a field and would press the button that is bound to the fullscreen toggle, you would toggle fullscreen.

Now this doesn't happen anymore.

### DIFF
--- a/source/funkin/util/WindowUtil.hx
+++ b/source/funkin/util/WindowUtil.hx
@@ -117,6 +117,11 @@ class WindowUtil
 
     #if (desktop || html5)
     openfl.Lib.current.stage.addEventListener(openfl.events.KeyboardEvent.KEY_DOWN, (e:openfl.events.KeyboardEvent) -> {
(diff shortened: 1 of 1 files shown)


AbnormalPoof:
Might need to clarify more about what this does, maybe with a before/after.

Hundrec:
Tested and clarified! The PR description is accurate.

Lasercar:
Yeah, this PR description is accurate, it just doesn't fix the the focus manager thing: https://github.com/FunkinCrew/Funkin/issues/3923#issuecomment-2646179742, but it does make it use it so that it'll work when it is fixed.

cyn0x8 (approved):

Hundrec (approved):
Tested again, definitely works for the fullscreen button!

Issue #2227, opened 2024-05-01
Bug Report: Hold notes stay held without input after unpausing


MafiaPlay2:
[weed]: <> (FILL THIS ISSUE THING OUT AS MUCH AS POSSIBLE)
[weed]: <> (OR ELSE YOUR ISSUE WILL BE LESS LIKELY TO BE SOLVED!)
[weed]: <> (DO NOT POST ABOUT ISSUES FROM OTHER FNF MOD ENGINES! I CANNOT AND PROBABLY WON'T SOLVE THOSE!)
[weed]: <> (GO TO THEIR RESPECTIVE GITHUB ISSUES AND REPORT THEM THERE LOL!)

[weed]: <> (ALSO MAKE SURE THAT YOU USE PROPER LABELS, IF YOU'RE RUNNING INTO COMPILER ISSUES, USE THE compiler issue LABEL!!!)

### If you are playing the game in a browser, what site are you playing it from?

[weed]: <> (Put an X in the [ ] thingies to fill out checkbox!)
[weed]: <> (something like [x] pretty much, don't screw up or you will look stupid)

- [ ] [Newgrounds](https://www.newgrounds.com/portal/view/770371)
- [ ] [Itch.io](https://ninja-muffin24.itch.io/funkin)? Specify below
- - [X] Windows
- - [ ] Mac
- - [ ] Linux

### If you are playing the game in a browser, what browser are you using?

[weed]: <> (Again, put an x in the [ ] box!)

- [ ] Google Chrome (or chomium based like Brave, vivaldi, MS Edge)
- [ ] Firefox
- [ ] Safari

## What version of the game are you using? Look in the bottom left corner of the main menu. (ex: 0.2.7, 0.2.1, shit like that)

0.3.0 Prototype

## Have you identified any steps to reproduce the bug? If so, please describe them below in as much detail as possible. Use images if possible.

Pause while holding a hold note and after unpausing it will finish the hold on its own (not really sure how to explain it lol)

## Please describe your issue. Provide extensive detail and images if possible.

https://github.com/FunkinCrew/Funkin/assets/62815557/2b019d88-3088-4168-a295-60015bc39c67

amyspark-ng:
I don't know about this, i think this is cool.....

Hundrec:
Still happens in 0.5.0

amyspark-ng:
i still think it's cool...

Lasercar:
Is this really a bug though? 

[...]

PR #4145, opened 2025-02-15
Hold note still pressed after resuming game fix

Lasercar:
https://github.com/FunkinCrew/Funkin/issues/2227
Pressing a hold note then pausing the game without unpressing, then releasing your hold and resuming, results in the hold note still being pressed despite having removed your input.

This fixes it for a single hold note, but it probably doesn't work properly on multiple hold notes?

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -332,6 +332,11 @@ class PlayState extends MusicBeatSubState
    */
   public var disableKeys:Bool = false;
 
+  /**
+   * The previous difficulty the player was playing on.
+   */
+  public var previousDifficulty:String = Constants.DEFAULT_DIFFICULTY;
+
   public var isSubState(get, never):Bool;
 
   function get_isSubState():Bool
@@ -603,6 +608,7 @@ class PlayState extends MusicBeatSubState
     // Apply parameters.
     currentSong = params.targetSong;
     if (params.targetDifficulty != null) currentDifficulty = params.targetDifficulty;
+    previousDifficulty = currentDifficulty;
     if (params.targetVariation != null) currentVariation = params.targetVariation;
     if (params.targetInstrumental != null) currentInstrumental = params.targetInstrumental;
     isPracticeMode = params.practiceMode ?? false;
@@ -813,7 +819,11 @@ class PlayState extends MusicBeatSubState
 
       prevScrollTargets = [];
 
-      dispatchEvent(new ScriptEvent(SONG_RETRY));
+      var retryEvent = new SongRetryEvent(currentDifficulty);
+
+      previousDifficulty = currentDifficulty;
+
+      dispatchEvent(retryEvent);
 
       resetCamera();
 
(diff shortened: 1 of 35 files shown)


source/funkin/play/PlayState.hx
```diff
@@ -2173,7 +2178,7 @@ class PlayState extends MusicBeatSubState
      */
   function onKeyRelease(event:PreciseInputEvent):Void
   {
-    if (isGamePaused) return;
+    // if (isGamePaused) return;
```
AbnormalPoof, line 2181: FYI The style guide advises against leaving commented out code

Lasercar: Sure, I'll remove this then.

AbnormalPoof (approved):
LGTM

EliteMasterEric:
Definitely test and make sure it works on multiple hold notes before requesting review again.

Lasercar:
> Definitely test and make sure it works on multiple hold notes before requesting review again.

I've just tested it on multiple hold notes and it definitely works!

Hundrec:
I think that's what Eric meant

Can you record and send a video of the current behavior?

Hundrec:
In your videos, are you still holding the note while unpausing? 

The behavior in the videos don't seem any different from 0.5.3

If so, can you record a video of Bopeebo without holding the note while unpausing?

Lasercar:
> In your videos, are you still holding the note while unpausing?
> 
> The behavior in the videos don't seem any different from 0.5.3
> 
> If so, can you record a video of Bopeebo without holding the note while unpausing?

Ah, yeah, I see another problem:

https://github.com/user-attachments/assets/af3507bd-c71f-4442-ada4-1f1c6acda21c

Sometimes, a note hold still sticks?

Hundrec:
Yes, I believe this is what Eric meant.
Please mark this PR as a draft for now since it doesn't seem to fix the notes staying held

Lasercar:
I think this is going to remain a draft for a while - I basically need a way to check if one of the keys is being held down and return of so but it turns out Just_Pressed and Pressed end up true at the same time when the key is pressed, so there's no way I can fix this issue atm:
https://haxeflixel.com/documentation/keyboard/

Lasercar:
> > If there was another filter called Hold or something that only returns true after one frame has passed of the key being down, then I'd be able to fix the remaining issue.
> 
> Can't you just check `if (key.pressed && !key.justPressed)`?

I think I tried that, but since this is happening right after a key is pressed, both of them are always returning true, so it sill gets passed.

I'll have another go at fixing it tomorrow when I get back from work though.

Lasercar:
Ok, I figured out that I can use `FlxG.keys.anyJustPressed` and `FlxG.keys.anyPressed`

Though weirdly, it appears that a key held into the into pause menu doesn't result it in that key being seen as held down until that key is released and pressed again, so while my code _will_ work, due the the way flixel works, it still doesn't _sigh_.

Lasercar:
I did it!

AbnormalPoof (approved):
Works

JackXson-Real (approved):
works

Hundrec (approved):
The current implementation doesn't allow for inputs to carry through after releasing and re-pressing the notes. This seems fair though, so I approve!

Issue #2273, opened 2024-05-04
Enhancement: Have the chart editor save your settings for hitsound & instrumental volume


verecundius:
I feel this would be a nice QOL addition, thats all

Hundrec:
This game still doesn't save all the volume values as of 0.5.0
For some reason, it only saves the metronome and player hitsounds, but resets everything else to 100%

![image](https://github.com/user-attachments/assets/ba46f2fe-4781-4ec2-b637-3f76b952703f)

Keoiki:
jesus this issue is from MAY 4TH and it's still not fixed

AbnormalPoof:
Supposedly it does save the preferences:
```haxe
  public function writePreferences(hasBackup:Bool):Void
  {
    var save:Save = Save.instance;

    // Can't use filter() because of null safety checking!
    var filteredWorkingFilePaths:Array<String> = [];
    for (chartPath in previousWorkingFilePaths)
      if (chartPath != null) filteredWorkingFilePaths.push(chartPath);
    save.chartEditorPreviousFiles = filteredWorkingFilePaths;

    if (hasBackup) trace('Queuing backup prompt for next time!');
    save.chartEditorHasBackup = hasBackup;

    save.chartEditorNoteQuant = noteSnapQuantIndex;
    save.chartEditorLiveInputStyle = currentLiveInputStyle;
    save.chartEditorDownscroll = isViewDownscroll;
    save.chartEditorPlaytestStartTime = playtestStartTime;
    save.chartEditorTheme = currentTheme;
    save.chartEditorMetronomeVolume = metronomeVolume;
    save.chartEditorHitsoundVolumePlayer = hitsoundVolumePlayer;
    save.chartEditorHitsoundVolumeOpponent = hitsoundVolumeOpponent;
    save.chartEditorThemeMusic = this.welcomeMusic.active;

    // save.chartEditorInstVolume = audioInstTrack.volume;
    // save.chartEditorVoicesVolume = audioVocalTrackGroup.volume;
    // save.chartEditorPlaybackSpeed = audioInstTrack.pitch;
  }
```
However, they likely don't become saved because `Save.instance.flush()` isn't being called.

Lasercar:
Every single chart editor option set does flush after setting, so I'm assuming this was fixed at some point between 0.5.0 and 0.5.3?

@Hundrec Can you test this and confirm if this issue can be closed?

Hundrec:
This issue is still present in my experience 

[...]

PR #4149, opened 2025-02-15
Chart Editor - Save all audio settings

Lasercar:
https://github.com/FunkinCrew/Funkin/issues/2273
Only the metronome and player hitsounds were saved. This PR makes it save and load all audio settings. 
https://github.com/user-attachments/assets/3b963cc4-bdb9-459d-b99c-c934f32a62f6

### DIFF
--- a/source/funkin/save/Save.hx
+++ b/source/funkin/save/Save.hx
@@ -143,6 +143,10 @@ class Save
           metronomeVolume: 1.0,
           hitsoundVolumePlayer: 1.0,
           hitsoundVolumeOpponent: 1.0,
+          instVolume: 1.0,
+          playerVoiceVolume: 1.0,
+          opponentVoiceVolume: 1.0,
+          playbackSpeed: 1.0,
           themeMusic: true
         },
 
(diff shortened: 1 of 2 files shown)


Hundrec:
Huge if true!

Lasercar:
> Huge if true!

Well it is true! You can always pull these changes into a local branch to test them, rather than relying on video/screencaps, just copy the clone command from my fork and use it in the place of origin and then specify my branch for this PR, like this:

HTTPS: ```git pull https://github.com/Lasercar/Funkin.git chart-editor-save-all-audio-settings```
SSH: ```git pull git@github.com:Lasercar/Funkin.git chart-editor-save-all-audio-settings```

source/funkin/save/Save.hx
```diff
@@ -143,6 +143,10 @@ class Save
           metronomeVolume: 1.0,
           hitsoundVolumePlayer: 1.0,
           hitsoundVolumeOpponent: 1.0,
+          instVolume: 1.0,
+          playerVoiceVolume: 1.0,
+          opponentVoiceVolume: 1.0,
+          playbackSpeed: 1.0,
```
Hundrec, line 149: Tested this and it works!

My only concern is that the four values you added set the song tracks' volumes to 1% rather than 100%, and the playback speed to the minimum 0.05x rather than 1x. 

I guess all 7 of these values need to be 100 instead?

Lasercar: Ahh, my bad. I decided against making new variables to store the values for those and instead just set them directly on the sliders and forgot to multiply the number by 100 lol.

I'll fix it right now.

Lasercar: Alright, done. Pull my changes and try it again, and if you can also confirm that it works for you, then resolve this.

Hundrec:
Weirdly enough, the default settings on a wiped save file are still 1%
I have no idea how the code works, but I'm sure you've got this!

Lasercar:
> Weirdly enough, the default settings on a wiped save file are still 1% I have no idea how the code works, but I'm sure you've got this!

huh? That's weird. I made sure to test that it all worked...

I was just about to move onto doing something else, _sigh_

Hundrec:
On second thought, I might've messed something up
Don't worry about this for now, I'll test it again tomorrow!

JackXson-Real (approved):
Works flawlessly! They save both when exiting the chart editor and when restarting the game. I would recommend one thing though: On new save files instead of the default playback speed being 2x, it should be changed to 1x. Other than that I think it's okay to keep every other setting's default at 100%.

Hundrec:
Will approve when the playback speed is set to 1x rather than 2x!

Hundrec (approved):
Very nice!

image

AbnormalPoof (approved):
works

cyn0x8 (approved):

Issue #4153, opened 2025-02-16
Bug Report: In the editor, when moving notes and having the cursor outside of the lanes, the note move sound will play


amyspark-ng:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

If you're selecting notes as to move them, and your cursor is outside of the lane in the editor, the note move sound will still play, even tho no notes were moved, my guess is that it checks distance rather than checking state of notes

https://github.com/user-attachments/assets/d7304f13-6bc6-4fe3-ab17-b25eb3a62a19

### Steps to Reproduce

1. Open the editor
2. Select notes
3. Grab one and try to move it outside of the editor

Lasercar:
Hmm, I think you forgot to record audio in your videos for the 3 issues you opened, lol.

Hundrec:
I hear it. Can you check your volume?

Lasercar:
> I hear it. Can you check your volume?

oh, I see, it's just very, very low volume.

Lasercar:
This issue now has a pending PR.

PR #4157, opened 2025-02-16
fix notelay sound when not in grid

amyspark-ng:
https://github.com/FunkinCrew/Funkin/issues/4153
The editor would play notelay even when the cursor was outside the grid

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -4565,7 +4565,7 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
           if (dragTargetCurrentStep != dragDistanceSteps || dragTargetCurrentColumn != dragDistanceColumns)
           {
             // Play a sound as we drag.
-            this.playSound(Paths.sound('chartingSounds/noteLay'));
+            if (overlapsGrid) this.playSound(Paths.sound('chartingSounds/noteLay'));
 
             trace('Dragged ${dragDistanceColumns} X and ${dragDistanceSteps} Y.');
             dragTargetCurrentStep = dragDistanceSteps;


source/funkin/ui/debug/charting/ChartEditorState.hx
```diff
@@ -4565,7 +4565,7 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
           if (dragTargetCurrentStep != dragDistanceSteps || dragTargetCurrentColumn != dragDistanceColumns)
```
Lasercar, line 4565: Wouldn't it be easier to just do this?
```if (dragTargetCurrentStep != dragDistanceSteps && overlapsGrid || dragTargetCurrentColumn != dragDistanceColumns && overlapsGrid)```

Lasercar: Well sure, but that isn’t too big of a deal because it’ll jump to whatever position your mouse is when it overlaps the grid again.

amyspark-ng: then it won't feel as responsive, is a big change from the previous behaviour, i wish it was as simple 😞

Lasercar: Then perhaps remove first overlap check, that way it only runs when you move it up or down outside the grid.

amyspark-ng: i will check later today! it's a difficult thing to test since it's chart editor and I'd have to build for windows, which takes at least 30 minutes on my laptop for some reason! you know if i can build and play the same when building for neko?

Lasercar:
> it does seem to work now!, the logic might be a little "faulty" in the sense that i have to "recheck" the change in steps but it works
> 
>  2025-02-17_13-37-18.mp4

Oh, uh, you don’t have to rerun the check like that. Skipping setting the stuff entirely makes more sense, does the same thing and makes the code look much cleaner.

amyspark-ng:
you mean like if (!overlarpsGrid) return? that changes completely the behavior!

Lasercar:
Nooo, just this ‘’’if ((dragTargetCurrentColumn != dragDistanceColumns && overlapsGrid) || dragTargetCurrentStep != dragDistanceSteps)’’’ at 4565, that’s it.

amyspark-ng:
ohhh i see what you mean! i will change it right now

amyspark-ng:
while checking if the fix worked, i noticed there was a different bug where an event selected won't move properly if it's not the "leader" item selected

this is a whole different issue that i will report rn

Lasercar:
Thought I might help you find the spots for fixing the new issue you noticed:

https://github.com/FunkinCrew/Funkin/blob/5054c58dca6ab6184b50a290d4d155e7e33f6b16/source/funkin/ui/debug/charting/ChartEditorState.hx#L3743-L3751

https://github.com/FunkinCrew/Funkin/blob/5054c58dca6ab6184b50a290d4d155e7e33f6b16/source/funkin/ui/debug/charting/ChartEditorState.hx#L3801-L3810

https://github.com/FunkinCrew/Funkin/blob/5054c58dca6ab6184b50a290d4d155e7e33f6b16/source/funkin/ui/debug/charting/components/ChartEditorNoteSprite.hx#L176-L199

https://github.com/FunkinCrew/Funkin/blob/5054c58dca6ab6184b50a290d4d155e7e33f6b16/source/funkin/ui/debug/charting/components/ChartEditorEventSprite.hx#L171-L187

amyspark-ng:
i'd like to help on  this issue but moving items is actually very hard! i had to do something like this recently and struggled a lot, in fact, if you do understand how it works i'd appreciate an explanation

What i did was when you're selecting a bunch of notes and you click one, that note is now a "leader note" and for every selected note i store the difference to that "leader note", then the leader note will move to the hovered step and every other selected note will move to the same step, but adding the difference, so they keep the respective offset

will open the issue in a bit

amyspark-ng:
is this good? can it be merged pretty please

EliteMasterEric (approved):
Seems like a good change

amyspark-ng:
im just seeing this now lol, yippee

Issue #4155, opened 2025-02-16
Bug Report: In the editor, song will not play at all if playback speed is below 0.1 (will also stutter)


amyspark-ng:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

If you try to put the playback speed below or equal to 0.1, the editor will start to stutter and the song will basically not play, even tho the editor is still moving, very slowly
https://github.com/user-attachments/assets/d15f1738-bd5a-4f78-acd5-6481c917c88a

### Steps to Reproduce

1. Open the editor
2. Put playback speed below or in 0.1

Lasercar:
Ok, now you're just finding these chart editor bugs on purpose, aren't you? /j

This seems to be an issue with the easeSongToScrollPosition function? Or the scrollPositionInPixels, since holding alt makes the issue apply to the gridplayhead instead.

It might just be because it has a decimal point cutoff or something.

Lasercar:
Now has a pending PR.

amyspark-ng:
I'm not! i was just messing around and found them 😭

Lasercar:
> I'm not! i was just messing around and found them 😭

I know! I jest. That’s what the /j is for.

amyspark-ng:
/jest....

Hundrec:
Looks like this was resolved internally
Once we're sure it's good to go, I'll close the linked PR.

PR #4162, opened 2025-02-17
Chart editor low playspeed song not playing/stuttering fix

Lasercar:
https://github.com/FunkinCrew/Funkin/issues/4155
Low playspeeds cause the audiotrack's time to sometimes be set ~10ms before somehow. To fix this, a global float stores the previous time used to update the conductor and it'll be used instead for that if the audiotrack time is lower than it. 

This isn't a perfect fix ofc, but it's better than the song stuttering because it goes back in time.

https://github.com/user-attachments/assets/82b6a600-291f-4414-9c24-e05e526c6c54

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -728,6 +728,11 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
   // Audio
 
   /**
+   * This is used rather than the audiotrack time if it's higher. Low playspeeds cause the audiotrack's time to sometimes be set ~10ms before somehow.
+   */
+  var oldTime:Float;
+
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
It works!

Hundrec:
Note: Looks like this issue was resolved internally. Once we're sure things are working, I'll close this PR.

Hundrec:
I'm preeetty sure we're good with this issue. Can you test to be sure?

Lasercar:
> I'm preeetty sure we're good with this issue. Can you test to be sure?

Sure.

Tested the latest version of develop and the audio track no longer does mini backwards time travel at low playspeeds.

I've also tested the PR on the latest develop and for second I thought it was still broken, but it was checking equal times against eachother and running the other conductor update with the trace.

Issue #4165, opened 2025-02-18
Bug Report: chart editor gets song id incorrectly


cyn0x8:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Compiled from GitHub Source Code

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

when opening a song via chart editor, the editor incorrectly assumes the song id is always the same as the song name in lower-kabab-case

```haxe
var currentSongId(get, never):String;

function get_currentSongId():String
{
  return currentSongName.toLowerKebabCase().replace(' ', '-').sanitize();
}
```

this should instead be the default case when the song has no explicit id in the chart manifest...

to avoid name collisions with other mods, i usually prefix the song id with something specific to the mod (like `exampleMod.exampleSong`)

ill try to work on a pr to fix this

### Steps to Reproduce

1. make a scripted song, with an id that isnt lower-kebab-case, that traces something to console
2. optional: load the song by id (via freeplay or directly, without chart editor) and then see the song script output
3. open the song in the chart editor and then play it
4. observe console, there wont be any output from the song script

Lasercar:
This issue has a pending PR.

PR #4168, opened 2025-02-18
[BUGFIX] chart editor song id fix

cyn0x8:

solves #4165

the chart editor should now attempt to use the song id from the song manifest instead of only using the song name in lower-kebab-case (the default if none found)


lunacynne:
one questionn
is this how submodules are supposed to be dealt with in prs? should there be a separate pr in the submodule instead? im just curious since the diff only shows the submodule commit change instead of the actual changes

Hundrec:
It is pretty confusing, but I think including the submod change + opening a funkin.assets PR is the best way to go about it
That way others can easily playtest, and we can dummy-merge both PRs just fine

PR #3720, opened 2024-10-17
[ENHANCEMENT] More Stage Data additions + Grammatical fix

AbnormalPoof:
## Does this PR close any issues? If so, link them below.
Implements/Fixes #3719 (For scroll factor only)

## Briefly describe the issue(s) fixed.
This PR adds 3 new properties to stage data, specifically for characters: `angle`, `scroll`, and `alpha`.
These behave basically the same as how you would use them on a prop.

I also fixed a grammatical error in one of the comments for `StageData.hx`

## Include any relevant screenshots or videos.
<sub>Dinnerbone.</sub>
screenshot-2024-10-17-10-03-52
screenshot-2024-10-17-10-03-58

PR #4171, opened 2025-02-18
Quick `develop` branch fix (again)

AbnormalPoof:
I fucked up the scrollfactor by having the default be `[0, 0]` instead of `[1, 1]`. This made the characters move all weird as seen in https://github.com/FunkinCrew/Funkin/pull/3720#issuecomment-2664816013

My bad!!!


Hundrec:
The result is hilarious, thanks for making this happen 😆

Issue #3167, opened 2024-09-13
Enhancement: Add gunshot SFX when selecting a song as pico


JackXson-Real:
### Issue Checklist

- [X] I have properly named the enhancement
- [X] I have checked the issues/discussions pages to see if the enhancement has been previously suggested

### What is your suggestion, and why should it be implemented?

Read title, it just feels weird to not have SFX for it, but maybe it's just because I'm used to the teaser video since that had SFX.

Lasercar:
Now has a pending PR.

JackXson-Real:
Closing this cause this is kinda pointless and I’ve changed my mind about it 😭

PR #4189, opened 2025-02-20
Pico freeplay - gunshot SFX on song confirm

Lasercar:
https://github.com/FunkinCrew/Funkin/issues/3167
When a song is confirmed, if the character's pico, it'll play one of the 4 shot sfxs in weekend 1, rather than the song confirm sfx.

https://github.com/user-attachments/assets/57db4192-09da-4d4d-8cf2-0edb3b0359e2

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -1958,7 +1958,13 @@ class FreeplayState extends MusicBeatSubState
     }
 
     // Visual and audio effects.
-    FunkinSound.playOnce(Paths.sound('confirmMenu'));
+    if (currentCharacterId == 'pico')
+    {
+      // Gunshot SFX when confirming a song as pico, why not?
+      FunkinSound.playOnce(Paths.soundRandom('shot', 1, 4, 'weekend1'));
+    }
+    else
+      FunkinSound.playOnce(Paths.sound('confirmMenu'));
     if (dj != null) dj.confirm();
 
     grpCapsules.members[curSelected].forcePosition();


Hundrec:
I like the idea, but the sound effect from the [teaser trailer](https://www.youtube.com/watch?v=y5HB2WZE27Q) is what I had in mind

Lasercar:
> I like the idea, but the sound effect from the [teaser trailer](https://www.youtube.com/watch?v=y5HB2WZE27Q) is what I had in mind

Ohhh uhm, does that sound sfx exist in the files somewhere...?

Lasercar:
Sounds like we need to ask if the devs have the sound sitting in a folder on a hard drive somewhere

Hundrec:
Now that I think about it, that exact sound might be too loud and surprising to be a select sound
Maybe a less jumpscare-like sound effect would do for this idea

JackXson-Real:
I really wish we had the sfx used in the trailer. Im not too much of a fan of the single loud gunshot, especially since there are multiple gunshots shown in the animation. Maybe try and change it to be a bit more similar to the trailer? I'd recommend choosing a different sound, and also having it play multiple times.

JackXson-Real:
> > Now that I think about it, that exact sound might be too loud and surprising to be a select sound Maybe a less jumpscare-like sound effect would do for this idea
> 
> Well it could be played at a lower volume, like 25%.
> 
> Here's a video with it playing at 25% volume:
> 
>  2025-02-20.19-37-21.mp4

Its still kind of sudden and loud imo

Lasercar:
Alright then, I'll set it even _lower_.

JackXson-Real:
> Alright then, I'll set it even _lower_.

I think it’s just an issue with the sound itself, I’d recommend trying a different one. Lowering the volume wont make much of a difference.

JackXson-Real:
https://samplefocus.com/samples/machine-gun-sound-multiple-shots
@Lasercar This is an example of one that would fit better. The sound itself is not as deep.

Lasercar:
> https://samplefocus.com/samples/machine-gun-sound-multiple-shots @Lasercar This is an example of one that would fit better. The sound itself is not as deep.

Ah, I see. Well, let's not clog up the comments on this PR with this then. If you can somehow find the exact sound sfx they used then comment again?

trayfellow:
I think it would be better if stuff like these be soft-codable, accessed thru playable character jsons and all that good stuff

Lasercar:
> I think it would be better if stuff like these be soft-codable, accessed thru playable character jsons and all that good stuff

Sure, I'd only need some kind of existing and similar enough implementation that I could look at to figure out how to do it.

MaybeMaru:
I think it would be cooler if instead of replacing the confirm sound it played 4 gunshot sounds synced with the animation like in the trailer. Just slapping a gunshot sound feels a bit sloopy

JackXson-Real:
> I think it would be cooler if instead of replacing the confirm sound it played gunshot sounds synced with the backplate animation like in the trailer. Just slapping a gunshot sound feels a bit sloopy

I agree

Hundrec:
Closing because there are too many issues with an actual implementation of this, regardless of which sound is used.

PR #4203, opened 2025-02-22
Chart editor - ctrl click on hold note null reference fix

Lasercar:
Ctrl clicking on a hold note causes a null reference error because it tries to check if a regular note is selected, rather than a hold note.

This simply changes it to check for a hold note instead.


JVNpixels:
You should add some screenshots/videos to this PR.

Hundrec (approved):
No more crash, very good

AbnormalPoof (approved):
works

PR #4210, opened 2025-02-23
[CLEANUP] Remove unused code from `FreeplayState`

AbnormalPoof:
For some reason there's this useless boolean in Freeplay?? It doesn't do anything so this PR removes it.


Lasercar:
But what if, rather than removing it, you instead tried to find a use for it before removing it? /j

charlesisfeline:
i remember this being used in older fnf builds all the way to v0.2.8

AbnormalPoof:
> But what if, rather than removing it, you instead tried to find a use for it before removing it? /j

`FEATURE_DEBUG_FUNCTIONS` and `Constants.DEBUG_BUILD` in the corner:

Hundrec (approved):
True...

PR #4223, opened 2025-02-24
[FEATURE] Mod Menu

KoloInDaCrib:
## Note
UI is inspired by lemz' take of a mod menu however it has some additions.

## The menu
This mod menu allows you to not only load and unload specific/all mods, but to also view some information about the mod, like the description or the list of contributors.

## Footage

https://github.com/user-attachments/assets/8573a60b-db12-43ae-9c7f-6f9caedeef85

### DIFF
--- a/source/Main.hx
+++ b/source/Main.hx
@@ -46,8 +46,7 @@ class Main extends Sprite
     funkin.util.logging.AnsiTrace.traceBF();
 
     // Load mods to override assets.
-    // TODO: Replace with loadEnabledMods() once the user can configure the mod list.
-    funkin.modding.PolymodHandler.loadAllMods();
+    funkin.modding.PolymodHandler.loadEnabledMods();
 
     if (stage != null)
     {
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -43,7 +43,7 @@ class PolymodHandler
   /**
    * Where relative to the executable that mods are located.
    */
-  static final MOD_FOLDER:String =
+  public static final MOD_FOLDER:String =
     #if (REDIRECT_ASSETS_FOLDER && macos)
     '../../../../../../../example_mods'
     #elseif REDIRECT_ASSETS_FOLDER
@@ -414,8 +414,7 @@ class PolymodHandler
     Polymod.clearScripts();
 
     // Forcibly reload Polymod so it finds any new files.
-    // TODO: Replace this with loadEnabledMods().
-    funkin.modding.PolymodHandler.loadAllMods();
+    funkin.modding.PolymodHandler.loadEnabledMods();
 
     // Reload scripted classes so stages and modules will update.
     Polymod.registerAllScriptClasses();
--- a/source/funkin/ui/debug/mods/ModsSelectState.hx
+++ b/source/funkin/ui/debug/mods/ModsSelectState.hx
@@ -1,0 +1,155 @@
+package funkin.ui.debug.mods;
+
+import flixel.FlxG;
+import funkin.audio.FunkinSound;
+import funkin.input.Cursor;
+import funkin.modding.PolymodHandler;
+import funkin.save.Save;
+import funkin.ui.debug.mods.components.ModInfoWindow;
+import funkin.ui.debug.mods.components.ModButton;
+import haxe.ui.backend.flixel.UISubState;
+import haxe.ui.events.UIEvent;
+import haxe.ui.components.Button;
+import haxe.ui.containers.VBox;
+import haxe.ui.containers.windows.WindowManager;
+import haxe.ui.tooltips.ToolTipManager;
+
+@:build(haxe.ui.ComponentBuilder.build("assets/exclude/data/ui/mod-select/main-view.xml"))
+class ModsSelectState extends UISubState
+{
+  var modListLoadedBox:VBox;
+  var modListUnloadedBox:VBox;
+  var modListLoadAll:Button;
+  var modListUnloadAll:Button;
+  var modListApplyButton:Button;
+  var modListExitButton:Button;
+
+  var prevPersistentDraw:Bool;
+  var prevPersistentUpdate:Bool;
+
+  override public function create()
+  {
+    super.create();
+
+    // prev shit copied from latencystate
+    prevPersistentDraw = FlxG.state.persistentDraw;
+    prevPersistentUpdate = FlxG.state.persistentUpdate;
+
+    FlxG.state.persistentDraw = false;
+    FlxG.state.persistentUpdate = false;
+
+    Cursor.show();
+    WindowManager.instance.reset();
+    WindowManager.instance.container = windowContainer;
+
+    for (mod in PolymodHandler.getAllMods())
+    {
+      var loaded = Save.instance.enabledModIds.contains(mod.id);
+
+      var button = new ModButton(mod);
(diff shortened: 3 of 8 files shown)


Starexify:
Why are contribuitors in a button? 😭 There's so much space left aswell for the mods page to look bigger

KoloInDaCrib:
> Why are contributors in a button? 😭 There's so much space left aswell for the mods page to look bigger

Because it opens a different window and I'm not sure how long a typical mod's description is - could be only a sentence or an entire paragraph long, which is why the description takes a lot of space

Starexify:
> > Why are contributors in a button? 😭 There's so much space left aswell for the mods page to look bigger
> 
> Because it opens a different window and I'm not sure how long a typical mod's description is - could be only a sentence or an entire paragraph long, which is why the description takes a lot of space

Place it over Dependencies , not in Description

KoloInDaCrib:
> Also would suggest adding a new menu entry in the main menu for mods instead, since I don't see them as options really

I'm sorry but that would require me to draw an asset for the main menu and not only would it be out of place but it would also probably not be something typically in line for a *code* pr
Besides, the currently unused mods menu is also in options

lemz1:
Wait a minute, fuck our repos are linked

KoloInDaCrib:
> Wait a minute, fuck our repos are linked

Is it cause of that one time we collabed on the character creator?

lemz1:
> Is it cause of that one time we collabed on the character creator?

yup

Lasercar:
The unused mod menu class could probably be deleted in this https://github.com/FunkinCrew/Funkin/blob/develop/source/funkin/ui/options/ModMenu.hx

cyn0x8:
this looks awesome

a couple suggestions i think would be cool!!
- mod view as just a single full panel instead of windowed (i dont think ppl really need to open multiple)
- currently viewed mod highlighted in list (dependencies too in another color?)
- mod priority order buttons in list instead of window (and/or drag&drop order?)
- dependencies cant go after mods that require them
- preserve mod priority even when enabling/disabling
- description and contributors as tabs for the large text field instead of popup window
- priority goes first-bottom last-top instead of vice versa (more intuitive for dependencies to be visually "below" methinks)
- open a dependency into the mod view when clicked on in dependency tab

amyspark-ng:
image
i thought you could double click these files to open them, that would be crazy

cyn0x8:
> i thought you could double click these files to open them, that would be crazy

that would be cool but it probably wouldnt work with zipped mods?

AbnormalPoof:
It could work but it'd require decompressing the zip like how FNFCs do it probably

KoloInDaCrib:
I've added most of the requested features above + some of my own (for example disabling a dependency disables all of the mods that depend on it)

If there are any more critiques or suggestions, let me know!

https://github.com/user-attachments/assets/536b0884-d5aa-40c8-8f40-146ad3e9b084

Starexify:
> If there are any more critiques or suggestions, let me know!

So I would say some additions that could be good if they aren't already in the PR:
* Resizeable mod sidebar, so you can adjust it by dragging it left and right and it would make the mod info side bigger and list smaller, so you can see more from the mod (tho it's just a tweak)
* Would be cool if you would have backgrounds behind the dark bg of the mod info side
* A button to disable/enable the mod from the mod info panel (near/under the Visit Homepage)
* Visit Source button like for GitHub or stuff like that
* Visit Wiki button for a wiki of the mod
* I would say the Files should be moved on right side and be a panel that you can open or close

I did this edit if it would help give you an idea of how it should look like in my head:
image

cyn0x8:
> * Would be cool if you would have backgrounds behind the dark bg of the mod info side

> * Visit Source and Visit Wiki buttons like for GitHub and Wiki or stuff like that

these would both probably require either changes to polymod (cuz `ModMetadata`) or some new "mod registry" thingy? but idk which would be better

> Also DO NOT enable/disable the dependent mods if the dependency is enabled/disabled(if required do disable), that would make people not be able to turn off dependent mods (which should let people have disabled dependent mods), they will be stuck with them unless they move them out of the folder/deltete (if you added that)

also this confuses me... this is kinda how most dependency systems in games are supposed to work

if a mod is a required dependency of another and is disabled then the dependent mod should also be disabled as it cant reliably run without it and might error or crash or whatever

and if a mod has required dependencies and is enabled it should also enable those dependencies for the same reason as above (or stay disabled/do nothing if one or more arent found in the list)

cyn0x8:
also some more suggestions
- a bit darker highlight/lighter mod entry border for visual clarity/consistency
- idk if im viewing it wrong but just in case!! dependency colors should be just relative only to whatever mod is currently being viewed so theres no ambiguity as to who is dependent to who
- file tree ui should be much more compact (less passing, text smaller idk) as itll feel kinda cramped otherwise
  - should also be scrollable both horizontally and vertically or itll be impossible to navigate to certain parts

Starexify:
> and if a mod has required dependencies and is enabled it should also enable those dependencies for the same reason as above (or stay disabled/do nothing if one or more arent found in the list)

Alright so what I mean:
- If mod has required dependencies -> enable/disable the mod that needs them, not the dependencies

- If mod has optional dependencies -> don't enable/disable them, only the mod 

- If required dependency (mod2) is enabled -> don't enable mod1 that requires it, because the dependency (mod2) doesn't have the mod as dependency

- But if required dependency is disabled -> disable the mod that needs it

- If optional dependency is enabled/disabled -> don't enable/disable the mod

cyn0x8:
> Alright so what I mean:
> 
> * If enabled/disabled mod has required dependencies -> enable/disable the mod that needs them, not the dependencies
> 
> * If enabled/disabled mod has optional dependencies -> don't enable/disable them, only the mod
> 
> * If required dependency (mod2) is enabled -> don't enable mod1 that requires it, because the dependency (mod2) doesn't have the mod as dependency
> 
> * But if required dependency is disabled -> disable the mod that needs it
> 
> * If optional dependency is enabled/disabled -> don't enable/disable the mod

ohhh wait the confusion was from how "dependency" was defined lol

i agree with that

i meant "dependencies" of a mod as in the list of mods it requires (optionally or not) to function
just how the word is used in the polymod json

KoloInDaCrib:
I'm getting mixed signals here. Should I change the system or no?

Reminder, the system currently works as follows:
1) Let's assume modA depends on modB to work
2) When modA gets enabled, it also automatically enables modB, if it's detected by Polymod as a valid mod, since modA cannot function without modB
3) When modA gets disabled, it has no effect on modB
4) When modB gets enabled, it has no effect on modA
5) When modB gets disabled, it automatically also disables modA, since modA cannot function without modB

Also, I hope I'm not asking for much, but I would appreciate a concrete list of features since it appears everyone has differing opinions ~~not like this gonna get reviewed by eric or cam any time soon and thus I cannot prioritize working on any features besides the ones that were requested here~~
Reminder that you can compile the branch from this PR and check it out for yourself, since I haven't shown everything (for example, there already exists a feature to add a label with the role under a contributor, it's just that neither test mods have much detail about contributors and thus the label isn't present)

Hundrec:
Uh oh, merge conflicts in da Mod Menu

MightyTheArmiddilo:
Hi just had a suggestion. Could the mod list be moved or something so people with debug display (like me) could see it better? (This would especially be nice for people who have the display alpha at 100%) Also this is great stuff.

Hundrec:
I will have to close this because we plan to design and implement our own Mod Menu at some point in the future.
Thank you for opening this PR!
KoloInDaKrib

PR #4230, opened 2025-02-26
[ENHANCEMENT] ModStore

cyn0x8:
adds a simple helper class for scripts to store arbitrary data during runtime

this is different from storing data in modules as this data will persist as long as the game is open, and is also different from `Save.instance.modOptions` in that its intended to be temporary and also compatible with any type

ive frequently encountered scenarios where certain things need to persist through mod reloads, and made a kinda complex [workaround script](https://github.com/rizz-residence/beast-engine/blob/main/scripts/be/reloader/Reloader.hxc) to compensate, but itd be a lot nicer to have this class to use instead!!

<details><summary>it works well</summary>

</details>

### DIFF
--- a/source/funkin/modding/ModStore.hx
+++ b/source/funkin/modding/ModStore.hx
@@ -1,0 +1,59 @@
+package funkin.modding;
+
+import haxe.ds.StringMap;
+
+/**
+ * Temporary persistent data storage for mods to use.
+ */
+class ModStore
+{
+  /**
+   * All registered stores for this runtime.
+   */
+  public static final stores:StringMap<Dynamic> = new StringMap<Dynamic>();
+
+  /**
+   * Registers a new store with the given ID.
+   * If a store with the same ID already exists, it will be returned instead.
+   * 
+   * @id The ID for this store.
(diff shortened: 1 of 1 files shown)


MAZ12211:
Wtf, beast engine?

source/funkin/modding/ModStore.hx
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Temporary persistent data storage for mods to use.
+ */
+class ModStore
+{
+  /**
+   * All registered stores for this runtime.
+   */
+  public static final stores:StringMap<Dynamic> = new StringMap<Dynamic>();
```
AbnormalPoof, line 13: Is this supposed to be a final? Considering you're modifying it.

lemz1: the final just means that we can't reassign the stores map.
we can still modify the values inside of the stores map

AbnormalPoof: Ah, my bad.

AbnormalPoof (approved):

PR #4247, opened 2025-03-02
[ENHANCEMENT] Chart Editor - Toggle audio volume hotkeys

### LINKED ISSUE #2731: Enhancement: Option to toggle vocals/hitsounds
## What is your suggestion, and why should it be implemented?

In the chart editor there should be a button to toggle easily the hitsounds or vocals in a song, since it's kinda annoying to have to drag them everytime, i want them to be either 100 or 0, so yeah

image

Lasercar:
Closes #2731
Adds hotkeys for toggling the volume of everything in the audio menu.

Shift + M for metronome
Shift + H for hitsounds
Shift + I for instrumental
Shift + V for vocals
Shift + B for Boyfriend vocals
Shift + D for Dad vocals

https://github.com/user-attachments/assets/ebe154ec-efc1-45c8-b762-bdc766cd7609

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -5879,80 +5879,68 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
       // Changing values of the audio slider directly because it'll update the audio anyway and makes this code much cleaner, though more verbose.
       var oldValue = previousAudioVolumes[0];
       previousAudioVolumes[0] = menubarItemVolumeMetronome.value;
-      if (oldValue > menubarItemVolumeMetronome.value) menubarItemVolumeMetronome.value = oldValue;
-      else
-      {
-        menubarItemVolumeMetronome.value = (menubarItemVolumeMetronome.value == 0) ? 100 : 0;
-      }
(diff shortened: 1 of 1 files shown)


amyspark-ng:
does this account for a previous value, like if you had it at 37, takes it to 0 and when you do it again does it take you to 37? or to 100

Lasercar:
> does this account for a previous value, like if you had it at 37, takes it to 0 and when you do it again does it take you to 37? or to 100

No

I don't like the idea of adding a variable that's only use once. I could do it anyway, or

> I currently have it set so that it doesn't invert unless it's at 0 or 100. I could instead make it invert at any number, so a volume above 50 gets set to zero, while a volume below 50 gets set to 100.

Lasercar:
Ok, I chose to use a float array to store the previous values. If the the volume's been set to zero by the save preferences, then the previous value becomes 100, so that it toggles properly from the get-go.

amyspark-ng:
that's nice thank you

Hundrec (changes requested):
It's great! The only thing I'd suggest is to mute both hitsounds if either hitsound's value is above 0%, and then restore both to their most recently saved values above 0.

https://github.com/user-attachments/assets/c9e8ef1b-e7c2-4316-a489-1721e7580d6d

Hundrec (approved):
Took me 5 months to decide, but I think this is perfect!

PR #4248, opened 2025-03-02
Chart editor - Custom note kind field fix

Lasercar:
The custom would only get hidden if the toolbox note kind was changed from custom by hand. Now gets hidden if the note selected isn't a custom one. 

https://github.com/user-attachments/assets/fc2fe197-7987-402a-9db9-72cfe887614e

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -6,7 +6,6 @@ body:
   - type: markdown
     attributes:
       value: "# PLEASE READ THE [CONTRIBUTING GUIDE](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md) BEFORE OPENING ISSUES!"
-  
   - type: checkboxes
     attributes:
       label: Issue Checklist
@@ -49,6 +48,7 @@ body:
       placeholder: ex. 0.5.3
     validations:
       required: true
+
   
   - type: markdown
     attributes:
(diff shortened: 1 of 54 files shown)


Lasercar:
Aw, medium size label? But it's a perfect 10 changes....

Hundrec:
This seems to be how the Chart Editor behaves now anyway.
This PR's behavior might differ from 0.5.3, but does it differ from 0.7.4?

Lasercar:
> This seems to be how the Chart Editor behaves now anyway. This PR's behavior might differ from 0.5.3, but does it differ from 0.7.4?

Seems like some change has broken the custom kind type, setting anything to it then changing the text in the box doesn't change the note's kind to the text you inputted, that and clicking on a note set to a unknown custom type resets it to no kind.

After looking into it more, I think it's always been broken...

Ah, it's doing the weird bug again, the dropdown value cannot be updated at all?

Lasercar:
> Now when you select an note with a note kind, then Ctrl-click on another note, it changes all the notes to have the note kind of the most recent thing you clicked.
> 
> It should only mutate the note kind if you mess with the dropdown in the Notes window.

Whoops, I completely forgot to fix this again.

The change I made to initalise null toolboxes that get refreshed broke the failsafes that prevent events/notes from being modified by the toolbox when they get first opened. 

It was supposed to fix a crash or something but reverting it back to just returning doesn't seem to have caused any problems so I'm not sure why I did the change in the first place (or why I previously tried to find a fix without reverting it).

Lasercar:
Wait, hold on, I'll give that commit a rename

Lasercar:
Rebased to remove an unnecessary commit

PR #4264, opened 2025-03-04
[ENHANCEMENT] Chart Editor - Snap selection to current note snap

### LINKED ISSUE #4122: Enhancement: All Notes Snap Selection Move
### Issue Checklist

- [x] I have read the pinned [Issues Guide](https://github.com/FunkinCrew/Funkin/issues/4050)
- [x] I have properly named my enhancement

I'm suggesting an way to select a bunch of notes, and then when you move them, it does the note snapping to *every* note, instead of just the one you use as the anchor point for dragging the notes. This allows you to easily resnap a large amount of notes. The most obvious use case is to fix when you realize you got a songs BPM wrong by like 1 beat and you've already charted a bunch.

Lasercar:
Closes #4122
Adds a command and hotkey (Shift + N) to snap the current selection to the current note snap. 

>[!IMPORTANT]
> Requires https://github.com/FunkinCrew/funkin.assets/issues/335

Now with Buttons!

https://github.com/user-attachments/assets/3992a99c-0277-4bae-8d23-bef8a832f15e

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -1937,6 +1937,11 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
   var menubarItemDelete:MenuItem;
 
   /**
+   * The `Edit -> Snap Selection` menu item.
+   */
+  var menubarItemSnap:MenuItem;
+
+  /**
    * The `Edit -> Delete Stacked Notes` menu item.
    */
   var menubarItemDeleteStacked:MenuItem;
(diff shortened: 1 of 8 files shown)


Lasercar:
~~Hmm, I wonder if I should modify this to allow the note snap to be passed into the command instead~~ YES

And I've just realised that I'd have to add a button for every single note snap ratio to make it useful... Maybe I'll do that later

EliteMasterEric (changes requested):
Went to review this PR and it has a whole bunch of conflicts!

Lasercar:
Should be all updated now.

Kade-github (changes requested):
While I really like this, I actually think a better version of this would to:
- Whenever scrolling you are snapped to your snap preference at that moment
- Holding ALT (or some other modifier allows for unsnapped movement)

I just don't see a reason for this to exist past this idea (of not even allowing unsnapped notes to begin with)

Lasercar:
> While I really like this, I actually think a better version of this would to:
> 
>     * Whenever scrolling you are snapped to your snap preference at that moment
> 
>     * Holding ALT (or some other modifier allows for unsnapped movement)
> 
> 
> I just don't see a reason for this to exist past this idea (of not even allowing unsnapped notes to begin with)

Ok, first point, a selection can't be reorganised on the fly like that easily or cleanly. I think most might end up setting the snap too small and break their selection positioning accidentally / in an unintended way.

Second point is for a completely different change, not related to this simple command at all (it's out of scope!).

PR #4270, opened 2025-03-06
[BUGFIX] Funkin sound restarting music fix

### LINKED ISSUE #3656: Bug Report: Freeplay song preview restarts unnecessarily when switching difficulties if one difficulty has unique songs
### Issue Checklist

### Platform

### Browser

None

### Version

0.5.2

0.5.2 implemented a fix to prevent the freeplay song preview from restarting when switching difficulties that are not variations. This behavior works correctly under normal circumstances. However, if certain difficulties have songs that are exclusive to them (such as a song with a normal chart only), the song preview restarts when switching between them, even though the variation is the same. Currently, all official songs use every available difficulty for their respective variations, but modded songs may not.

The video below demonstrates the issue. The mod used as an example here features normal and hard charts, but no easy charts. Notice how the preview doesn't restart when I switch between normal and hard, but it does when I switch between normal and easy. I tested this on multiple mods and the issue persists across several of them.

https://github.com/user-attachments/assets/f77f60fd-0567-4cfb-babc-48bfe5e347b6

This is not an urgent issue, as the only way you can observe this is by modding the game, but it does seem like unintended behavior.

### Steps to Reproduce

_No response_

### LINKED ISSUE #4821: Bug Report: Freeplay song preview not restarting no longer works
### Issue Checklist

### Platform

### Browser

None

### Version

0.6.3

Since version 0.5.2, switching between difficulties within the same variation wouldn't restart the song preview (#3587). In version 0.6.3, this no longer works:

https://github.com/user-attachments/assets/e2c3ee65-5669-4973-9933-7d2c0d954880

Maybe it has to do with using the animation when switching difficulties?

### Steps to Reproduce

1. Go to the Freeplay menu
2. Switch between difficulties for a song in the same variation
3. The song preview will always restart

Lasercar:
Fixes #3656, fixes #4821
The funkin sound play music function does not check against music suffixes, leading to it restarting the current music if one was passed when the function is called.

That, and the load partial doesn't set the label to the path given afterwards, so the label ends up being 'unknown' and causes the preview to still be reset.

### DIFF
--- a/source/funkin/audio/FunkinSound.hx
+++ b/source/funkin/audio/FunkinSound.hx
@@ -545,7 +545,8 @@ class FunkinSound extends FlxSound implements ICloneable<FunkinSound>
 
       soundRequest.future.onComplete(function(partialSound)
       {
-        var snd = FunkinSound.load(partialSound, volume, looped, autoDestroy, autoPlay, false, onComplete, onLoad);
+        var snd:Null<FunkinSound> = FunkinSound.load(partialSound, volume, looped, autoDestroy, autoPlay, false, onComplete, onLoad);
+        if (snd != null) snd._label = path;
         promise.complete(snd);
       });
     }
--- a/source/funkin/audio/FunkinSound.hx
+++ b/source/funkin/audio/FunkinSound.hx
@@ -288,13 +288,25 @@ class FunkinSound extends FlxSound implements ICloneable<FunkinSound>
   {
     if (!(params.overrideExisting ?? false) && (FlxG.sound.music?.exists ?? false) && FlxG.sound.music.playing) return false;
 
+    var pathsFunction = params.pathsFunction ?? MUSIC;
+    var suffix = params.suffix ?? '';
+    var pathToUse = switch (pathsFunction)
+    {
+      case MUSIC:
+        Paths.music('$key/$key');
+      case INST:
+        Paths.inst('$key', suffix);
+      default:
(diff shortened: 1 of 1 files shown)


AbnormalPoof (changes requested):
... Are you sure this works?

https://github.com/user-attachments/assets/90f99a8a-81e6-4300-861e-b900c27f2c72

Lasercar:
Hmmmmmmmmmm
I guess I'm going to need to get an example to try this myself.

Lasercar:
Wait, hold on, it still isn't fixed properly I think?

AbnormalPoof (approved):
Works

Lasercar:
Also fixes #4821

https://github.com/user-attachments/assets/c30f07a9-6cdb-4734-a4e1-01347f3faa28

Hundrec:
Freeplay song previews do properly continue across difficulties now, so does this PR still fix an issue?

Lasercar:
> Freeplay song previews do properly continue across difficulties now, so does this PR still fix an issue?

The issue is 'fixed' in the freeplay state, but the actual issue, loaded partical music being labeled as unknown (and therefore always been replayed when the same music track is played) and music with suffixes not being properly checked for, as it only compares the current music's label against the path given in the params, rather than the path it actually uses.

Lasercar:
Looks like someone added my fix in without merging this, though turns out they forgot one tiny thing

Issue #4274, opened 2025-03-07
Enhancement: catch shader errors


cyn0x8:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

currently shaders will fully crash the game when an error happens; itd be nice if they were caught like hscript errors are

psych engine does this now so it is something thats possible to do !!

PR #4275, opened 2025-03-08
[BUGFIX] Prevent the game from crashing upon a shader error

KoloInDaCrib:
Fixes https://github.com/FunkinCrew/Funkin/issues/4274

OpenFL's logging system, which was responsible for throwing errors upon some bugs either with asset loading or shaders, had a nifty little variable called `throwErrors` which was set to true by default. This PR sets the variable to false. This should also hopefully fix several issues, I'm just too lazy to search for which issues in particular myself lmao

https://github.com/user-attachments/assets/ae1d84a7-096c-4f79-b395-480e5bee8768

(reminder: the stage editor shader would normally crash the game)


AbnormalPoof:
I feel like a better way to do this would be to catch the shader error directly and show it in a prompt using `PolymodErrorHandler`'s `showAlert()` function. Disabling its error throwing entirely could have some consequences.

KoloInDaCrib:
> I feel like a better way to do this would be to catch the shader error directly and show it in a prompt using `PolymodErrorHandler`'s `showAlert()` function. Disabling `Log`'s error throwing entirely could have some consequences.

Good point, however, with the variable disabled in this PR, the errors are printed directly on the console without calling haxe's `trace` (the class has it's own trace function 💔 ) and without a way to obtain the error message besides making an additional PR in the funkin fork of the `lime` library (since openfl's log is quite literally just an extension of lime's log, if the library `lime` is provided in the project)

EliteMasterEric:
This issue is quite problematic right now. Without a functionality to display a pop-up for the shader error, the user will simply experience incorrect functionality without any explanation.

I am good with displaying the shader error message and then continuing the game (since a shader error will typically only cause visual issues rather than loss of save) but I will not accept a PR which silently ignores critical errors.

KoloInDaCrib:
I've added the feature requested, though it requires the https://github.com/FunkinCrew/lime/pull/10 PR to be accepted to run, as I've internally added an event in log for it to be possible to obtain the error message

The error dispatched now looks something like this: 
image

Hundrec:
Does this crash still happen?
Anyway there are merge conflicts

KoloInDaCrib:
> Does this crash still happen? 

The game still causes a crash whenever there is any shader error due to how openfl/lime handles logging errors, which is what this PR aims to circumvent, the shader in the video demonstration does not cause crashes any more

AbnormalPoof:
Looks like this PR needs a rebase.

Kade-github (approved):

EliteMasterEric:
After an internal change which greatly improves error handling for shader errors (thus making them easier to fix), we've concluded that letting the game continue running despite an error is too problematic.

PR #4297, opened 2025-03-12
Hit window offset fix

### LINKED ISSUE #2877: Enhancement: Improve offsets for timing judgements
Currently getting a sick rating feels really luck based for me since it's a little bit off. You are able to click an arrow pretty early and get a sick, but clicking it when the arrow is just a tiny bit above the strumline causes it to be just a good. Even though I play the song perfectly on beat, I somehow manage to almost always get 1 or 2 goods due to this issue. I recommend giving a little more leeway when hitting notes a little bit later so that its the same as hitting a note early. I mainly used Psych Engine to play the game before the update came out, and the judgements for sick rating on there feels more consistent than in base game currently. The difference between that and base game is like night and day. I highly recommend making it similar to Psych Engine so that getting a golden P rank for songs is a lot less painful and more consistent.

### LINKED ISSUE #2969: Bug Report: Hit window is offset from strumline

Lasercar:
Fixes #2969, closes #2877
The hit window appeared to be offset, these changes ensure that is no longer the case.

Requires this assets change: https://github.com/FunkinCrew/funkin.assets/pull/142 (otherwise it'll still be offset, lol)

https://github.com/user-attachments/assets/d6b86630-f2be-4d70-b3fd-8149d89cb803

My first perfect gold too, just goes to show that the changes really work! 

I've since fixed the hold notes and hold cover from being offset from the new y offset as seen in the video, too.

A "perfect" hit before the change (aka `Conductor.instance.songPosition > hitWindowCenter`):
Screenshot 2025-03-14 203126

After the change:
Screenshot 2025-03-14 203732
Screenshot 2025-03-14 203748
Screenshot 2025-03-14 203824
Screenshot 2025-03-14 203833

I'd imagine this: https://github.com/FunkinCrew/Funkin/pull/3544 would fix those first two still appearing slightly off though.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -891,34 +891,6 @@ class PlayState extends MusicBeatSubState
       needsReset = false;
     }
 
-    // Update the conductor.
-    if (startingSong)
-    {
-      if (isInCountdown)
-      {
-        // Do NOT apply offsets at this point, because they already got applied the previous frame!
-        Conductor.instance.update(Conductor.instance.songPosition + elapsed * 1000, false);
-        if (Conductor.instance.songPosition >= (startTimestamp + Conductor.instance.combinedOffset))
-        {
-          trace("started song at " + Conductor.instance.songPosition);
-          startSong();
-        }
-      }
-    }
-    else
-    {
-      if (Constants.EXT_SOUND == 'mp3')
-      {
-        Conductor.instance.formatOffset = Constants.MP3_DELAY_MS;
-      }
-      else
-      {
-        Conductor.instance.formatOffset = 0.0;
-      }
-
-      Conductor.instance.update(Conductor.instance.songPosition + elapsed * 1000, false); // Normal conductor update.
-    }
-
     var androidPause:Bool = false;
 
     #if android
@@ -1096,6 +1068,33 @@ class PlayState extends MusicBeatSubState
     if (!isInCutscene) processNotes(elapsed);
 
     justUnpaused = false;
+    // Update the conductor last.
+    if (startingSong)
+    {
+      if (isInCountdown)
+      {
+        // Do NOT apply offsets at this point, because they already got applied the previous frame!
+        Conductor.instance.update(Conductor.instance.songPosition + elapsed * 1000, false);
+        if (Conductor.instance.songPosition >= (startTimestamp + Conductor.instance.combinedOffset))
+        {
+          trace("started song at " + Conductor.instance.songPosition);
+          startSong();
+        }
+      }
+    }
+    else
+    {
+      if (Constants.EXT_SOUND == 'mp3')
+      {
+        Conductor.instance.formatOffset = Constants.MP3_DELAY_MS;
+      }
+      else
+      {
+        Conductor.instance.formatOffset = 0.0;
+      }
+
+      Conductor.instance.update(Conductor.instance.songPosition + elapsed * 1000, false); // Normal conductor update.
+    }
   }
 
   function moveToGameOver():Void
@@ -2525,7 +2524,10 @@ class PlayState extends MusicBeatSubState
 
     // Get the offset and compensate for input latency.
     // Round inward (trim remainder) for consistency.
-    var noteDiff:Int = Std.int(Conductor.instance.songPosition - note.noteData.time - inputLatencyMs);
+    var noteDiff:Int = Std.int(Conductor.instance.songPosition - note.noteData.time);
+    // Correctly apply the input latency
+    if (noteDiff > 0) noteDiff -= Std.int(inputLatencyMs);
+    else if (noteDiff < 0) noteDiff += Std.int(inputLatencyMs);
 
     var score = Scoring.scoreNote(noteDiff, PBOT1);
     var daRating = Scoring.judgeNote(noteDiff, PBOT1);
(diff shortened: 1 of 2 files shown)


JackXson-Real (approved):
Fix works perfect. I'm so glad that someone finally got a fix for this. Been playing on this for a while and it just feels so much better to play. I'm constantly getting new highscores, and getting goods instead of sicks now feels like it's my fault instead of the game's and I no longer need to condition myself to click when the arrow is below the strumline instead of right on it.

JackXson-Real:
@EliteMasterEric Hop on this shit we need this merged immediately 🔥

Hundrec (approved):
I cannot overstate how big of a difference this fix makes in every aspect of gameplay.

The notes feel perfectly responsive, playing along with the song actually hits all Sick ratings, and I'm beating all of my high scores on my first attempts.

As a musician, I feel that this is the true Funkin' experience.

Please bless all Funkin' fans with this wonderful gift.

(and also nothing broke while I tested it)

JackXson-Real:
It’s also worth mentioning that the pixel week offsets have always been correct, so this PR only changes the default note style (for when Eric sees this).

Lasercar:
> It’s also worth mentioning that the pixel week offsets have always been correct, so this PR only changes the offset for the default note style (for when Eric sees this).

That said though, this makes a few changes the score state, namely moving the conductor update to happen _after_ the input and note hits, so that there's no cases of a hit that'd would have been a sick ending up as a good because the conductor's position was updated and now it's outside of the window, and it properly cancels out the input latency too, as before it was always subtracted from the note diff, even if it was negative - so it only further offset the hit window.

Basically what I'm saying is that the pixel week songs will still feel a bit different.

source/funkin/play/PlayState.hx
```diff
@@ -2525,7 +2524,10 @@ class PlayState extends MusicBeatSubState
 
     // Get the offset and compensate for input latency.
     // Round inward (trim remainder) for consistency.
-    var noteDiff:Int = Std.int(Conductor.instance.songPosition - note.noteData.time - inputLatencyMs);
+    var noteDiff:Int = Std.int(Conductor.instance.songPosition - note.noteData.time);
+    // Correctly apply the input latency
+    if (noteDiff > 0) noteDiff -= Std.int(inputLatencyMs);
+    else if (noteDiff < 0) noteDiff += Std.int(inputLatencyMs);
```
cyn0x8, line 2530: the original code for `noteDiff` was correct (red line in the vid)

https://github.com/user-attachments/assets/2791e58b-40b8-4c0a-88ab-ce2c5b86a5a3

this new one makes 2 perfect hit points on positive latency and none on negative

cyn0x8 (changes requested):

source/funkin/play/PlayState.hx
Lasercar: What happens when you use a positive value instead?

In my testing I've never seen a _negative_ input latency value, because that isn't possible in the first place.

Lasercar: Screenshot 2025-03-15 103932
Screenshot 2025-03-15 103952
Screenshot 2025-03-15 104044
Screenshot 2025-03-15 104053

cyn0x8: > In my testing I've never seen a _negative_ input latency value, because that isn't possible in the first place, and it is only added if the note diff is negative.

yeaa it was just an example to show how the math is different

but in the new implementation hitting a note when `songPos` is `note.noteData.time - inputLatencyMs` or `note.noteData.time + inputLatencyMs` will be considered perfect when it should only be the latter

image

source/funkin/play/PlayState.hx
```diff
@@ -891,34 +891,6 @@ class PlayState extends MusicBeatSubState
       needsReset = false;
     }
 
-    // Update the conductor.
```
EliteMasterEric, line 894: The conductor should definitely be processed before the note hit so that the song position used when calculating note hits is as accurate as possible. If you don't update the conductor first, the game will be comparing the time the note was hit with the time the song was at on the PREVIOUS frame.

source/funkin/play/PlayState.hx
Lasercar: The line that goes past the hit window shouldn't be able to happen anyway?

EliteMasterEric: `inputLatencyMs` is the time between when the note was hit and the current time. For example, if the note was hit at 10500 ms and it takes 10 ms to get around to processing the note, the current time will be 10500 ms and `inputLatencyMs` will be 10 ms. You should ALWAYS subtract it, because otherwise the input delay will be DOUBLED.

EliteMasterEric (changes requested):
I don't know that this actually even solves the problems it claims to solve, only makes things worse. The logic behind these changes doesn't really make any sense.

Lasercar:
Alright, then. I'll close this.

source/funkin/play/PlayState.hx
cyn0x8: oh
completely misunderstood what `inputLatencyMs` was oops



maybe i should read the rest of the code next time loll

PR #4177, opened 2025-02-19
[ENHANCEMENT]: Miss Sound Preference

JVNpixels:
## Does this PR close any issues? If so, link them below.
No, this does not close any issues.

## Briefly describe the issue(s) fixed.
I added the option to turn on/off miss sounds, with the default setting being true.
For some people, the miss sound can get annoying, so you can either turn it on or off.

## Include any relevant screenshots or videos.
image
image
https://youtu.be/R4EUZcBPL2g

PR #4302, opened 2025-03-12
[ENHANCEMENT] Better miss sound support

KoloInDaCrib:
It addresses this [comment by cam](https://github.com/FunkinCrew/Funkin/pull/4177#issuecomment-2716033489) that reads

> [...] I think rather than a toggle on/off preference, maybe can have the miss note options be more easily configurable via modding, right now it's hardcoded to only pick randomly between 3 different sounds. So I think rather enabling MORE options like that is the preferred solution.

The system to load the miss sounds is similar to the one to load the sticker sounds. To compensate with that, all miss sounds were moved from `assets/shared/sounds` to `assets/shared/sounds/gameplay/miss` ~~which would need to be changed later for 0.7 since that's when the assets would be reorganized iirc~~. This should also work fine with mods adding additional miss sounds.

https://github.com/user-attachments/assets/1d0106c6-e5d3-4399-9079-39f5eb9462d4

~~the video probably isn't the best way to show it since the scrollMenu sound isn't very loud on top of miss sounds in game being played at half the volume~~


Lasercar:
> > Easy fix - add an option to choose the volume the hitsound is played at!
> 
> Imo that would be a very redundant and specific option, besides, as cam said in the same comment linked above, hearing miss sounds glaring while missing a bunch of notes is a part of The Funkin' Experience™ so having the option to change the volume would also enable making the sounds quieter

Hmm, then make it a volume multiplier instead? (and you can't go below zero?)
I don't really see the issue with adding one, as you can just delete the sound if you don't want to hear it in the first place, but I guess no one really does that anyway.

Hundrec:
Merge conflicts, you know the drill!

Lasercar:
> Merge conflicts, you know the drill!

This is not a drill

TechnikTil:
yucky and evil merge conflicts

NotHyper-474:
I'm sorry, little one.

Closing this due to our 90-day policy for stale PRs that need revision, as stated in the Contributing Guide.

Please let us know if you'd want to bring this back so that we reopen it.

PR #4327, opened 2025-03-16
example of softcoded menu replacements

7oltan:
would be better if it used events but i dont know

it does not but it helps modding greatly

none

none needed


AbnormalPoof:
You can just... make an HScript file that does this without editing the source code.

7oltan:
> You can just... make an HScript file that does this without editing the source code.
how?? i found ways but its really bad on the games memory, since it loads the state then switchs to the cusotm one, im talking about menus in general, like freeplay gameover

PR #4333, opened 2025-03-16
[CHORE] Fix a typo in PlayState.hx

Hundrec:
## change -> chance
change -> chance
* change -> chance


Hundrec:
Nice finds! I've fixed the two typos, but couldn't quite figure out what the unfinished sentences were supposed to say.
We'll have @EliteMasterEric deal with those.

EliteMasterEric:
Oops, if you find an unfinished comment that was probably me! I have a tendency to start a thought and then forget to finish it.

> * "These are encoded with an OS timestamp, so they-" (lines 361 and 367)

- "These are encoded with an OS timestamp, so we can account for input latency."

> * "Respawns notes that were b-" (line 2432)

- "Respawns notes that were between the previous time and the current time when skipping backward, or destroy notes between the previous time and the current time when skipping forward."

PR #4337, opened 2025-03-17
File drop feature flag fix

Lasercar:
No. No one's going to open an issue for this one, trust me.
I noticed that the flag for the file drop feature was outdated, this updates it everywhere it's used and adds it where it currently isn't but should be used too.

However, this PR doesn't fix the issues that I've noticed the file drop system has: 

- Not working when there's something focused or a tooltip/notification on the screen

- Working on any part of the screen, meaning only the last added drop handler works, the rest cannot

- Not being removed after the dialog closes, for example, the vocals/ instrumental drop handler (the vocals one is strange since it _should_ be removing them, but it isn't?)


EliteMasterEric:
Side note, `FEATURE_FILE_DROP` exists because the feature doesn't work on MacOS.

Lasercar:
> Side note, `FEATURE_FILE_DROP` exists because the feature doesn't work on MacOS.

Yeah.
Screenshot 2025-03-18 152220

PR #4350, opened 2025-03-19
Character icon clickable in playstate fix

Lasercar:
The character icon in the chart editor was still clickable in while playstate - it wasn't checking if the chart editor state was actually currently active or not.


Hundrec (approved):
No more lag machine

AbnormalPoof:
death to the lag switch

PR #4352, opened 2025-03-19
[BUGFIX] Fix the Inst and Vocal files not being played when the Sound Channel Limit gets reached

### LINKED ISSUE #2732: Bug Report: enabling practice mode when starting a freeplay song and spamming arrow keys or wasd it stops everything like no music and the notes stop in place
## Describe the bug

the bug is that when enabling practice mode when starting a freeplay song and spamming arrow keys or wasd it stops everything like no music and the notes stop in place

## To Reproduce

start a freeplay song and turn on practice mode and try and spam arrow keys or wasd

## Expected behavior

not sure

## Screenshots/Video

WIINDOWS 11

* Browser
* Version:

v0.4.0

## Additional context

KoloInDaCrib:
Closes #2732 

Introduced the sound channel priority to `FunkinSound` with the default value `false`. Inst and vocal files have this value set to `true`. 
When enabled, this value adds the sound to a priority list, which means that when the sound channels get cleared (which happens before playing another sound) they remain untouched.

https://github.com/user-attachments/assets/dab098d0-0a9c-4720-bde2-616c63e6c3aa

(You can kind of see the effect because the "Go!" sound isn't being played (since it doesn't have the priority) however the music remains intact)

### DIFF
--- a/source/funkin/audio/FunkinSound.hx
+++ b/source/funkin/audio/FunkinSound.hx
@@ -16,6 +16,7 @@ import funkin.Paths.PathsFunction;
 import openfl.Assets;
 import lime.app.Future;
 import lime.app.Promise;
+import openfl.media.SoundChannel;
 import openfl.media.SoundMixer;
 
 #if (openfl >= "8.0.0")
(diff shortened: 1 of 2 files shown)


AbnormalPoof (approved):
Peak

Hundrec (approved):
Awesome!
Spamming in Practice Mode and restarting enough times still crashes the game with a Null Object Reference, but that's a whole other issue

EliteMasterEric:
I'd add the countdown to high priority as well, but otherwise this seems like a good solution (at least unless [OpenFL changes things on their end](https://discord.com/channels/@me/1331411070371496008/1352125270030553090)).

Hundrec:
@EliteMasterEric We found out that the pause screen also crashes pretty easily while spamming, maybe that could be added too?

KoloInDaCrib:
I've added eric's suggestion but couldn't really reproduce any of the crashes mentioned above - the closest issue I could think of that happens is a `RangeError` crash when restarting any song with nene in it (either normal or dark) however that's unrelated to this PR and it may just be on my end since the crash happens primarily from the `funkin.vis` library

Hundrec:
Spam and hit pause, there it is!

https://github.com/user-attachments/assets/1109c815-1e7d-4c3d-a573-ec47c2d6b428

KoloInDaCrib:
Changed the system for one last time (hopefully) so that instead of subtracting sounds from the array (which turned out to be more convoluted than anticipated), the sound's channel is forcefully pushed onto the array if the sound is marked as important. Hopefully that doesn't cause any further crashes since testing with this method worked without problems

Hundrec (approved):
It works!

AbnormalPoof (approved):
Works

Kade-github:
accepted into 0.6.4.

good changes!

PR #4370, opened 2025-03-23
Remove `Alphabet`

AbnormalPoof:
Old unused class from legacy that has no use now. `AtlasText` is far better.

o7


Hundrec (approved):
Yep, builds fine!

image

Lasercar (approved):
I was going to null safety it, but then I realised that it was deprecated

moxie-coder:
kinda unrelated but I wish psych could use this instead because it does seem better

Keoiki:
rip bozo

Lasercar:
> kinda unrelated but I wish psych could use this instead because it does seem better

https://github.com/Psych-Slice/P-Slice

Michu0204:
> Yep, builds fine!
> 
> image

Why does bro have all L's 😭

Lasercar:
> > Yep, builds fine!
> >
> 
> Why does bro have all L's 😭

Funy

Issue #3205, opened 2024-09-13
buguei os botões do Friday Night Funkin agr não consigo jogar o jogo


AmebaFud1da:
### Issue Checklist

- [X] I have properly named the issue
- [ ] I have checked the issues/discussions pages to see if the issue has been previously reported

### What platform are you using?

Newgrounds (Web)

### If you are playing on a browser, which one are you using?

Microsoft Edge

### Version

0.4.1

### Context (Provide screenshots or videos of the crash happening)

eu estava alterando os botões do Friday Night Funkin mas sem querer selecionei o botão "enter" como um botão de saída do jogo, agora eu não consigo jogar ele (juro q não foi burrice minha)

### Steps to reproduce

eu quero resetar o FNF da minha conta

### Crash logs (can be found in the logs folder where Funkin.exe is)

i want to reset [my](url) game

Hundrec:
Keybinding softlocks have been reported multiple times now, so maybe we should add some restrictions on which buttons can be bound to which controls?

Issue #2780, opened 2024-06-13
Bug Report: Note Pressing Keybinds Override Eachother When Bound, Preventing Multiple Keybinds


Average-FNF-Modder:
## Describe the bug

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;When changing the 2 keybinds sets for game play, upon entering a song, one of the 2 sets will take over the other and the over written one will not be functional. Upon exiting said song and going back to settings, both keybind sets will read the same bindings. This makes it impossible to have to keybinds for each key as intended.

## To Reproduce

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Go into settings, change you note hit keybinds, enter a song, one set of keys will not work. You can exit the song and then also see the bug in settings with the keybinds.
## Expected behavior
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;It should let me have to keybinds per note. I only end up getting one binding. Very sad.

## Screenshots/Video
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;It's not really need here. You get the idea. Both keybind slots say the same thing, the end.

## Desktop
 - OS: It happens on the Windows 10 version.
 
 - Browser: I know it happens on Microsoft Edge.
 
 - Version: At LEAST up to 0.4.0, might also be in 0.4.1, didn't see it in the patch fixes.

Hundrec:
I tested both sets of arrow inputs in a song on 0.5.1 and it worked fine.
I think this is resolved!

Average-FNF-Modder:
> I tested both sets of arrow inputs in a song on 0.5.1 and it worked fine. I think this is resolved!

I just tested it myself once again, and the bug still occurred for me; maybe the bug varies by device, but it's definitely still present.

Hundrec:
Can you provide a video? I'm really curious to know more about this issue

Average-FNF-Modder:
Sure thing bro. 

https://github.com/user-attachments/assets/bd0da08c-09b4-4f96-a279-6f36531276f5

I try to go play Pico 2-handed, one side doesn't work, and I'm fucked. It's annoying for me especially cause I like to switch between one handed arrow keys only and 2 handed a s up right.

Hundrec:
So if you try to bind WASD to either the first or the second column, it resets into the arrow keys?

Average-FNF-Modder:
[...]

PR #4382, opened 2025-03-25
[BUGFIX] Prevent softlock caused by removing all keybinds for a UI control

VioletSnowLeopard:
#3205
Currently, the game allows you to remove all keybinds for any control. This includes controls important for navigating menus, like DOWN, UP, ACCEPT, and BACK. If you remove all keybinds for DOWN, then move up, you cannot reset DOWN. This PR prevents removing a keybind for these important controls if doing so would result it that key having no keybinds.

(The issue linked above talks about setting ENTER as a BACK key. I tested this, and ENTER still works as ACCEPT for the "Press ENTER to start the game" while bound to BACK)

I wouldn't recommend testing breaking your controls unless you know how to copy and replace your savefile.

NOTE: When you rebind a UI control, you cannot use that new keybind until you leave the controls menu. This is because the game does not load the new control scheme you made until after the controls menu is exited. This doesn't apply when you remove a keybind for a UI control, however. When you remove a keybind for a UI control, you won't be able to use it in the controls menu without having to leave it first.

This video shows me breaking the DOWN key. After I remove both keybinds for it and move up, I cannot go back to fix it.

https://github.com/user-attachments/assets/3ca4547f-2eb4-485c-860d-de984a4b2ef4

This video shows what my change does. I remove one keybind for DOWN, but the game won't let me remove the other one.

https://github.com/user-attachments/assets/b956f67b-12ac-4285-9198-d2fc472cff9c

This video shows me being able to use ENTER to start the game even though I keybound it to BACK

https://github.com/user-attachments/assets/9d479307-c32e-4817-9fa8-f635c557bfc5


AbnormalPoof (commented):
Would it be possible to have a popup display that says something like "You can't unbind this key!" to let the user know instead of silently not allowing the key to be unbound?

VioletSnowLeopard:
I added a popup to alert the user that they cannot remove the keybind:


https://github.com/user-attachments/assets/f72fee2a-2e6e-4775-9e3a-7b67931fede9

VioletSnowLeopard:
Just fixed a weird bug I found in the controls menu. When you leave the menu with one key bound to a UI control, then return, the control will appear duplicated. This duplicate doesn't actually exist, as you can see by me replacing one with "S". When I leave and go back, they both become "S".

https://github.com/user-attachments/assets/7470d522-8459-4e26-8304-1e93a20c69d5

Turns out this happens because the `getInputsFor` function inside the `getInput` function inside the `InputItem` class returns a list that duplicates the values three times sometimes:

Screenshot 2025-03-29 131724

I think this only happens for controls that had keybinds changed as this did not happen for the RESET control. This bug causes the length of a list that has one value to have three, which won't allow the program to figure out that there is only one distinct key bound to that control.

I added `list = list.distinct()` after this list is generated to ensure that the list only contains unique values to fix this issue.

VioletSnowLeopard:
This bug that I found and fixed may be related to #2780.

PR #4386, opened 2025-03-26
Chart editor - Disable edit buttons if no selection and stuff

Lasercar:
The edit buttons don't disable if there's no selection or whatever.

Also fixes the paste items command to _not_ use the current clipboard, but rather the clipboard at the time the command was created.

https://github.com/user-attachments/assets/3a1949e0-90f8-4ba8-bfa4-e3343f912dd6


Lasercar:
Rebased to allow for faster testing.

Lasercar:
Note, I'll have to do a follow-up for this the new commands once they're all in.

PR #4360, opened 2025-03-22
[ENHANCEMENT] Metadata pixel icons

AbnormalPoof:
This PR adds pixel icon data to the metadata, allowing for offsets adjustment and flipping.

This is different from the implementations seen in #2465 and #3042 in the sense that it uses the ***song*** metadata instead of the character metadata, Eric mentioned it here: https://github.com/FunkinCrew/Funkin/pull/3042#issuecomment-2254349610
This allows pixel icons to be configured on a song-by-song basis!

Example:
```json
{
  "version": "2.2.4",
  "songName": "Blazin'",
  "artist": "Kawai Sprite",
  "charter": "fabs + PhantomArcade",
  "playData": {
    "difficulties": ["easy", "normal", "hard"],
    "characters": {
      "player": "pico-blazin",
      "girlfriend": "nene",
      "opponent": "darnell-blazin"
    },
    "stage": "phillyBlazin",
    "noteStyle": "funkin",
    "ratings": { "easy": 3, "normal": 4, "hard": 5 },
    "album": "volume3",
    "previewStart": 0,
    "previewEnd": 15000,
    "freeplayIcon": {
      "id": "mom",
      "offsets": [0, 0],
      "flipX": true
    }
  },
  "generatedBy": "Chart Editor Import (FNF Legacy)",
  "timeFormat": "ms",
  "timeChanges": [
    { "t": 0, "b": 0, "bpm": 180, "n": 4, "d": 4, "bt": [4, 4, 4, 4] }
  ]
}
```
If `freeplayIcon` isn't in the song metadata, it defaults to the opponent's character ID, as it did before.

screenshot-2025-03-21-20-50-24

Let me know if there's any issues, I haven't been able to test this *as* much as my other PRs!

PR #4497, opened 2025-04-01
[ENHANCEMENT] Pixel Icon Metadata in Character Metadata

MadBear422:
* Had to make a new PR for this for clean up. Works with the latest version of the develop branch. Apologies for the inconvenience.

Freeplay and the chart editor now read from character metadata instead of hard-coding assigned character pixel icons. Existing character metadata files have updated with respective icons previously assigned in the source code . Additionally, the ability to flip the icons horizontally and adjusting the origin offset is now included via the metadata.

Situation:
For the modders who have multiple versions of some character and would like to use the same character icon, it was previously not possible simply due to the fact that prior to this change, the pixel icon is determined by the character's id name rather than a specified pixel ID. In this case, I have a characters labeled as "hugenate-opponent" and "hugenate-playable". I also have an icon file named "hugenate", and I want both of these characters to use the same icon. However, neither will work since they would have to be the same name as the freeplay icon file.
image
Now with this change, it corrects the issue and allows for me to individually assign the icon id to the players in their metadata. NOTE: If the id is left as null, the game will automatically set it to the same name as their character id.
image
After modifying it with this data, this is the new result!
image
New parameters such as flipX and origin are also included to adjust icons if they overlap the text:
image
Before:
image
After:
image

### DIFF
--- a/source/funkin/play/character/CharacterData.hx
+++ b/source/funkin/play/character/CharacterData.hx
@@ -12,6 +12,7 @@ import funkin.util.assets.DataAssets;
 import funkin.util.VersionUtil;
 import haxe.Json;
 import flixel.graphics.frames.FlxFrame;
+import openfl.utils.Assets;
 
 class CharacterDataParser
 {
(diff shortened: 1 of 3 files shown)


anysad (approved):
W

MadBear422:
Had to make some corrections and clean up the code

Lasercar:
Hmm, this conflicts with https://github.com/FunkinCrew/Funkin/pull/4360

Might want to consider merging them or the like.

MadBear422:
> Hmm, this conflicts with https://github.com/FunkinCrew/Funkin/pull/4360
> 
> 
> 
> Might want to consider merging them or the like.
> 
> 

I do think a good middle ground between the two implementations would be very good. I like the idea of having the icon being set per song since it is more modular. Meanwhile my implementation would be able address the discrepancies in menus such as the chart editor to display the respective pixel icon for each different character variation

MadBear422:
I'm not sure if it would be too much to integrate some form of prioritization as to how the icon metadata is handled. Maybe if there is no icon information provided within the song's metadata, the game then resorts to checking for it in the opponent character's metadata instead. If all else fails, resort to using the opponent character's ID or just have no icon listed if there is no pixel icon provided

anysad:
I'd say this would be a good middle ground with what you suggested

Lasercar:
> I'm not sure if it would be too much to integrate some form of prioritization as to how the icon metadata is handled. Maybe if there is no icon information provided within the song's metadata, the game then resorts to checking for it in the opponent character's metadata instead. If all else fails, resort to using the opponent character's ID or just have no icon be displayed if there is no asset provided.
> 
> What do you guys think? Would this be an ideal way to go?

Yeah, I like that.

Hundrec:
New source new merge conflicts!
Can you fix them up real quick?

Hundrec:
Thanks!

charlesisfeline:
merge conflicts spotted


again

Hundrec:
Uh oh! Please fix!

Hundrec:
Closing because this PR is stale.

Issue #4433, opened 2025-03-31
Bug Report: The debug mouse cursor stays visible if game loads unfocused


Hundrec:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.1

### Description (include any images, videos, errors, or crash logs)

Same as issue https://github.com/FunkinCrew/Funkin/issues/3712 but with a searchable title.

The cursor is usually hidden when the game starts loading, but letting it load unfocused keeps it visible.

![Image](https://github.com/user-attachments/assets/fca9fcc2-4223-4557-9ba3-ab5ab9d9844f)

### Steps to Reproduce

1. Open the game
2. Unfocus while the game loads
3. Reopen the game once Title Screen music starts playing

PR #3881, opened 2024-11-19
[BUGFIX] Mouse cursor is no longer visible outside of the ingame editors

sphis-Sinco:
## Does this PR close any issues? If so, link them below.
https://github.com/FunkinCrew/Funkin/issues/3879
https://github.com/FunkinCrew/Funkin/issues/3712

## Briefly describe the issue(s) fixed.
- The mouse cursor is no longer visible outside of ingame editors

## Include any relevant screenshots or videos.
Before:

screenshot-2024-11-19-13-54-38

After:

https://github.com/user-attachments/assets/625af0c6-3d0a-4b27-b94a-6128d321eef6

PR #4520, opened 2025-04-03
[BUGFIX] Hide debug cursor when Title Screen starts

Hundrec:
## Linked Issue
Fixes https://github.com/FunkinCrew/Funkin/issues/4433

## Changes
Hides the debug cursor when the Title Screen starts up using the same line of code as #3881.

## Notes
In 0.6.2, loading into the Title Screen while the game is unfocused keeps the debug cursor visible after clicking back into focus.
#3881 intended to fix this issue, but only ended up removing the debug cursor flickering in the top left for a split second.
Both that change and this one are necessary to make the debug cursor never show up during launch, focused or unfocused.

### DIFF
--- a/source/funkin/ui/title/TitleState.hx
+++ b/source/funkin/ui/title/TitleState.hx
@@ -231,6 +231,8 @@ class TitleState extends MusicBeatState
 
     Conductor.instance.update();
 
+    funkin.input.Cursor.hide();
+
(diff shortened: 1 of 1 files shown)


NotHyper-474 (approved):
Tested and does fix the issue described :rocket:

cyn0x8 (approved):

PR #4544, opened 2025-04-04
Chart Editor - Invalid chart difficulty playstate crash fix

Lasercar:
Fixes a crash when getDifficulty returns null in the loading state and it tries to get the stage of said null difficulty, when loading the playstate.

This crash can be replicated in 0.5.3 by loading a song template with an erect difficulty, switching to it, then loading a FNFC with an erect difficulty (which it doesn't load properly) and then try playtesting, hence the title as such.


Lasercar:
Hmm, this PR has basically been invalidated by the new changes

PR #4589, opened 2025-04-05
Fix memory overflow in crash log

ACrazyTown:
No

Previously when generating the crash log, `cpp.vm.Gc.memInfo()` was used which returns a 32bit integer that overflows after ~2.14GB. Ez fix by just using `memInfo64` instead

Before, memory overflows after ~2.14GB
Screenshot 2025-04-06 001451 2

After, no overflow! Memory can grow as much as it wants! Hooray!
Screenshot 2025-04-06 001741 2

### DIFF
--- a/source/funkin/util/MemoryUtil.hx
+++ b/source/funkin/util/MemoryUtil.hx
@@ -17,10 +17,10 @@ class MemoryUtil
   {
     #if cpp
     var result:String = 'HXCPP-Immix:';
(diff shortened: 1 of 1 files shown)


cyn0x8 (approved):

ACrazyTown:
This seems to have disappeared from develop?

Hundrec:
Oh whoops, we might've messed something up while force pushing develop
I'll let Eric know

Hundrec:
Yep, we done messed up!
Please open a new identical PR based on `develop` and we'll merge it!

PR #4621, opened 2025-04-06
Replace Deprecated State Switching

VioletSnowLeopard:
No
The way of switching states with Flixel has changed, resulting in the old way being deprecated. Many instances of state switching in the code were changed to the new way in [this commit](https://github.com/FunkinCrew/Funkin/commit/e5fb1de4ba74412e7f3e6673a7223e8dbe373d37#diff-d735da8a36bf5a753af53c828f8a2213a4cd5bf5bad648e262c3391f9fb8f442). There are still two instances of using the old way to switch states in the code. This PR replaces these last two instances with the new way.

Screenshot 2025-04-06 075823

### DIFF
--- a/source/funkin/ui/debug/DebugMenuSubState.hx
+++ b/source/funkin/ui/debug/DebugMenuSubState.hx
@@ -112,7 +112,7 @@ class DebugMenuSubState extends MusicBeatSubState
 
   function openCharSelect()
   {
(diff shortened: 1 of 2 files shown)


VioletSnowLeopard:
This is messy, but I don't need to redo this. I believe that all deprecated function calls have been replaced except for this:

Screenshot 2025-09-04 132139

And this is in a separate library.

Hundrec:
Hooray!

PR #4219, opened 2025-02-24
[CLEANUP + ENHANCEMENT] rework polymod buildImports

lunacynne:
- organizes default imports/aliases/blacklists
- eases restriction on `sys` package, instead only blocks its possibly malicious classes like `sys.io.*` and `sys.FileSystem`
- added a few more common imports to default imports
  - `flixel.FlxG`
  - `funkin.Conductor`
  - `funkin.modding.module.ModuleHandler`
  - `funkin.play.PlayState`

> [!warning]
> existing prs that change `buildImports` will conflict

PR #4634, opened 2025-04-07
[MODDING] Ease restrictions on `funkin.api.*`

AbnormalPoof:
There's no reason for `DiscordClient` to be blacklisted. This PR changes it so `DiscordClient` is unblacklisted, while keeping everything else in `funkin.api.*` blacklisted.

### DIFF
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -237,6 +237,9 @@ class PolymodHandler
 
   static function buildImports():Void
   {
+    // A list of classes from blacklisted packages that are allowed to be imported.
+    var allowedClasses:Array<String> = ['funkin.api.discord.DiscordClient'];
+
     // Add default imports for common classes.
     Polymod.addDefaultImport(funkin.Assets);
     Polymod.addDefaultImport(funkin.Paths);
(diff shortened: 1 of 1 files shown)


cyn0x8 (approved):

cyn0x8:
implemented in #4219

Hundrec:
Oh, I guess this would be redundant then

AbnormalPoof:
> I blacklisted the entire `funkin.api.*` package as a standard, since many functions that are (and will be) in this package will be important or secure, and should not be accessible to mods.
> 
> A better solution would to be to implement a `DiscordClientSandboxed` and allow access to that by adding an alias in Polymod.

Here's what I came up with:

```haxe
// A list of classes from blacklisted packages that are allowed to be imported.
var allowedClasses:Array<String> = ['funkin.api.discord.DiscordClient'];

// `funkin.api.*`
// Contains functions which allow for cheating and such.
for (cls in ClassMacro.listClassesInPackage('funkin.api'))
{
  if (cls == null) continue;
  var className:String = Type.getClassName(cls);
  if (allowedClasses.contains(className)) continue;
  Polymod.blacklistImport(className);
}
```
This allows `DiscordClient` to be used directly, removing the need for an extra wrapper class in the source code!

AbnormalPoof:
closing in favor of https://github.com/FunkinCrew/Funkin/pull/5040

PR #4671, opened 2025-04-09
[ENHANCEMENT + BUGFIX] Better dialogue wrapping, fix empty text hang

xenkap:
This PR adds a new class `FunkinTypeText` which extends FlxTypeText.
It does generally nothing but override insertBreakLines() with a better implementation, ensuring words no longer start on one line and immediately move to the next. Also fixes a bug where empty text will cause the dialogue to never 'finish'.

(Feels like it'd be useful outside Funkin too, but... blehh. If you want to push my changes to [upstream FlxTypeText](https://github.com/HaxeFlixel/flixel-addons/blob/master/flixel/addons/text/FlxTypeText.hx), go right ahead!)

https://github.com/FunkinCrew/Funkin/issues/4907
Text in dialogue don't properly account for text wrapping, starting words on one line and wrapping them to the next. This PR fixes that as explained above.
BEFORE: (Incorrect wrapping, freezes on empty text. More issues not shown here)

https://github.com/user-attachments/assets/deb5b92e-ad0c-4e22-9406-4b1c2fa694f0

AFTER:

https://github.com/user-attachments/assets/8be51202-3531-494f-b6ce-e986357ce097


EliteMasterEric:
I think I know what this PR is referring to but a video comparison would be nice.

xenkap:
> I think I know what this PR is referring to but a video comparison would be nice.

Ah! Sorry, got busy and didn't get around to it.
Funnily enough, while editing the dialogue to record some footage, I encountered another bug: if the dialogue string is empty, then the text softlocks and becomes un-advance-able (doesn't trigger the onComplete() callback!!).

Feels like a PR to upstream there, but for now I'll just include a fix in the class I made =w=

Hundrec:
Interesting! Can you open a quick issue about this?

xenkap:
> Interesting! Can you open a quick issue about this?

Sure! Although it'll just be resolved by this PR when I get to it lol

xenkap:
So I started having trouble recreating what exactly my issue was?
And then I started playing around with segmented lines (lines that advance bit by bit) 
Then I discovered weird inconsistencies on how the text wrapped depending on *where* you put a space, due to FlxTypeText's own implementation. And now I'm gonna fix that too by removing the original implementation :boom:

xenkap:
All ready! :yippe: x3c
Added the video comparisons as promised :+1:

EliteMasterEric (approved):

Lasercar:
I wouldn't merge this one yet, it still needs a rebase to remove those unnecessary commits.

Though maybe only the relevant commits were picked, idk.

AbnormalPoof:
> I wouldn't merge this one yet, it still needs a rebase to remove those unnecessary commits.
> 
> Though maybe only the relevant commits were picked, idk.

It’s gonna be merged internally anyway. We can dummy-merge the public PR after.

ninjamuffin99:
i may have mentioned this with eric but i think this PR should be in upstream flixel, rather than funkin specific code!

Issue #4678, opened 2025-04-10
Enhancement: Add volume slider in settings


amyspark-ng:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

There should be a percentage option in the preferences menu, it would be nice for people who do not know that you can change it with - and +

Starexify:
Don't get me wrong, but this feels really unnecessary and if you open the Controls it basically shows you that you can set the Volume Up and Down so 🤷

amyspark-ng:
i mean, someone might still find it useful? i don't know

Starexify:
- i think this is a forgotten duplicate of: #4667

PR #4680, opened 2025-04-10
[ENHANCEMENT] Add volume slider in preferences

amyspark-ng:

closes https://github.com/FunkinCrew/Funkin/issues/4678

I actually can't seem to get it working, i didn't find anything online about something like `FlxG.sound.volume` and it seems like i can't do Save.instance.set_volume() because it's a private field, am i cooked chat?

### DIFF
--- a/source/funkin/save/Save.hx
+++ b/source/funkin/save/Save.hx
@@ -1205,7 +1205,7 @@ class Save
 }
 
 /**
- * An anonymous structure containingg all the user's save data.
+ * An anonymous structure containing all the user's save data.
  * Isn't stored with JSON, stored with some sort of Haxe built-in serialization?
  */
 typedef RawSaveData =
--- a/source/funkin/ui/options/PreferencesMenu.hx
+++ b/source/funkin/ui/options/PreferencesMenu.hx
@@ -14,6 +14,7 @@ import funkin.ui.TextMenuList.TextMenuItem;
 import funkin.ui.options.items.CheckboxPreferenceItem;
 import funkin.ui.options.items.NumberPreferenceItem;
 import funkin.ui.options.items.EnumPreferenceItem;
+import funkin.save.Save;
 
 class PreferencesMenu extends Page<OptionsState.OptionsMenuPageName>
 {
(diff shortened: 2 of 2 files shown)


KoloInDaCrib:
> i can't do Save.instance.set_volume() because it's a private field

`@:privateAccess` on the table:

amyspark-ng:
What does this exactly do? I'm not too familiar with Haxe/Funkin source

KoloInDaCrib:
> What does this exactly do? I'm not too familiar with Haxe/Funkin source

It basically bypasses whether or not a field is private

For example, if a class `Test` were to have a private bool field `_test` you would be able to access it like

```
@:privateAccess
{
 Test._test = true;
}
```

amyspark-ng:
Ah, this would work in keeping track of the volume

Even tho i just remembered, i don't think the save volume was used anywhere? Couldn't really find other places it was mentioned using Save.instance.volume

How does funkin store the value from the soundtray? guessing FlxG.sound.changeVolume() changes FlxG.sound.volume

Lasercar:
https://github.com/FunkinCrew/Funkin/blob/fcb47ff4fca1624ac6e97e890841178770130578/source/funkin/ui/options/FunkinSoundTray.hx#L100-L108

Hopefully that helps!

amyspark-ng:
will probably use FlxG.sound.volume, since that's what the sound tray uses

amyspark-ng:
was trying this
    createPrefItemPercentage('Master volume', 'The general volume of the game.', function(value:Int):Void {
      FlxG.sound.volume = value;
    }, FlxG.sound.volume);
    
And doesn't seem to work due to FlxG.sound.volume being floats rather than Int, which is what prefItemPercentage asks for, i tried casting but i get unsafe casting i think? how should i do this?

anysad:
might be a bit of a hack, but when getting the value you could probably do
```
FlxG.sound.volume = Std.parseFloat(Std.string(value))

**Comment by KoloInDaCrib** [comment]
> was trying this createPrefItemPercentage('Master volume', 'The general volume of the game.', function(value:Int):Void { FlxG.sound.volume = value; }, FlxG.sound.volume);
> 
> And doesn't seem to work due to FlxG.sound.volume being floats rather than Int, which is what prefItemPercentage asks for, i tried casting but i get unsafe casting i think? how should i do this?

Considering the sound volume is a value between 0 and 1 (0 and 100%) I recommend doing something like this

```
createPrefItemPercentage('Master volume', 'The general volume of the game.', function(value:Int):Void 
{ 
FlxG.sound.volume = value / 100; 
}, Std.int(FlxG.sound.volume * 100));
```

```Std.int()``` rounds a float so that it becomes an int (with the standard rules like anything below x.5 will become x and anything above would be x+1), and I'm multiplying by 100 to get the percentage value

amyspark-ng:
Turns out it's a lot more complex, the SoundTray and the setting would run different, it's just too much work and not worth it at all, sad

PR #4699, opened 2025-04-12
[CHORE] Add the v0.6 stages to the list of base game stages

KoloInDaCrib:
Adds `schoolErect`, `schoolPico`, `schoolEvilErect` and `tankmanBattlefieldErect` to the list of base game stages provided by `StageRegistry` since i guess the devs forgot, lol

At some point in the future these kinds of functions should be replaced with a macro imo.....


anysad:
well that was quick

EliteMasterEric:
This is good, though in the future I might want to look into putting together a macro that scans the data folders and hardcodes the names of all the data files.

EliteMasterEric:
We ended up replacing this with #4707 sowwy 🥺

PR #4707, opened 2025-04-12
[FEATURE] Use Macro For Base Game Assets

lemz1:
## REFERENCE
https://github.com/FunkinCrew/Funkin/pull/4699#issuecomment-2798901739

## DESCRIPTION
this pr adds a macro that retrieves the base game stages

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "assets"]
 	path = assets
-	url = https://github.com/FunkinCrew/Funkin-assets-secret
+	url = https://github.com/FunkinCrew/Funkin.assets
 [submodule "art"]
 	path = art
-	url = https://github.com/FunkinCrew/Funkin-art-secret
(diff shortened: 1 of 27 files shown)


KoloInDaCrib:
I think this should be moved into its own class because a lot of registries use hardcoded base game assets, this is pretty rad nonetheless!

lemz1:
> I think this should be moved into its own class because a lot of registries use hardcoded base game assets, this is pretty rad nonetheless!

what should be moved into its own class?

EliteMasterEric:
I have a neat idea, I'll use this PR as a foundation for it so you get credit

EliteMasterEric:
I made a branch which merges this with #3694 and then utilizes a macro to retrieve the path of the data files from the constructor call, and creates a `listBaseGameEntryIds()` function for you.

Notably I overrode this with a hard-coded one for `LevelRegistry` because the game needs that one to be in order for use by the Story Mode menu but otherwise it should work pretty well.

PR #4721, opened 2025-04-14
[BUGFIX] Ensure Song starts at Correct time when Restarting Songs with BPM Changes

### LINKED ISSUE #4434: Bug Report: Restarting Eggnog Erect at the end of the song causes the song to start 2 seconds late
### Issue Checklist

### Platform

### Browser

None

### Version

0.6.1

When you restart a song from the pause menu, it starts at the same time it normally should. I saw that this was not the case for Eggnog Erect, however.

Here is a video of when the song should start after a restart through the pause menu:

https://github.com/user-attachments/assets/cac56c9d-4695-47a7-97c1-56e8009cf5ba

If you restart after the last hold note of the song, the song ends up starting 2 seconds after the countdown ends:

https://github.com/user-attachments/assets/10e1704e-5c48-4dc4-8d1e-d66b5b7e97fc

### Steps to Reproduce

1. Play Eggnog Erect
2. After the last hold note of the song, restart the song through the pause menu.
3. The song will now start 2 seconds after the countdown ends.

VioletSnowLeopard:
Fixes #4434
Restarting a song that has BPM changes when the song has a different BPM than its starting BPM causes the song to start at the incorrect time. 
When the BPM is lower than the starting BPM, the song will start later than it should. This is noticeable in Eggnog Erect where restarting after the BPM change at the end causes the song to start two seconds after the countdown ends.

When the BPM is higher than the starting BPM, the song will start before the countdown ends. This is somewhat noticeable in Monster where restarting during its highest BPM causes the instrumental to start slightly before the countdown ends and the camera event focusing on Boyfriend happens before the countdown ends.

This happens because the time the song starts at in the countdown is based on `beatLengthMs`, but the BPM hasn't changed yet because the song position hasn't updated yet. This causes the song position to be set based on the current BPM rather than the starting BPM.

I fixed this issue by checking if `startingSong` is `true`, and if it is, the `bpm` variable is set to the BPM of the song at the `startTimestamp`.
### Songs with BPM changes starting at the incorrect time
Eggnog Erect:

https://github.com/user-attachments/assets/3e56c6cb-ade7-40e5-a64c-e8619069492d

Monster:

https://github.com/user-attachments/assets/bbd47c10-7d54-47e8-943f-18dff746bf05

### Starting at the correct time after changes

Eggnog Erect:

https://github.com/user-attachments/assets/03b5b9ad-8ad1-45a3-b3e4-28274aae76b1

Monster:

https://github.com/user-attachments/assets/4e17f59a-7b84-4903-a829-9c1a8b14ce39

### DIFF
--- a/source/funkin/Conductor.hx
+++ b/source/funkin/Conductor.hx
@@ -5,6 +5,7 @@ import flixel.util.FlxSignal;
 import flixel.math.FlxMath;
 import funkin.data.song.SongData.SongTimeChange;
 import funkin.data.song.SongDataUtils;
+import funkin.play.PlayState;
 import funkin.save.Save;
 import funkin.util.TimerUtil.SongSequence;
 import haxe.Timer;
@@ -112,6 +113,17 @@ class Conductor
     if (bpmOverride != null) return bpmOverride;
 
     if (currentTimeChange == null) return Constants.DEFAULT_BPM;
+
+    @:privateAccess
+    if (PlayState.instance != null && PlayState.instance.startingSong)
+    {
+      for (i in 0...timeChanges.length)
+      {
+        if (PlayState.instance.startTimestamp >= timeChanges[i].timeStamp) currentTimeChange = timeChanges[i];
+
+        if (PlayState.instance.startTimestamp < timeChanges[i].timeStamp) break;
+      }
+    }
 
     return currentTimeChange.bpm;
   }
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -557,11 +557,9 @@ class PolymodHandler
     Polymod.clearScripts();
 
     // Forcibly reload Polymod so it finds any new files.
+    // This will also register all scripts.
     // TODO: Replace this with loadEnabledMods().
     funkin.modding.PolymodHandler.loadAllMods();
-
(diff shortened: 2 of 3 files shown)


VioletSnowLeopard:
> added medium and pending triage and removed medium and pending triage 💀

The bot tends to remove tags it just added a lot and I don't know why.

AbnormalPoof (commented):
Can't you just have `beatLengthMs` use the starting BPM directly? (Unless there's a reason to keep them separate)

AbnormalPoof (approved):
Thanks for clarifying!

VioletSnowLeopard:
Perhaps I could have used the `startingSong` boolean from `PlayState` to have the getter function for `beatLengthMs` return a different value if `startingSong` is true, but `startingSong` is currently a private variable.

AbnormalPoof:
> Perhaps I could have used the `startingSong` boolean from `PlayState` to have the getter function for `beatLengthMs` return a different value if `startingSong` is true, but `startingSong` is currently a private variable.

`@:privateAccess` in the corner:

VioletSnowLeopard:
Actually, I found I can use the public `isInCountdown` variable which is set to true before the countdown uses `beatLengthMs`, so I updated it to do this instead.

TechnikTil (approved):
works

PR #4369, opened 2025-03-23
Add null safety for `FileUtil`

NotHyper-474:
A contribution towards #4303.
## Briefly describe the issue(s) fixed.
Adds null safety to the `FileUtil` class. Tested it with HTML5, C++ and HL targets to ensure there's no compile errors.

Issue #4303, opened 2025-03-12
Chore: Enable null safety on many classes


EliteMasterEric:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

[Null safety](https://haxe.org/manual/cr-null-safety.html) is a feature of Haxe which reduces the risk of null object reference errors at compile time. Adding `@:nullSafety` to a class mandates that, for all the expressions in that class, any variables which can be assigned null explicitly use the `Null<T>` type. It also mandates that functions which can return null specify `Null<T>` as the return value, and mandates that any attempts to access properties of that type, or pass that type to another function, handle the edge case that the variable is null (typically using null-safe field accesses or the null coalescing operator).

In short, it requires that any uses of `null` be properly handled.

This could be enabled project-wide via a flag, but this causes many issues, as there are many places in other libraries which do not adhere to null safety; for example, `FlxG.sound.music` can be null but does not use the type `Null<FlxSound>`, and thus attempting to assign `FlxG.sound.music = null` requires turning off null safety for that line.

However, null safety can easily and safely be enabled on a per-class basis, and is currently enabled on over 80 classes in the repository.

Any PRs which apply the `@:nullSafety` annotation, and then resolve any resulting compilation errors, and add proper checks in places where previously nullable values were not properly handled, would help in preventing future crashes, and thus are greatly appreciated.

Lasercar:
Do you think there could be list created for all the classes that currently have full null safety, partial null safety (null safety is turned off for a few functions) and no null safety? That, and which classes have received full or partial null safety internally so we can ignore them? (It also goes without saying that new classes don't need to be listed)

It'd help speed things up.

Hundrec:
Compiling a list sounds like it would take a long time.
I'll leave this code search here since it would be the next best thing:

https://github.com/search?q=repo%3AFunkinCrew%2FFunkin%20%40%3AnullSafety&type=code

Lasercar:
@EliteMasterEric You should consider pinning this so it doesn't get lost in the coming storm of issues.

EliteMasterEric:
[...]

PR #4732, opened 2025-04-16
chore: Add null safety to various classes in `funkin.util`

NotHyper-474:
A small contribution towards #4303.

### Description
This adds null safety for nearly every class in the module, the ones that weren't included were:

* `WatchPlugin`, `VersionUtil`, `ReflectUtil`: Doesn't compile on HTML5, an error with `Type.getClass` that I couldn't figure out a way around.
* `FileUtil` - Already done in #4369.
* `ScreenshotPlugin` - Already done internally.
* `CrashHandler` and `CLIUtil` - Already has null safety.


Hundrec (changes requested):
Looks like these changes break all song events (camera movement, animations, etc.)

image

Hundrec (approved):
Balance has been restored.

Hundrec (approved):
Yep, still compiles!

Issue #4737, opened 2025-04-16
Bug Report: Major Vulnerabilities involving opening and writing files, as well as opening URLs within FlxG, WindowUtil, and FIleUtil.


Burgerballs:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Newgrounds (Web/HTML5)

### Browser

None

### Version

Any version between 0.3.0 and when the bug should be fixed

### Description (include any images, videos, errors, or crash logs)

You can convert files including executables to Base64 and back using strings and haxe.crypto.Base64. Allowing actors to write bytes to a paths is a big nono! openFolder also opens files, such as executables, also extremely bad!

WindowUtil also has a duplicate openFolder function. openUrl can also open files if you attach file:// beforehand, FlxG has the same issue with openUrl.

AbnormalPoof:
oh

DustedDustin:
uh

anysad:
uh oh

Keoiki:
`openFolder` could be changed so that any and all traces of extensions (`.exe`, `.png`, just anything that fits `.*`) are removed, since it just plainly runs `Sys.command`. In that event would `writeBytesToPath` (not File) continue to cause any risks?

nebulazorua:
> `openFolder` could be changed so that any and all traces of extensions (`.exe`, `.png`, just anything that fits `.*`) are removed, since it just plainly runs `Sys.command`. In that event would `writeBytesToPath` (not File) continue to cause any risks?

Allowing mods to use Sys.command at all is a very very very bad idea

Burgerballs:
keep in mind there is a lot more than JUST fileutil which is the issue

ninjamuffin99:
if people use fnf mod for malware i think thats awesome

Burgerballs:
thank you mr ninja muffin

anysad:
alright, who will make the malware now

AbnormalPoof:
> alright, who will make the malware now

```haxe
import funkin.modding.module.Module;

/**
 * EXPLODES YOUR FUCKING COMPUTER
 */
class Malware extends Module
{  
  public function new()
  {
    super('Malware');

    explodeComputer();
  }

 function explodeComputer():Void
 {
[...]

PR #4738, opened 2025-04-16
[VULNERABILITY] FileUtil restriction in Polymod

anysad:
Fixes https://github.com/FunkinCrew/Funkin/issues/4737, thanks a lot @Burgerballs 
* As described in the issue, giving the access to `FileUtil` can give bad actors the possibility to run malicious code on users devices via mods. This is why this tiny change just outright removes mods the ability to use `FileUtil`.
* Whenever a mod tries to use the `FileUtil` class, it will just give them an error that the class can't be found, thus not being imported.

### DIFF
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -284,6 +284,11 @@ class PolymodHandler
     // System.load() can load malicious DLLs
     Polymod.blacklistImport('lime.system.System');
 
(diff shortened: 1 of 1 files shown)


Keoiki:
But.. FileUtil can have legitimate uses in mods, this is total overkill.

anysad:
Good point, https://github.com/FunkinCrew/Funkin/pull/3032 does a better job, it was indeed an overkill

PR #4749, opened 2025-04-17
chore: Add null-safety to `Leaderboards` and `Medals`

NotHyper-474:
A tiny contribution towards #4303.

Seems like it wasn't implemented in these classes because of a compile error when retrieving values from the internal maps.
Turns out there's no need to iterate through the lists, you can just `@:privateAccess` them, and since both `Medal` and `Leaderboard` are just abstracts over integers the internal map type is compatible with them, which allows you to just return them directly.

> [!NOTE]
> ~`_map` itself and the rest of the Newgrounds API isn't null-safe so this might still cause a Null Object Reference error if either the medals or leader boards aren't fully loaded.~ I've made it fallback to an empty map just in case so this might not be a problem anymore.

### DIFF
--- a/source/funkin/api/newgrounds/Leaderboards.hx
+++ b/source/funkin/api/newgrounds/Leaderboards.hx
@@ -9,6 +9,7 @@ import io.newgrounds.objects.User;
 import io.newgrounds.objects.events.Outcome;
 import io.newgrounds.utils.ScoreBoardList;
 
+@:nullSafety
 class Leaderboards
 {
   public static function listLeaderboardData():Map<Leaderboard, LeaderboardData>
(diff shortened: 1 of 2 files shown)


Hundrec (approved):
Builds fine!

Hundrec (approved):
Compiles fine in 0.6.3

Hundrec (changes requested):
Looks like this needs an update!

NotHyper-474:
```diff
- return @:privateAccess list._map.copy();
+ return @:privateAccess list._map?.copy() ?? [];
```
Also decided to make them fallback to an empty map like done above just in case, so the problem I described in the Note shouldn't be an issue anymore.

Hundrec (approved):
It built for me yay

PR #4318, opened 2025-03-15
Fix strumline offset when downscroll

Lasercar:
## Does this PR close any issues? If so, link them below.
Fixes #4317

PR #4789, opened 2025-04-23
[CHORE] Add null-safety to `PlayState` and `LoadingState`

NotHyper-474:
A quite large contribution towards #4303.

~This still needs a bit of polishing so I'm making this a draft for the moment.~ I think I'm satisfied with how it turned out.

I've also included `LoadingState` here as it made sense for me since it's responsible for switching the game to the `PlayState`.

`PlayState` might've been the most troublesome class to add null-safety to since it has so many fields and whatnot. It isn't complete null-safety since ~some fields still need to be initialized later in `create` and also the fact~ the null-safety checker has many oddities. But some notes regarding it:

* I've made it so `currentSong` **cannot** be null at all, so there's gonna be errors thrown in the case something may set it null.

* I added a new optional parameter to Strumline's constructor, `scrollSpeed`. This was done because `resetScrollSpeed` would always get the current scroll value from the PlayState, which doesn't have its `instance` set until it gets called on `create` which would cause the scroll speed to always be 1.
I could just call `resetScrollSpeed` in `initStrumline` but I didn't like that.

* ~`playerStrumline`, `opponentStrumline`, and other fields I've disabled null-safety on would require moving their initialization to `new`, I felt like the constructor was big enough already with the changes I made to ensure the cameras and the follow point are not nullable but if that is not a problem then I can move them there and remove their init functions.
Unfortunately I cannot just move their `init` functions from `create` to `new` as the compiler requires explicitly setting the field in the constructor.~

### DIFF
--- a/source/funkin/play/GameOverSubState.hx
+++ b/source/funkin/play/GameOverSubState.hx
@@ -162,7 +162,7 @@ class GameOverSubState extends MusicBeatSubState
     if ((parentPlayState?.isMinimalMode ?? true)) {}
     else
     {
-      boyfriend = parentPlayState?.currentStage.getBoyfriend(true);
+      boyfriend = parentPlayState?.currentStage?.getBoyfriend(true);
       if (boyfriend != null)
       {
         boyfriend.canPlayOtherAnims = true;
@@ -193,10 +193,9 @@ class GameOverSubState extends MusicBeatSubState
     });
   }
 
-  @:nullSafety(Off)
   function setCameraTarget():Void
   {
-    if ((parentPlayState?.isMinimalMode ?? true) || boyfriend == null) return;
+    if (parentPlayState == null || parentPlayState.isMinimalMode || boyfriend == null) return;
 
     // Assign a camera follow point to the boyfriend's position.
     cameraFollowPoint = new FlxObject(parentPlayState.cameraFollowPoint.x, parentPlayState.cameraFollowPoint.y, 1, 1);
@@ -207,6 +206,7 @@ class GameOverSubState extends MusicBeatSubState
     cameraFollowPoint.y += offsets[1];
     add(cameraFollowPoint);
 
+    @:nullSafety(Off)
     FlxG.camera.target = null;
     FlxG.camera.follow(cameraFollowPoint, LOCKON, Constants.DEFAULT_CAMERA_FOLLOW_RATE / 2);
     targetCameraZoom = (parentPlayState?.currentStage?.camZoom ?? 1.0) * boyfriend.getDeathCameraZoom();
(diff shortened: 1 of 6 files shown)


Hundrec (changes requested):
Amazing work!
It currently gives me this error when compiling, just another hole to patch up!

image

NotHyper-474:
> Amazing work!
> It currently gives me this error when compiling, just another hole to patch up!
> 
> image
> 

mfw `NO_FEATURE_DISCORD_RPC`

Hundrec (approved):
This is huge.

Compiles fine on Windows!

source/funkin/play/GameOverSubState.hx
```diff
@@ -299,9 +304,9 @@ class GameOverSubState extends MusicBeatSubState
           PlayStatePlaylist.reset();
         }
 
-        var playerCharacterId = PlayerRegistry.instance.getCharacterOwnerId(PlayState.instance.currentChart.characters.player);
+        var playerCharacterId = PlayerRegistry.instance?.getCharacterOwnerId(PlayState.instance?.currentChart?.characters?.player);
```
VioletSnowLeopard, line 307: As I just did null-safety for `PlayerRegistry` in #4798, I know that the public `instance` for it will never be null, so the `?` after `PlayerRegistry.instance` isn't necessary. I don't think it hurts though, so you don't have to remove it.

NotHyper-474: Ah, good to know. I think I may have added that check there because I mistook it for `PlayState.instance`

AbnormalPoof:
Will test this later but it makes me really happy to see PlayState get null safety!

VioletSnowLeopard:
Merge conflicts? In null-safe PlayState?

Hundrec (approved):
Compiles on 0.6.3!

Hundrec (approved):
Yep, still compiles!

NotHyper-474:
guuhh thanks to https://github.com/FunkinCrew/Funkin/pull/4318 I had to move some strumline shit to the constructor so I'm re-requesting review just to make sure nothing broke

Hundrec (approved):
Builds on 0.6.4!

Hundrec:
This is `huge`.

source/funkin/play/PlayState.hx
```diff
@@ -565,8 +565,8 @@ class PlayState extends MusicBeatSubState
 
   function get_currentStageId():String
   {
-    var stage = currentChart?.stage;
-    return (stage == null || stage == '') ? Constants.DEFAULT_STAGE : stage;
+    var stage = currentChart?.stage ?? '';
```
AbnormalPoof, line 568: should this have the type of `Null<String>`?

PR #4798, opened 2025-04-25
[CHORE] Add null-safety to registries and 5 other classes `funkin.data`

VioletSnowLeopard:
A contribution towards https://github.com/FunkinCrew/Funkin/issues/4303.

This PR adds null-safety to all registries in `funkin.data` (`SongRegistry` already has null-safety). These include:
- `BaseRegistry`
- `ConversationRegistry`
- `DialogueBoxRegistry`
- `SpeakerRegistry`
- `SongEventRegistry`
- `AlbumRegistry`
- `PlayerRegistry`
- `FreeplayStyleRegistry`
- `NoteStyleRegistry`
- `StageRegistry`
- `LevelRegistry`

I also had to fix `Song.hx` because that file has null-safety, but was dependent on `BaseRegistry` which wasn't null-safe.

Aside from the registries, I included null-safety for some other classes in `funkin.data` too. These include:
- `DataError`
- `DataParse`
- `DataWrite`
- `AnimationData`
- `SongDataUtils`


Lasercar:
Just something to note, this'll have merge conflicts with 0.6.3, which to my understanding is being released... [now](https://github.com/FunkinCrew/Funkin/tree/0.6.3-release)

VioletSnowLeopard:
fuck

Hundrec (approved):
Builds fine on 0.6.3!

Hundrec (approved):
Builds on 0.6.4!

Hundrec (changes requested):
Doesn't compile right now:

ERROR source/funkin/util/macro/ClassMacro.hx:64: characters 18-95

64 | return macro funkin.util.macro.CompiledClassList.getTyped(vrequest, {targetClassExpr});
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
| Null safety: Cannot cast nullable value to not nullable type.

VioletSnowLeopard:
Fixed the error, although it might not be the best fix. While the error message points to that location in the code, trying to modify that class doesn't seem to fix the issue. This may be due to the function being a macro function. I was able to fix it by turning off null safety for the one line in the code that uses the `listSubClassesOf` function, which is in `SongEventRegistry`. I could only think of turning of null safety here because modifying where the function is actually defined in the code doesn't seem to do anything.

NotHyper-474:
The issue is that now `CompiledClassList#get` returns a nullable value, and `getTyped` is an inline function that just returns `cast CompiledClassList.get` and returns a non-nullable type, so there's a type mismatch between those two methods which for some reason wasn't catched by the compiler before (probably because of the unsafe cast used).

I think a proper fix would be to modify `CompiledClassList` so that `get` returns a non-nullable type
```diff
-  public static function get(request:String):Null<List<Class<Dynamic>>>
+  public static function get(request:String):List<Class<Dynamic>>
   {
    if (!initialized) init();

    if (!classLists.exists(request))
    {
      trace('[WARNING] Class list $request not properly generated. Please debug the build macro.');
      classLists.set(request, new List()); // Make the error only appear once.
    }

+    // A faulty request sets the value to an empty list above so this will never be null
+    @:nullSafety(Off)
     return classLists.get(request);
   }
```

VioletSnowLeopard:
> The issue is that now `CompiledClassList#get` returns a nullable value, and `getTyped` is an inline function that just returns `cast CompiledClassList.get` and returns a non-nullable type, so there's a type mismatch between those two methods which for some reason wasn't catched by the compiler before (probably because of the unsafe cast used).
> 
> I think a proper fix would be to modify `CompiledClassList` so that `get` returns a non-nullable type
> 
> ```diff
> -  public static function get(request:String):Null<List<Class<Dynamic>>>
> +  public static function get(request:String):List<Class<Dynamic>>
>    {
>     if (!initialized) init();
> 
>     if (!classLists.exists(request))
>     {
>       trace('[WARNING] Class list $request not properly generated. Please debug the build macro.');
>       classLists.set(request, new List()); // Make the error only appear once.
>     }
> 
> +    // A faulty request sets the value to an empty list above so this will never be null
> +    @:nullSafety(Off)
>      return classLists.get(request);
>    }
> ```

Thank you, this indeed did fix the issue.

PR #4832, opened 2025-04-27
[BUGFIX] Fix the song preview restarting when changing difficulties within the same variation

### LINKED ISSUE #4821: Bug Report: Freeplay song preview not restarting no longer works

VioletSnowLeopard:
Fixes #4821
The recent update changed the calls to `generateSongList` after difficulty changes to set `onlyIfChanged` to `false`. This would make them do their switching animation, but it would also call `changeSelection` which resets the song preview.

This PR fixes this issue by adding a parameter to `changeSelection` with a default value of `true`. If this value is `false`, `playCurSongPreview` won't get called. I set the call to `changeSelection` in `generateSongList` to pass in a value of `false` over the default `true`. Because `changeSelection` is called in several different places, this change doesn't cause the song preview to not restart when it should.

I also had to fix `changeDiff` to ensure its call to `playCurSongPreview` worked. When switching to erect difficulty on a song that doesn't have it, it will change the selected song to the closest song which does have it. I added in an update to `characterVariations` for the new selected song so `currentVariation` would be updated correctly.

https://github.com/user-attachments/assets/02b3b9be-ed5b-4006-a781-dbf0cbd79fc3

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -787,7 +787,7 @@ class FreeplayState extends MusicBeatSubState
     FlxG.console.registerFunction('changeSelection', changeSelection);
 
     rememberSelection();
-    changeSelection();
+    changeSelection(0, false);
     refreshCapsuleDisplays();
   }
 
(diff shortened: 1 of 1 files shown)


Lasercar:
Do you think you could add one extra change to this? Play the random song if the song still has the 'NEW!' label on it, so that it doesn't spoil what the song sounds like until it's actually played.

DustedDustin:
I know it isn't related to this pr but someone should definitely open one related to random playing on new songs. That way the random song would get more use and fix the issue with spoiling some of the surprise

Hundrec:
Yes, someone will probably pick that one up

PR #4852, opened 2025-04-27
Remove Changelog from game files

Hundrec:
## Change
Adds `CHANGELOG.md` to .gitignore

Eric mentioned potentially excluding the Changelog from the game files since it keeps getting revised after updates are released.
Perhaps we could find a better way to direct players to the latest Changelog.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ shitAudio/
 .build_time
 .swp
 NewgroundsCredentials.hx
+CHANGELOG.md
 
 # Exclude JS stuff
 node_modules/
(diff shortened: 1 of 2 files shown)


NotHyper-474 (commented):
I don't think that would be enough. Adding it to .gitignore just prevents it from being tracked by Git.

This line would need to be removed to not include it with the game.
https://github.com/FunkinCrew/Funkin/blob/db3e012cb0dae07a3cecf18b0c06ac68460369d0/project.hxp#L782

Hundrec:
True...
Thanks for catching this!

amyspark-ng:
i like changelog in assets :( but i guess it's fine not having it there i guess

PR #4870, opened 2025-04-28
[BUGFIX] Fix the song preview restarting when changing difficulties within the same variation

VioletSnowLeopard:
Fixes https://github.com/FunkinCrew/Funkin/issues/4821

*Reopening of #4832*
The recent update changed the calls to `generateSongList` after difficulty changes to set `onlyIfChanged` to false. This would make them do their switching animation, but it would also call `changeSelection` which resets the song preview.

This PR fixes this issue by adding a parameter to `changeSelection` with a default value of true. If this value is false, `playCurSongPreview` won't get called. I set the call to `changeSelection` in `generateSongList` to pass in a value of false over the default true. Because `changeSelection` is called in several different places, this change doesn't cause the song preview to not restart when it should.

I also had to fix `changeDiff` to ensure its call to `playCurSongPreview` worked. When switching to erect difficulty on a song that doesn't have it, it will change the selected song to the closest song which does have it. I added in an update to `characterVariations` for the new selected song so `currentVariation` would be updated correctly.

https://github.com/user-attachments/assets/77e642ff-837d-4804-a721-300a4a6af384

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -787,7 +787,7 @@ class FreeplayState extends MusicBeatSubState
     FlxG.console.registerFunction('changeSelection', changeSelection);
 
     rememberSelection();
-    changeSelection();
+    changeSelection(0, false);
     refreshCapsuleDisplays();
   }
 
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
Yeah, it works!

Hundrec:
Good change, though it seems a similar change was made internally.
I'll leave this open until we know for sure

Hundrec:
Song previews function properly now, so I will close this.
Thank you for contributing!

PR #4873, opened 2025-04-28
Remove leftover from merge-conflict in CHANGELOG

NotHyper-474:
I don't think this one requires much explanation.


EliteMasterEric:
I was wondering why VSCode kept complaining, thanks 👍

Issue #4858, opened 2025-04-28
Bug Report: choosing nightmare in freeplay on "random" and re-entering freeplay bug


DemiSans:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.3

### Description (include any images, videos, errors, or crash logs)

If you choose nightmare and select "random", exit from freeplay and enter again, it will display normal difficulty, but every song will display rank from nightmare, and pressing enter on "random" will choose nightmare difficulty song

https://github.com/user-attachments/assets/9fbb8a2a-f442-4f70-84e8-aba5329a956a

### Steps to Reproduce

explained above

VioletSnowLeopard:
This issue also occurs with any difficulty aside from Normal selected. It can happen when Easy and Hard are selected too. As soon as you switch off of Random, the difficulty displayed in the top left updates:

https://github.com/user-attachments/assets/3d5223c8-3170-4577-a5f2-b4e4c30268cc

Lasercar:
That's very weird, because the random capsule should no longer be able to do that....

Basically, the way it worked before is any time the random capsule was selected, it'd change the difficulty to the Constants.Default_difficulty, aka, normal, but the line that does that was removed.

PR #4885, opened 2025-04-29
[BUGFIX] Fix two issues with the Random Capsule in Freeplay

### LINKED ISSUE #4856: Bug Report: Freeplay capsule bugs
### Issue Checklist

### Platform

### Browser

None

### Version

0.6.3

If you choose "random" capsule, and switch to erect or nightmare difficulty, it will switch to the "Bopeebo".
Also if you choose nightmare and select "random", exit from freeplay and enter again, it will display normal difficulty, but every song will display rank from nightmare, and pressing enter on "random" will choose nightmare difficulty song

https://github.com/user-attachments/assets/d082ddf9-c3bf-48a3-9faf-34c4a1db1663

### Steps to Reproduce

Ugh, I explained above

VioletSnowLeopard:
Fixes #4856 and #4858.

I put the fixes for these together because the fix for the second issue doesn't work without the fix for the first. (Well it technically does, but it introduces another bug)

When switching to Erect difficulty while the Random capsule is selected, the selected song would change to Bopeebo Erect. This happens because the new functionality added for switching variations when a song doesn't have it doesn't check if the song is null (aka random). This causes it to execute the code to switch the selected song to the closest one that does have that variation, Bopeebo. I fixed this issue by making the game check if the song is null too before attempting to select a different song.

The second issue is when you leave Freeplay with the Random capsule selected, the difficulty displayed in the top left would always show Normal while the songs and ranks would be from the difficulty you left the menu on previously. This bug happens because there isn't a call to `changeDiff` in `changeSelection` when the selected song is null (aka random). The difficulty text displayed in the top left is updated inside the `changeDiff` function, so this needs to be called when the selected song is random for it to update properly.

Video showing both bugs fixed:

https://github.com/user-attachments/assets/97ff6d11-cb3c-46f4-832d-0ed0c2419ca8

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -1717,14 +1717,14 @@ class FreeplayState extends MusicBeatSubState
   {
     touchTimer = 0;
     var previousVariation:String = currentVariation;
+    var daSong:Null<FreeplaySongData> = grpCapsules.members[curSelected].freeplayData;
 
     // Available variations for current character. We get this since bf is usually `default` variation, and `pico` is `pico`
     // but sometimes pico can be the default variation (weekend 1 songs), and bf can be `bf` variation (darnell)
-    var characterVariations:Array<String> = grpCapsules.members[curSelected].freeplayData?.data.getVariationsByCharacter(currentCharacter) ?? Constants.DEFAULT_VARIATION_LIST;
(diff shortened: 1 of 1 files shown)


Lasercar:
VioletSnowLeopard's Freeplay UI fixes

Kade-github:
thats my bad, this change looks good tho. Nice job :D

PR #4279, opened 2025-03-09
Chart editor - None character

Lasercar:
## Does this PR close any issues? If so, link them below.
Closes #4273
## Briefly describe the issue(s) fixed.
Adds a none/null option to the character dropdown. 

Also fixes a few null reference errors when one or both of the characters in the playstate are null.
## Include any relevant screenshots or videos.

https://github.com/user-attachments/assets/f843f36e-ba83-4d4d-822c-dedf41516925

PR #4065, opened 2025-01-30
FreeplayScore seventh/million number not being cleared fix

Lasercar:
## Does this PR close any issues? If so, link them below.
Fixes #2318 
## Briefly describe the issue(s) fixed.
The last digit of the FreeplayScore doesn't get cleared before it sets a new number.
## Include any relevant screenshots or videos.
https://github.com/user-attachments/assets/bd2cd058-8754-47cb-a888-264ed1ba06da

I uhhhhh think I understand how the code for the FreeplayScore now? I spent way too much time staring at the code trying to figure out how it works (and where I could set the intendedscore to 1M to test the fix), like a few hours. Ah well, I fixed it in the end, didn't I?

Ok, let's try that again, this time without accidentally starting the new branch from another fix.

PR #4930, opened 2025-05-02
<RECOVERY> Restore two of Lasercar's commits

Hundrec:
Original PRs: #4065 and #4279

We lost two of Lasercar's commits while merging a ton of stuff, so we're bringing them back.


Lasercar:
Future: < RECOVERY > `<RECOVERY> Restore two of Lasercar's commits`

Hundrec:
True...

PR #4934, opened 2025-05-02
<RECOVERY> Restore a commit from Hundrec

VioletSnowLeopard:
Original PR: #4777

This commit was lost somewhere, so the changelog doesn't specify the shortcut for the "flip horizontally" functionality in the Animation Editor. This re-adds the specification for the shortcut key.


Hundrec:
Wow, nice catch! Thanks for making sure our history is intact

Hundrec (approved):

PR #4952, opened 2025-05-03
[BUGFIX?] Comment-accurate offsets

KoloInDaCrib:
As of right now, `combinedOffset` in Conductor is calculated using `instrumentalOffset + audioVisualOffset + formatOffset`. (view [here](https://github.com/FunkinCrew/Funkin/blob/0b976b0d711af80f6a8ed39e6f97fadd554805e9/source/funkin/Conductor.hx#L282))
However, later in the code, there exists a comment stating that it should be `instrumentalOffset + formatOffset + audioVisualOffset`. (view [here](https://github.com/FunkinCrew/Funkin/blob/0b976b0d711af80f6a8ed39e6f97fadd554805e9/source/funkin/Conductor.hx#L403))

This PR accomodates the combined offsets to the said comment.
I'm not sure if this completely fixes the input offsets or not, but I always found it weird that conductor updates based on the input offsets.

### DIFF
--- a/source/funkin/Conductor.hx
+++ b/source/funkin/Conductor.hx
@@ -279,7 +279,7 @@ class Conductor
 
   function get_combinedOffset():Float
   {
-    return instrumentalOffset + audioVisualOffset + inputOffset;
+    return instrumentalOffset + audioVisualOffset + formatOffset;
   }
 
   /**


Hundrec:
This might be huge, we need @JackXson-Real to test this

VioletSnowLeopard:
According to the comment above the variable,`formatOffset` is "an offset tied to the file format of the audio file being played". This value is set in `PlayState.hx` and will always be zero unless `Constants.EXT_SOUND == 'mp3'`. 

But this might actually fix the issue because the conductor shouldn't need to update based on the input offset. The input offset is already being accounted for in the hit windows for the notes in the `processNotes` function in `PlayState.hx`.

AbnormalPoof:
This PR is GLOOBY! (Already done internally)

PR #4962, opened 2025-05-03
[ENHANCEMENT] Freeplay - Ability to reset saved song score

### LINKED ISSUE #2708: Enhancement: Reset song/week scores
Now that I'm working towards improving my scores I've noticed that this feature is missing. Currently the only ways to change your scores that I'm aware of are directly manipulating the json file and deleting your save data, neither of which are ideal. Being able to set your score for a week in story mode or a song in freeplay back to 0 in-game would be a great convenience.

The main reason for a quick reset would be to keep up with new versions of charts. Say that you release a chart that has a triple, and a player gets a gold P rank on it. But then the charter decides that triples are lame, so in the next version they make it a double. Now the song has one less note, and the maximum possible score is one note smaller. That means the player who got a gold P rank on the first version will likely never be able to beat their saved score on the second. A reset button would let that player go for a legit perfect score on the new chart without any hassle.

It's also nice future-proofing for modded weeks that can have volatile developments. For example, I played a mod whose release version had a week with five songs in it. In an update, they moved two of those songs to their own week and left the original week with three. That left the score I got on the old version several hundred thousand points higher than the new maximum. This isn't an issue I expect the official game to deal with, but the feature would be a nice safeguard against big changes like that.

There's also just some value in getting a perfect score and then saying "Screw it, I'm gonna do it all over again just to prove that I can".

I don't know if this is already in the cards (I couldn't find any open matching issue) but I thought I'd put it forward.

Lasercar:
Closes #2708 (partially, this is only for freeplay only)
Adds a `CapsuleOptionsMenu` for deleting the save data of a song.

Is opened by holding delete (currently hardcoded) when entering a song, to actually delete the data the second option must be selected. 

Also makes a few slight modifications to the capsule options menu, such as renaming all the variables to better match the class name and adding a variable for setting the label header in the new function.

https://github.com/user-attachments/assets/9a41c0dc-4fd4-4285-8ef4-20484934677c

### DIFF
--- a/source/funkin/ui/freeplay/CapsuleOptionsMenu.hx
+++ b/source/funkin/ui/freeplay/CapsuleOptionsMenu.hx
@@ -20,14 +20,16 @@ class CapsuleOptionsMenu extends FlxSpriteGroup
 
   var queueDestroy:Bool = false;
 
-  var instrumentalIds:Array<String> = [''];
-  var currentInstrumentalIndex:Int = 0;
+  var options:Array<String> = [''];
+  var currentOptionIndex:Int = 0;
 
-  var currentInstrumental:FlxText;
+  var currentOption:FlxText;
 
   var busy:Bool = false;
-  var leftArrow:InstrumentalSelector;
-  var rightArrow:InstrumentalSelector;
+
+  var leftArrow:OptionSelector;
+
+  var rightArrow:OptionSelector;
 
   public function setBusy(b:Bool):Void
   {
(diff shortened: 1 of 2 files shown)


TechnikTil:
merge conflicts jigsaw_tweaking

NotHyper-474:
Closing this due to our 90-day policy for stale PRs that need revision, as stated in the Contributing Guide.

Please let us know if you'd want to bring this back so that we reopen it.

Lasercar:
meh, if someone wants this I'll update it.

PR #4965, opened 2025-05-03
[ENHANCEMENT] Story Menu Score Commas

JVNpixels:
No, this Pull Request does not close any open issues.

This Pull Request adds commas to the Story Menu score, the reason is to make the score easier to read.

image
image

### DIFF
--- a/source/funkin/ui/story/StoryMenuState.hx
+++ b/source/funkin/ui/story/StoryMenuState.hx
@@ -8,6 +8,7 @@ import flixel.tweens.FlxEase;
 import flixel.tweens.FlxTween;
 import flixel.util.FlxColor;
 import flixel.util.FlxTimer;
+import flixel.util.FlxStringUtil;
 import funkin.audio.FunkinSound;
 import funkin.data.story.level.LevelRegistry;
 import funkin.data.song.SongRegistry;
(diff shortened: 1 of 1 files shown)


Lasercar:
But the planned story menu rework...

JVNpixels:
> But the planned story menu rework...

why would this have to do with that, they can just port all of this over

JVNpixels:
> But the planned story menu rework...

also its 3 lines so its okay

source/funkin/ui/story/StoryMenuState.hx
```diff
@@ -314,7 +315,8 @@ class StoryMenuState extends MusicBeatState
 
     highScoreLerp = Std.int(MathUtil.smoothLerp(highScoreLerp, highScore, elapsed, 0.25));
 
-    scoreText.text = 'LEVEL SCORE: ${Math.round(highScoreLerp)}';
+    var commaSeparated:Bool = true;
+    scoreText.text = 'LEVEL SCORE: ${FlxStringUtil.formatMoney(highScoreLerp, false, commaSeparated)}';
```
AbnormalPoof, line 319: This can be shortened.
```suggestion
    scoreText.text = 'LEVEL SCORE: ${FlxStringUtil.formatMoney(highScoreLerp, false, true)}';
```
You're defining a boolean that's always true so you might as well use `true` directly.

Hundrec: From what I remember, this code is written this way in `PlayState.hx`

Maybe the devs want this value to be changeable?

JVNpixels: > From what I remember, this code is written this way in `PlayState.hx`
> 
> Maybe the devs want this value to be changeable?

That is the reason I made a variable, otherwise I would have  did AbnormalPoof's method.

Hundrec (approved):
Looks good to me!

JVNpixels:
@AbnormalPoof, do you agree with me and Hundrec or should something be changed here?

source/funkin/ui/story/StoryMenuState.hx
JVNpixels: It is also one line anyway.

PR #4972, opened 2025-05-04
[ENHANCEMENT] Extra compile instruction for linux

Kade-github:
For people who are building on other distros, like Arch Linux, or Fedora.

Just show them the documentation for HXVlc

### DIFF
--- a/docs/COMPILING.md
+++ b/docs/COMPILING.md
@@ -24,6 +24,7 @@
     - Linux: [`lime setup linux` Documentation](https://lime.openfl.org/docs/advanced-setup/linux/)
         - One of Funkin's dependencies uses libVLC, which requires you to install some development packages to be able to compile.
           Command for Ubuntu/Debian based systems: `sudo apt install libvlc-dev libvlccore-dev libvlccore9`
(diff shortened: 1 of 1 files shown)


NotHyper-474 (approved):

NotHyper-474:
Actually, maybe this section could be rewritten to something like 

"Funkin uses hxvlc, which requires you to install some packages to be able to compile. Please refer to [hxvlc's documentation](https://github.com/MAJigsaw77/hxvlc?tab=readme-ov-file#dependencies)"

Kade-github:
> Actually, maybe this section could be rewritten to something like
> 
> "Funkin uses hxvlc, which requires you to install some packages to be able to compile. Please refer to [hxvlc's documentation](https://github.com/MAJigsaw77/hxvlc?tab=readme-ov-file&rgh-link-date=2025-05-04T00%3A30%3A59Z#dependencies)"

Maybe, but then it'll require for the docs to remove the previous install argument, which I think is helpful for people since it's the most used distro.

EliteMasterEric:
LGTM

PR #5033, opened 2025-05-06
Change the way strums work with Note Hits. Reorganize classes in `funkin.play.notes` and Helper Functions.

PurSnake:
## Changes
- A lot of classes in`funkin.play.notes`  got reorganized.
- Strumline now have _targetCharacter_ (to determine what character uses this strum) and _downScroll_ (self explanatory) variables.
- Some classes in `funkin.play.notes` now have __parentStrum_.
- Rewrites how **BaseCharacter** handle **NoteScriptEvent**.
- NoteSprite now have _targetCharacter_ , thats override Strumline _targetCharacter_ , and it can be set manually via scripts.
It should work with NoteKinds, i guess

- BaseCharacter got new Helper Functions (Especially _setShader_, because its now used for **AnimateAtlasCharacter** and can be overriden for complex characters, like Nene)

Also a lot of scripts got cleaned up. (No more PlayState.instance.currentStage code in STAGE scripts, duh -_-).

Its hard to write EVERY change in this pr, so check code and write your opinions and what to change.

(the hell is that .github files)

### DIFF
--- a/source/funkin/modding/events/ScriptEvent.hx
+++ b/source/funkin/modding/events/ScriptEvent.hx
@@ -10,6 +10,7 @@ import funkin.play.cutscene.dialogue.Conversation;
 import funkin.play.Countdown.CountdownStep;
 import funkin.play.notes.NoteDirection;
 import openfl.events.KeyboardEvent;
+import funkin.play.character.BaseCharacter;
 
 /**
  * This is a base class for all events that are issued to scripted classes.
@@ -97,6 +98,11 @@ class NoteScriptEvent extends ScriptEvent
   public var note(default, null):NoteSprite;
 
   /**
+   * The character, parent strumline attached to.
+   */
+  public var targerCharacter(default, default):BaseCharacter;
+
+  /**
    * The combo count as it is with this event.
    * Will be (combo) on miss events and (combo + 1) on hit events (the stored combo count won't update if the event is cancelled).
    */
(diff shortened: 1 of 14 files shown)


AbnormalPoof:
nearly 2000 changes...
image

PurSnake:
I wouldn't say it's a breaking change, it still works fine with the previous scripts. (Tested with Itch.io 0.6.4 release scripts)

Starexify:
Can this PR also add better support for mods that add extra or less keys/directions to the notes ?
For example: KEY_COUNT ,SPLASHES_CAP shouldn't be final, the hold covers shouldn't be limited to 4, and maybe aswell be a variable from the strumline instead, like the KEY_COUNT.

Only if possible aswell for these functions from SongData which are forced to use 4, which mods can't modify most classes that use them:
image
image

Lasercar:
> I wouldn't say it's a breaking change, it still works fine with the previous scripts. (Tested with Itch.io 0.6.4 release scripts)

It's still a breaking change due to the large amount of changes. I wouldn't be surprised if a bug was found among all those changes.

PurSnake:
> It's still a breaking change due to the large amount of changes. I wouldn't be surprised if a bug was found among all those > changes.

Ehh, true

PurSnake:
oh god
conflicts

PurSnake:
TODO: REDO

PR #5050, opened 2025-05-09
[CHORE] "Auto Pause" -> "Pause on Unfocus" in Module.hx comments

JackXson-Real:
## Description
Comments in Module.hx incorrectly refer to `Pause on Unfocus` as `Auto Pause` even after the option was renamed, so this PR fixes that.

### DIFF
--- a/source/funkin/modding/module/Module.hx
+++ b/source/funkin/modding/module/Module.hx
@@ -176,13 +176,13 @@ class Module implements IPlayStateScriptedClass implements IStateChangingScripte
 
   /**
    * Called when the game regains focus.
-   * This does not get called if "Auto Pause" is disabled.
+   * This does not get called if "Pause on Unfocus" is disabled.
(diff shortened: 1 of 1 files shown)


JackXson-Real:
Thanks @Hundrec for letting me know about this terrifying error in the game's comments.

Hundrec:
What if I told you there are far more of these in other files...

charlesisfeline:
> What if I told you there are far more of these in other files...

i checked theres nothing of note

JackXson-Real:
> > What if I told you there are far more of these in other files...
> 
> i checked theres nothing of note

I think he’s referring to the fact that the actual variable thing or whatever is called “autoPause” but I can’t change those since it would break everything.

Hundrec (approved):
True...

AbnormalPoof (approved):
sure

PR #5051, opened 2025-05-09
[ENHANCEMENT] [MODDING] More state instances

AbnormalPoof:
Basically so you can do something like `ResultState.instance` instead of `FlxG.state.subState` for cleaner syntax and shit.


NotHyper-474:
image

Oops, watch out for the null-safe classes

AbnormalPoof:
> image
> 
> 
> 
> Oops, watch out for the null-safe classes!

image

AbnormalPoof:
i fixed it im an excellent coder yippeeeeeeeeeeeeeeeeeeee

Hundrec:
Friday Funkin'

EliteMasterEric:
Could we possibly switch some of these to use `ISingleton` maybe?

AbnormalPoof:
> Could we possibly switch some of these to use `ISingleton` maybe?

`SingletonMacro` makes a new instance of the class if it doesn't exist which doesn't really work well for states. I tried using it before for stuff like `PlayState` and it didn't work well.

Issue #5049, opened 2025-05-08
Bug Report: Song skips (resyncs) forward at least once during any song


Hundrec:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.4

### Description (include any images, videos, errors, or crash logs)

Just when you thought `resyncVocals` was defeated, it comes back for more.

Much like #4332, the song skips a bit forward. However, this happens somewhat randomly.

No special actions are required to trigger the bug:
* It happens in both regular PlayState and Chart Editor playtest
* It happens on any stage
* It happens with or without Botplay
* It happens on any FPS
* It isn't triggered by sustains (#4761) and can happen within the same song as a sustain lag spike
* It can happen multiple times on longer songs.
* There's some space between each occurrence. Something might be building up or drifting.

The only thing I haven't tested is using a device with better hardware.

## Video
I've tested this with many songs, stages, settings, etc. 
Here's DadBattle on Hard (no sustain bug). 
Skip to **0:41** to see the resync happen in game and in the console.

https://github.com/user-attachments/assets/fd042009-839d-4cd3-8ddf-ac14f3492102

  

This is a very minor issue compared to everything we've seen before, but I want `resyncVocals` to be fixed for good.

### Steps to Reproduce

1. Enter a song
2. Wait until the song subtly resyncs forward

VioletSnowLeopard:
I think I just got this at the very start of the song when I was playing Ugh Erect.

Unfortunately, my instant replay keybind was set to a dumb key so I couldn't get a video of it. I have now fixed my instant replay keybind so I will be prepared if I ever see it happen again.

Lasercar:
`I hate this function`

Hundrec:
`True...`

Hundrec:
So I disabled mid-song resyncing, but kept the traces. The song is drifting!

Maybe this problem will be a bit harder to solve than the previous bugs.

https://github.com/user-attachments/assets/417eb04c-e4a7-454e-a196-2238f6ae16bb

Hundrec:
So what could be happening here?
[...]

PR #5052, opened 2025-05-09
[BUGFIX] Hardware audio delay

mikolka9144:
## Linked Issues
https://github.com/FunkinCrew/Funkin/issues/5049

## Description
This pull request request adds support for hardware playback delay while synchronizing vocals, which was very apparent when playing using wireless headphones.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -930,8 +930,8 @@ class PlayState extends MusicBeatSubState
         Conductor.instance.formatOffset = 0.0;
       }
 
-      Conductor.instance.update(Conductor.instance.songPosition + elapsed * 1000, false); // Normal conductor update.
-
+      // Conductor.instance.update(Conductor.instance.songPosition + elapsed * 1000, false); // Normal conductor update.
+      Conductor.instance.update(FlxG.sound?.music?.time ?? 0.0);
       // If, after updating the conductor, the instrumental has finished, end the song immediately.
       // This helps prevent a major bug where the level suddenly loops back to the start or middle.
       if (Conductor.instance.songPosition >= (FlxG.sound.music.endTime ?? FlxG.sound.music.length))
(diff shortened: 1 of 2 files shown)


Hundrec:
Interesting! I'll note that I only use speakers, so I will be testing #5049 on this PR using speakers as well

NotHyper-474:
This is kind of what Funkin's fork of Lime does already when retrieving audio time?

image

Hundrec (changes requested):
Unfortunately, I was able to encounter #5049 while using this PR
The bug still behaves the same way :(

VioletSnowLeopard:
I wouldn't say that this issue is common or apparent when playing with wireless headphones. I recently switched to using wireless headphones and I rarely encounter the bug.

Hundrec:
@VioletSnowLeopard Have you tried watching the console while running songs on Botplay? It's possible that you might not notice every instance of the bug due to how subtle it is

VioletSnowLeopard:
> @VioletSnowLeopard Have you tried watching the console while running songs on Botplay? It's possible that you might not notice every instance of the bug due to how subtle it is

I can try to do that. I just noticed another incident of it happening though, and it is very subtle, but the vocals do skip forward a little:

https://github.com/user-attachments/assets/c1a9a6c1-a720-45bb-9064-e6e4a51a0b91

Although considering you don't use wireless headphones, I don't think they have any impact on encountering the bug.

Hundrec:
That's strange, I've never seen that behavior
Is this with headphones on the base game or headphones on this PR?

VioletSnowLeopard:
That was on base game

ACrazyTown:
As NotHyper-474 said, Funkin's Lime fork already compensates for latency when fetching the audio time:

https://github.com/FunkinCrew/lime/blob/535f3c98803c657f9f6b87497ed9b48692c07e31/src/lime/_internal/backend/native/NativeAudioSource.hx#L410-L416

What you're doing is effectively undoing that and reverting to the old method that returned the time as it is in the audio buffer and not the time when the sound actually plays through the speakers

mikolka9144:
> As NotHyper-474 said, Funkin's Lime fork already compensates for latency when fetching the audio time:
> 
> https://github.com/FunkinCrew/lime/blob/535f3c98803c657f9f6b87497ed9b48692c07e31/src/lime/_internal/backend/native/NativeAudioSource.hx#L410-L416
> 
> What you're doing is effectively undoing that and reverting to the old method that returned the time as it is in the audio buffer and not the time when the sound actually plays through the speakers

You talk about the recent commit?

mikolka9144:
The problem I discovered was while lime took the hardware delay into account, the conductor didn't. 
If the hardware delay is higher than 100ms, the game will attempt to resync the audio to the conductor time, which will cause lime to resync with the hardware delay, which will cause game to attempt to resync audio again...

ACrazyTown:
I'm a bit unfamiliar with Funkin's code so forgive me if I'm missing something glaringly obvious, but if the Conductor gets updated with the audio time (which is latency compensated) how is the Conductor not also compensating for latency?

mikolka9144:
> I'm a bit unfamiliar with Funkin's code so forgive me if I'm missing something glaringly obvious, but if the Conductor gets updated with the audio time (which is latency compensated) how is the Conductor not also compensating for latency?

Conductor assumes that the audio's timer starts passing as soon as the ".play()" is called, which is not the case (it's delayed by the hardware delay you mention here)

VioletSnowLeopard (changes requested):
It seems that playing the audio for the song in the chart editor, then playtesting the song causes a crash:

https://github.com/user-attachments/assets/35962994-f5f0-4f1c-87e6-7113148819ef

(For the record, I didn't have the "start at current time" setting enabled)

NotHyper-474:
This PR is not gonna work anymore since `getSourcedvSOFT` was removed https://github.com/FunkinCrew/lime/commit/79abb97b256760dc30c5ed55367d313d669292c6

AbnormalPoof (changes requested):
Please resolve the merge conflicts on this PR within 90 days.

Hundrec:
This PR is stale.

PR #5054, opened 2025-05-10
[ENHANCEMENT] Support for more than 9 Characters in `CharSelectSubState`

KoloInDaCrib:
## Description
So far, `CharSelectSubState` only supports 9 characters, with cursor and icon positions being hardcoded to fit that. This PR changes that, since with mods, you could end up with more than 9 playable characters, which would make it so that some of them would never be accessed. 
To fit with this change, I made it so that `CharSelectSubState` doesn't use two variables for determining the currently selected slot, but rather only one. The game calculates the current row and column on the run.

## Screenshots/Videos

https://github.com/user-attachments/assets/5f6e9b9e-1bf9-4318-97cf-70d8b1b9e672

### DIFF
--- a/source/funkin/ui/charSelect/CharSelectSubState.hx
+++ b/source/funkin/ui/charSelect/CharSelectSubState.hx
@@ -46,11 +46,17 @@ class CharSelectSubState extends MusicBeatSubState
   /**
    * The default index for the cursor.
    */
-  final DEFAULT_CURSOR_INDEX:Int = 4;
+  static final DEFAULT_CURSOR_INDEX:Int = 4;
 
+  /**
+   * The amount of slots per page.
+   */
+  static final SLOTS_PER_PAGE:Int = 9;
+
+  final SLOTS_PER_ROW:Int = Math.floor(Math.sqrt(SLOTS_PER_PAGE));
+  var totalSlots:Int = SLOTS_PER_PAGE;
   var cursors:CharSelectCursors;
-  var cursorX:Int = 0;
-  var cursorY:Int = 0;
+  var curSelected:Int = DEFAULT_CURSOR_INDEX;
   var cursorFactor:Float = 110;
   var cursorOffsetX:Float = -16;
   var cursorOffsetY:Float = -48;
@@ -136,6 +142,8 @@ class CharSelectSubState extends MusicBeatSubState
       trace('Placing player ${playerId} at position ${targetPosition}');
       availableChars.set(targetPosition, playerId);
 
+      totalSlots = Std.int(Math.max(targetPosition, totalSlots));
+
       CharSelectAtlasHandler.loadAtlas('charSelect/${playerId}Chill');
 
       var gfPath:Null<String> = playerData.gf?.assetPath;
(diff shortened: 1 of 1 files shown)


AbnormalPoof:
IMG_0475

AbnormalPoof (approved):

amyspark-ng (approved):

amyspark-ng:
crazy cool

i wonder, does funkin chooose what position a character is it? or does the character choose it, what happens if a modded funker is on the same position of another

AbnormalPoof:
> crazy cool
> 
> 
> 
> i wonder, does funkin chooose what position a character is it? or does the character choose it, what happens if a modded funker is on the same position of another

The playable character chooses the position. If 2 characters overlap, then one gets pushed to the next available slot.

KoloInDaCrib:
> i wonder, does funkin chooose what position a character is it? or does the character choose it, what happens if a modded funker is on the same position of another

The game first prioritizes what slot a player is supposed to be in according to their .json file, and if it that slot is already taken, it shifts their position by one until it finds a free slot.
This system works st this time however if you have 8 mods that each add a custom player, one of them will never get accessed, depending on the initial slots and the mod order.
And in this PR that crisis is averted by adding a row of slots to the bottom if there is a character that occupies that row. So, unlimited playable characters, yay :D

PSI3x:
This is really cool, pretty much exactly how I envisioned it to be
If this gets added I think it could make sense if instead of having a fixed size, it just adds the amount of rows needed when considering how many characters are currently modded in

KoloInDaCrib:
> This is really cool, pretty much exactly how I imagined it to be If this gets added I think it could make sense if instead of having a fixed size, it just adds the amount of rows needed when considering how many characters are currently modded in

It already does that.

rodney528 (approved):

EliteMasterEric (approved):
New code looks good (I like that it's more readable than the original), we'd need to get approval from Dave before merging but it should be good since it would ever only show up with mods installed.

KoloInDaCrib:
> New code looks good (I like that it's more readable than the original), we'd need to get approval from Dave before merging but it should be good since it would ever only show up with mods installed.

I was told privately that Dave would prefer if there was no scrolling (following the standard fighting game rules) and I got a suggestion to make it be something like pages for characters - aside from that I'd have to implement some way for mobile controls to work for this system and resolve merge conflicts before this gets merged.

Will be drafting the PR in the meantime 'till I'm done with everything mentioned.

KoloInDaCrib:
Recently had the time to implement requested changes and fix merge conflicts - should be good to go now!

https://github.com/user-attachments/assets/429eb48e-1d43-4ab5-911d-8264b727db72

The cursor is shown because i didn't feel like compiling to mobile so I tested the mobile controls on desktop, with TouchUtil making the cursor visible.

rodney528 (approved):

Raltyro:
it got conflicts :(

rodney528 (approved):

AbnormalPoof:
Merged this into my little internal rework :)
No guarantee it'll be in the next patch but I'll merge this anyway so that it counts as merged.

immalloy:
love u brochcho for this

PR #5062, opened 2025-05-11
Disable V-Sync option on web builds

NotHyper-474:
## Linked Issues
...

## Description
Disables the option on web builds since it's pretty much ineffective.

## Screenshots/Videos
image


charlesisfeline (changes requested):
keep the conditional for the fps option too btw

NotHyper-474:
> keep the conditional for the fps option too btw

The FPS option was not affected.

charlesisfeline:
> > keep the conditional for the fps option too btw
> 
> The FPS option was not affected.

oh right i didnt check twice aight

EliteMasterEric (approved):

PR #5067, opened 2025-05-11
[ENHANCEMENT] `FlxColorUtil` overhaul

AbnormalPoof:
This PR overhauls `FlxColorUtil`. Adding a bunch of new functions and adding an import alias for scripts so they can import `flixel.util.FlxColor` along with `funkin.util.FlxColorUtil`.


AbnormalPoof:
No longer needed since Polymod has native support for `FlxColor` now.

PR #5070, opened 2025-05-12
Rename "Input Offsets" menu to "Offset Adjustment" to be more accurate

JackXson-Real:
## Description
Renames the `INPUT OFFSETS` menu to `OFFSET ADJUSTMENT` because the menu allows you to change both the input offset and visual offset, so you could argue that the current name is a bit misleading. Also this is how it is named in other rhythm games (or at least Osu! cause thats the only one I looked at)

## Screenshots/Videos
screenshot-2025-05-11-23-27-45


JackXson-Real:
Fixed merge conflicts

Kade-github:
We're going to be renaming the menu to 'Lag Adjustment', and that'll be the final; from Dave.

Issue #4277, opened 2025-03-09
Bug Report: Chart Editor - Hold note selection boxes move with the screen


Lasercar:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Compiled from GitHub Source Code

### Version

Develop branch

### Description (include any images, videos, errors, or crash logs)

https://github.com/user-attachments/assets/f5f0729f-27f9-4b7d-ad78-180f6fa5215e

### Steps to Reproduce

1. Chart editor
2. Create or select a hold note
3. Scroll out of view of the hold note
4. ??????
5. Bug reproduced!

Issue #5064, opened 2025-05-11
Bug Report: [Crash Editor] Having Hold Note Strumlines in Chart Editor leaking RAM & selection box issue


Starexify:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Version

0.6.4

### Description (include any images, videos, errors, or crash logs)

When using Chart Editor I met these problems:

1&2. Selection Boxes for hold notes looking like they're overlapped and RAM leaking from having Hold Note Strums (because they have gaps between them?):

https://github.com/user-attachments/assets/05a455a9-2996-49f7-b83d-4ec67bc50313

2.5. RAM Leaking when staying at the start of the Note Grid:

https://github.com/user-attachments/assets/10e9623d-f9c5-4622-ab83-33c6db3580b5

3. Hold Note Strums increasing RAM Usage insignificantly when being passed through (not sure if fixable but I just found it unintended to happen)

https://github.com/user-attachments/assets/e83b6ee0-40d8-4969-941f-7882affafc53

### Steps to Reproduce

1. Open Chart Editor
2. Try what I did and place strumlines randomly and leave gaps sometimes between for that to happen
2.5. High RAM Usage/Selection Boxes looking like theyre overlapped

PR #5073, opened 2025-05-12
[BUGFIX] Chart Editor - Fix duplicating selection boxes 

NotHyper-474:
## Linked Issues
This seems to fix a few of the issues reported in https://github.com/FunkinCrew/Funkin/issues/5064, but I'm not too certain to link it.

## Description
`displayedNoteData` would be populated with duplicate notes for some reason, seems like it was just that `fastIndexOf` didn't work very well. This would cause multiple selection boxes and seemingly multiple hidden note sprites.

## Screenshots/Videos

https://github.com/user-attachments/assets/00ca7799-b0b7-4bbe-809c-3c12dbb2bcc1


Lasercar:
Fixes https://github.com/FunkinCrew/Funkin/issues/4277

NotHyper-474:
> Fixes #4277

I cannot reproduce that one even on a 0.5.3 build so maybe not.

Starexify:
Aight it fixes my #5064 issue it seems, thanks 👍

EliteMasterEric:
> `O(log n) + O(k + l)`

I'm pretty sure that's not how time complexity works, but it seems like it'll work fine regardless.

NotHyper-474:
> I'm pretty sure that's not how time complexity works, but it seems like it'll work fine regardless.

You're most certainly right lol. But the point I wanted to get across is that it comes with a cost, even though it's probably negligible unless we're considering a chart completely filled with notes of equal time (or so I believe would be a worst case scenario)

EliteMasterEric:
Note to self: reproduction of this issue requires moving the note to a row which has another note in a different column, it doesn't happen when there's a lone note.

PR #5091, opened 2025-05-14
[ENHANCEMENT] Chart Editor Leave Prompt

anysad:
## Description
* Whenever leaving the chart editor, there always was a problem, that there wasn't a prompt that asks people whether or not they actually want to leave, instead, it would just exit instantly.
* From what can be seen, people actually want this:

image

## Screenshots/Videos
screenshot-2025-05-14-22-40-53


anysad:
> 

pretty su... wait what the fuck?

PR #5108, opened 2025-05-17
[ENHANCEMENT] Sanitize `.json` text before parsing

KoloInDaCrib:
## Description
Sometimes when converting `Bytes` directly into a `String`, a bit of gibberish would be left at the start or end, causing the json parsing process to throw an error. This PR adds automatic removal of said gibberish before parsing a json text with `SerializerUtil` and before any registry json is parsed, which would spare modders having to do this themselves or struggling to find the issue.


cyn0x8:
may be cleaner to do `input = input.substring(input.indexOf('{'), input.lastIndexOf('}') + 1)`?

PR #5185, opened 2025-06-01
Vulnerability Patch: Blacklist `funkin.util.macro.*`

NotHyper-474:
## Linked Issues
N/A

## Description
Not only it has `CompiledClassList` from which you can retrieve `sys` and Newgrounds classes, it has nothing of use to mods.
Credits to @KoloInDaCrib for discovering the security vulnerability.

Delightfully Devilish Script Example:
```haxe
var classLists = [
  CompiledClassList.get("package~funkin.api.newgrounds~recursive"),
  CompiledClassList.get("package~sys~recursive")
];

for (list in classLists)
{
  for (class in list)
  {
    var className = Reflect.getClassName(class);
    if (className == "sys.io.Process")
    {
      var proc = new class('echo "HELLO WORLD" > FUCKKKKK');
      proc.close();
    }

    if (className == "funkin.api.newgrounds.Medals")
    {
      class.awardStoryLevel('weekend1');
    }
  }
}
```

## Screenshots/Videos
N/A

### DIFF
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -348,6 +348,16 @@ class PolymodHandler
       var className:String = Type.getClassName(cls);
       Polymod.blacklistImport(className);
     }
+
+    // `funkin.util.macro.*`
+    // CompiledClassList's get function allows access to sys and Newgrounds classes
+    // None of the classes are suitable for mods anyway
+    for (cls in ClassMacro.listClassesInPackage('funkin.util.macro'))
+    {
+      if (cls == null) continue;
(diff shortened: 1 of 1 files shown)


anysad:
using a macro to blacklist macros, a bit ironic

NotHyper-474:
> using a macro to blacklist macros, a bit ironic

Even more ironic is that this same macro is used to blacklist the sys and newgrounds classes, which is the reason they are in `CompiledClassList` in the first place.

anysad:
PFFT, that was a bit unexpected, quite funny to be honest

Issue #5206, opened 2025-06-08
Bug Report: Disappearing notes/events when switching songs in the Chart Editor


qt2k4:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.4

### Description (include any images, videos, errors, or crash logs)

For some reason, switching from a variation of a song (Erect, Pico, etc.) to a song that doesn't have that variation will cause some of the notes and events on that song to be missing.
It only affects the notes for the first difficulty loaded (which will invaraibly be Easy or Normal due to the nature of the bug), so this bug definitely has something to do with how things are loaded.

### Steps to Reproduce

Load a variation of a song in the Chart Editor (Erect, Pico or BF), then switch to a song that's longer and doesn't have that variation

Lasercar:
~~Could you try testing the changes in this PR here: https://github.com/FunkinCrew/Funkin/pull/4949~~

~~If it's not fixed by that, then it's definitely caused by something else~~

nvm, it is fixed

Basically, @Hundrec was opening an FNFC that had easy and normal difficulties for the erect variation metadata rather than the erect and nightmare difficulties the erect variation chartdata has, so of course there weren't any notes showing up!

Funnily enough, I have a fix of sorts for such situations in PR #4278:
https://github.com/FunkinCrew/Funkin/pull/4278/files#diff-ffcf4bc47099a77c7fd8d78217e4f2af499c7fe9b1aa878b47819b30177fd9e5R1299-R1305

Though I probably should also (or instead) make it add any unique difficulties used for the notes as well...?

PR #4949, opened 2025-05-03
[BUGFIX] Chart editor - Use first difficulty only as fallback

Lasercar:
## Does this PR close any issues? If so, link them below.
Fix #4947
## Briefly describe the issue(s) fixed.
A fix for my other PR. Use the first available difficulty as a fallback if the currently selected one cannot be found.
## Include any relevant screenshots or videos.

https://github.com/user-attachments/assets/d3ba1303-3439-4e36-90cd-2b9b29291e81

PR #5208, opened 2025-06-08
[BUGFIX] Fix Chart Editor forgetting Notes when changing to a Song with different Variations

KoloInDaCrib:
## Linked Issues
Closes https://github.com/FunkinCrew/Funkin/issues/5206

## Description
Changing the difficulty causes the `noteDisplayDirty` variable to be set to True... and the note displays to be refreshed, however this doesn't get called properly because the valid variation is checked for AFTER the difficulty, not the other way around, which causes the notes to only be rearranged. This PR fixes it.

## Screenshots/Videos

https://github.com/user-attachments/assets/846aa621-c85c-45a2-93ee-5b5d31646448

### DIFF
--- a/source/funkin/ui/debug/charting/handlers/ChartEditorImportExportHandler.hx
+++ b/source/funkin/ui/debug/charting/handlers/ChartEditorImportExportHandler.hx
@@ -117,12 +117,13 @@ class ChartEditorImportExportHandler
   {
     state.songMetadata = newSongMetadata;
     state.songChartData = newSongChartData;
-    state.selectedDifficulty = state.availableDifficulties[0];
 
     if (!newSongMetadata.exists(state.selectedVariation))
     {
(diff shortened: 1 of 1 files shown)


NotHyper-474:
True... 

I guess that means https://github.com/FunkinCrew/Funkin/pull/4949 unintentionally fixes it as well.

KoloInDaCrib:
> I guess that means #4949 unintentionally fixes it as well.

whoops didnt realize Lasercar unintentionally fixed the issue already, closing this PR

Hundrec:
We miiight actually need this one
The other one doesn't seem to fix the notes disappearing

Lasercar:
> We miiight actually need this one The other one doesn't seem to fix the notes disappearing

~~Turns out, the bug isn't fixed by my PR, which does the same thing, so it's something else which is doing this...~~

Nope we tripping

PR #5211, opened 2025-06-08
Chart Editor variation crash fix but better this time

dombomb64:
## Linked Issues
None

## Description
Currently, the Chart Editor ignores `playData.characters.instrumental` in the metadata when loading a song's instrumental. This PR points it in the right direction.

## Screenshots/Videos
In this video, I test out base game variations as well as two variations that have no `instrumental` value and therefore use the default instrumental.
https://github.com/user-attachments/assets/25f95579-3b44-43b5-9943-92e7261d9a3b

### DIFF
--- a/source/funkin/ui/debug/charting/handlers/ChartEditorImportExportHandler.hx
+++ b/source/funkin/ui/debug/charting/handlers/ChartEditorImportExportHandler.hx
@@ -62,13 +62,25 @@ class ChartEditorImportExportHandler
 
     for (variation in state.availableVariations)
     {
-      if (variation == Constants.DEFAULT_VARIATION)
+      var instId:String = '';
+      var firstDiff:Null<SongDifficulty> = song.getDifficulty(null, variation);
+      if (firstDiff != null && firstDiff.characters != null)
       {
-        state.loadInstFromAsset(Paths.inst(songId));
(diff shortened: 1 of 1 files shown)


Lasercar:
If you add a line break between the video and any text, it'll display as an embedded video, rather than a link.

Lasercar:
Also, fixes #5212

EliteMasterEric (changes requested):
I just tested in-game and this PR seems redundant since it works without the fix. Can you check if the original problem is still present, and make an issue to track if it is?

Hundrec:
I'm still unsure what this PR fixes, so I'm going to close it for now. If you can provide sufficient clarification, I'll open it again.

dombomb64:
Oops, I... kinda forgot about this PR... I think this PR was trying to fix an issue where, if I loaded a variation that also used the base instrumental (hence the "backup singer" stuff in the example video), then it would crash in the chart editor?
I'm not sure why it was behaving weird for me, because from what I recall, I didn't remember having anything on that messed with what audio files the game loaded. But if I were to check and it didn't turn out to be that, maybe I had forgot to pull changes to my fork and it was an already-resolved issue with the specific commit I compiled from? Who knows?
Anyways, sorry about all the flawed logic and lack of information on my part.

PR #5222, opened 2025-06-11
[VULNERABILITY PATCH] Blacklist `hscript.Interp`

KoloInDaCrib:
## Linked Issues
N/A

## Description
Interp's `cnew` function allows for instantiating blacklisted classes, such as `sys.io.Process`.
Note that it's also technically possible to access Polymod's `_interp` field considering scripted classes have the `_asc` field, which contains `_interp`.

Evil and fucked up script example:
```
var interp = new Interp();
var proc = interp.cnew("sys.io.Process", ['echo "HELLO WORLD" > FUCKKKKK']);
proc.close();
```

## Screenshots/Videos
N/A


AbnormalPoof:
Actually told Eric about this a few weeks ago! It's probably done internally already, sowwy.........

PR #5236, opened 2025-06-17
[BUGFIX] Fix scrolling Diffs in the Chart Editor when multiple Variations have the same Diff Name

KoloInDaCrib:
## Linked Issues
N/A

## Description
Formats the current difficulty as `[difficulty]-[variation]` when running the check for the currently selected index for incrementing/decrementing current difficulty. Also makes the `playbarDifficulty` text show the current variation in parentheses, for clarity.

## Screenshots/Videos

https://github.com/user-attachments/assets/f8bb4005-2e18-4fd9-920b-06079dcff368


Lasercar:
`Nightmare (Erect)`

I wonder if anyone's had one like that

Issue #5012, opened 2025-05-05
Bug Report: [Chart Editor] Hovering over Note Sustains Trails in chart editor grows memory constantly


Starexify:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.4

### Description (include any images, videos, errors, or crash logs)

For some reason in the Chart Editor while holding the cursor over Hold Notes (Sustain Trails) is constantly growing the memory and can lead to stutters aswell when using mods and such. It does clear but it doesn't seem like intended behavior

### Steps to Reproduce

1. Open the game
2. Open Chart Editor and place a hold note
3. Hover your mouse over the sustain trail and look at your memory going up

Lasercar:
_Holding_ the sustain trail of a hold note constantly increases the memory

Hundrec:
Really?
@Starexify Can you clarify whether the hold note needs to be clicked and held, or just hovered over?

Starexify:
> Really?
> @Starexify Can you clarify whether the hold note needs to be clicked and held, or just hovered over?

As stated, when you hold your mouse over them, without clicking on them or anything

Lasercar:
Ha, well that was an easy fix.

Hundrec:
Doesn't seem to happen on `develop` anymore

PR #5247, opened 2025-06-18
[BUGFIX?] "Fix" chart editor sustain trail memory leak

NotHyper-474:
## Linked Issues
`None... yet`

## Description
This doesn't fix the leak per se, just makes it so it calls the function that leaks memory only when necessary. As you can see in the video dragging a hold left and right like crazy will still increase memory usage because the trail's graphic is being updated.

## Screenshots/Videos

https://github.com/user-attachments/assets/9814ed3d-1bac-41f2-a7c9-023c312d01eb


source/funkin/ui/debug/charting/components/ChartEditorHoldNoteSprite.hx
```diff
@@ -71,7 +72,7 @@ class ChartEditorHoldNoteSprite extends SustainTrail
   }
 
   @:nullSafety(Off)
-  function updateHoldNoteGraphic():Void
+  public function updateHoldNoteGraphic():Void
```
NotHyper-474, line 75: Felt like mentioning the reason I did this instead of using `@:privateAccess` is because there's a equivalent function, `updateHoldNotePosition` which is public. Made more sense to me making them both public

NotHyper-474:
Just realized this unintentionally fixes the leak in https://github.com/FunkinCrew/Funkin/issues/5012 since it was being caused by the note style setter always triggering a trail update

AbnormalPoof:
You made revisions after this PR was pulled internally so we have to get those in as well before merging aaahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

NotHyper-474:
> You made revisions after this PR was pulled internally so we have to get those in as well before merging aaahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

Ah, eto... bleh!

image

Hundrec:
Looks like the leak doesn't happen anymore internally with only your first commit, so we won't be merging in the other commits.
I'll merge this as-is though since `develop` is temporary

Issue #5244, opened 2025-06-17
Bug Report: Exiting Freeplay after selecting a character breaks the transition


Hundrec:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.4 `develop`

### Description (include any images, videos, errors, or crash logs)

Pressing Escape right after leaving Character Select causes Freeplay to become translucent as it exits. Then, the album and album text linger for a moment before disappearing.

![Image](https://github.com/user-attachments/assets/3115567a-259c-457f-9b7c-4e156de2b014)

### Steps to Reproduce

1. Enter Character Select
2. Select a character
3. Press Escape as soon as you're back in Freeplay

PR #4935, opened 2025-05-02
[BUGFIX] Freeplay - After intro, busy = false only if no capsule options open

Lasercar:
## Does this PR close any issues? If so, link them below.
Fixes #4918
## Briefly describe the issue(s) fixed.
Title. If you opened the options for a song before the intro ends, busy (which was set to true) will still be set to false, allowing you to perform regular inputs again.
## Include any relevant screenshots or videos.

https://github.com/user-attachments/assets/1151e51b-334b-4bef-97a1-2599e0128883

PR #5251, opened 2025-06-18
[BUGFIX] Disabled Inputs During Freeplay Transition After Exiting Character Select

JackXson-Real:
## Linked Issues
None

## Description
Pressing ESC during the transition from character select to freeplay causes some weird behavior. I was inspired to fix this because of #5244. The bug shown in this issue was fixed, but I still think that exiting the freeplay menu during its transition from character select is unintended because of how weird/buggy it looks, so I disabled inputs all together.

NOTE: ChatGPT helped me make this fix, I do not take full credit for this.
### Before Fix:

https://github.com/user-attachments/assets/66b401a9-8a52-4e42-b94a-0ff4efc8c905

### After Fix:

https://github.com/user-attachments/assets/b9d93a0a-ab1c-4e55-990b-3622a1454e34

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -1215,21 +1215,18 @@ class FreeplayState extends MusicBeatSubState
   {
     busy = true;
     FlxG.keys.enabled = false;
-
     if (_parentState != null) _parentState.persistentDraw = false;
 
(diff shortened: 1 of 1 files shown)


Lasercar:
Isn't this PR basically just this one? 

https://github.com/FunkinCrew/Funkin/pull/4935

JackXson-Real:
> Isn't this PR basically just this one?
> 
> #4935

Ive tested your PR, and it does not fix the issue thats fixed in this PR.

JackXson-Real:
Duplicate of #4935

PR #5272, opened 2025-06-27
[ENHANCEMENT] Ability to prevent the Freeplay Song preview from playing

KoloInDaCrib:
## Linked Issues
N/A

## Description
There has been a big debate whether or not having the song labelled as "NEW" should prevent the preview from being played or not. This PR provides an alternative solution so that both parties are satisfied.
If the song's script has a function override `canPlayPreview` (with params being the same as with `isSongNew`), with the return `false`, the freeplay song preview would be replaced by the `Random` track. However, if multiple songs in a row have `canPlayPreview` disabled, the 'Random' track wouldn't get reset.

<details>
<summary><a>Script Example</a></summary>

```
import funkin.play.song.Song;
class PhillyNiceSong extends Song
{
	public function new() {
		super('philly-nice');
	}

	public override function canPlayPreview(currentDifficulty:String, currentVariation:String):Bool {
		return false;
	}
}
```
</details>

## Screenshots/Videos

https://github.com/user-attachments/assets/83443235-97ee-47a6-8da3-bb58f007a540


Hundrec:
I think this is for the best
image

Issue #3772, opened 2024-10-25
Bug Report: Hold Note Covers sometimes don't display


Hundrec:
### Issue Checklist

- [X] I have properly named my issue
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.5.3

### Description (include any images, videos, errors, or crash logs)

Occasionally, hold note covers won't display. This seemed to start happening in 0.5.0 and has been present since then.

This can happen under any conditions.

https://github.com/user-attachments/assets/92c76fec-4ea2-4af5-a2d8-f189454ee17b

### Steps to Reproduce

1. Play a song with hold notes
2. Hit the hold notes

MAZ12211:
I'd call this an exploit more because you could hit only the starting note of the hold notes and you don't get punished for it besides having a lower score lol
I hope they address this at some point because it feels unintentional

Hundrec:
I think what you mentioned is completely intentional, as legacy versions of the game always punished you for dropping hold notes.
In my opinion, the system where only an incentive is in place is better.

Hundrec:
For context, view this issue: https://github.com/FunkinCrew/Funkin/issues/2909

MAZ12211:
Ah got it
I was about to say the legacy versions punish you by decreasing your health a lot (considering the hold notes at the time was a row notes) and score together

Later versions does the latter only

trayfellow:
Ah i've experienced this bug before, but for the life of me i couldn't find a way to reproduce it, thanks!

Hundrec:
No problem! Funnily enough, I only found out how to reproduce it after submitting this report!

ztgds:
this bugs been driving me insane ngl 😭😭

Lasercar:
I'm pretty sure this is intentional:
![Image](https://github.com/user-attachments/assets/42da7e68-a617-4d79-92df-a25d4797c37e)

PR #5275, opened 2025-06-30
Fix a bug where holding down a note after dropping the previous one would sometimes make its cover over not show.

T5mpler:
This fixes [Issue #3772](https://github.com/FunkinCrew/Funkin/issues/3772)

I was able to fix this while adding hold covers to DnB 3.5 so I thought I'd submit a PR fixing it here.

Essentially, sometimes if you drop a hold note, and then you hit one right after (from my friends and I's experience, the hold note has to be RIGHT after the dropped one), the one you just hit doesn't give a cover. 

This bug happens periodically, (which is why I say _sometimes_), from my experience it takes a couple tries to get it but when you do end up getting it the first time it happens all the time.

I tested it with Lit Up (BF Mix) and Bopeebo since it was most easily debuggable, and it didn't appear to happen anymore.

**BEFORE (Bug happens at 0:15):**
https://github.com/user-attachments/assets/48cf9821-ad4e-46b6-89b3-9781e4f16033

**AFTER:**
https://github.com/user-attachments/assets/ed7ac8cd-bbc4-4ad4-a7a1-8c93e79a3d91

### DIFF
--- a/source/funkin/play/notes/NoteHoldCover.hx
+++ b/source/funkin/play/notes/NoteHoldCover.hx
@@ -71,6 +71,8 @@ class NoteHoldCover extends FlxTypedSpriteGroup<FlxSprite>
 
     this.visible = false;
 
+    holdNote.cover = null;
+
     if (glow != null) glow.visible = false;
     if (sparks != null) sparks.visible = false;
   }


Hundrec:
This bug has caught my eye for over half a year, and a simple null setting fixes it?

We'll check this out, thanks!

NotHyper-474:
Could it be yet another one-line banger? :eyes:

ztgds:
cant spell t5 without the T of GoaT5mpler

AbnormalPoof:
We'll take a look at this one, thanks!

AbnormalPoof (approved):
Thank you!

PR #5289, opened 2025-07-03
Experimental Polymod

EliteMasterEric:
This version of Polymod includes the following:

## New Language Features/Syntax

- Allow accessing static fields of other scripted classes.
    - NOTE: Pending merge of larsiusprime/polymod#194 You can't import scripts with a different `package`, they have to have matching packages for now.
- Added support for variables with getters and setters.
- Added support for defining enums in scripts.
- ~~Partial~~ **FULL** support for abstracts. You can finally use FlxColor properly!
  - Note that you can't access instance methods of abstracts directly, but you can still use them by calling the static method and passing the instance as the first argument.
- Added support for aliased imports (`import funkin.graphics.FunkinSprite as Sprite;`)
- Fixed try-catch blocks being pretty much completely broken (the catch block will run if there's an error, and the code afterwards will now be run properly!)

## Additional Fixes
- Linux now detects file paths as case-insensitive.
- Stricter handling of invalid syntax with good error reporting
    - NOTE: This will break some mods but only if they had bad syntax to begin with!
- Scripts now throw an error when attempting to declare a variable or function with the same name as an existing field.
    - The previous behavior was that the last function in the file with that name gets used. I spent 4 hours trying to figure out why changing a function didn't do anything, this also breaks a couple mods but only if they had redundant/broken code to begin with. Consider this change a blessing.
- Fixed being able to modify final variables (like values from `Constants`).
- Fixes for calling a function on a nullable object with `?.` throwing an error instead of returning null.
- Variable types will now sometimes be queried when initializing Arrays or Maps to ensure the correct class is instantiated.
- Fixes to `Assets.list()` returning inconsistent values, and fixed `library.list()` returning values from every library.

I'm still merging stuff so continue testing to make sure everything works okay.

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -192,7 +192,7 @@
       "name": "polymod",
       "type": "git",
       "dir": null,
-      "ref": "6a6b5cdfc7c4a209c4f49913021a132a8ee96de5",
+      "ref": "23a27ec2458aff5dae44de5a383236e2bd4b644a",
       "url": "https://github.com/larsiusprime/polymod"
     },
     {
--- a/hmm.json
+++ b/hmm.json
@@ -77,7 +77,7 @@
       "name": "hscript",
       "type": "git",
       "dir": null,
-      "ref": "27c86f9a761c1d16d4433c4cf252eccb7b2e18de",
+      "ref": "b6799be77b3d194a7bd1ae106221ed935a3bbb48",
       "url": "https://github.com/FunkinCrew/hscript"
     },
     {
@@ -192,7 +192,7 @@
       "name": "polymod",
       "type": "git",
       "dir": null,
-      "ref": "0fbdf27fe124549730accd540cec8a183f8652c0",
+      "ref": "45beae7151ab80321d8dc1318d0b1e9994d89426",
       "url": "https://github.com/larsiusprime/polymod"
     },
     {
--- a/hmm.json
+++ b/hmm.json
@@ -77,7 +77,7 @@
       "name": "hscript",
       "type": "git",
       "dir": null,
-      "ref": "b6799be77b3d194a7bd1ae106221ed935a3bbb48",
+      "ref": "d60bb2947fa609fdc875ccfae89666a6984eeaf2",
       "url": "https://github.com/FunkinCrew/hscript"
     },
     {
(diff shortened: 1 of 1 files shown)


lunacynne:
yess this is everything i needed

pucci-jojo

MAZ12211:
>  FULL support for abstracts. You can finally use FlxColor properly!
> Scripts now throw an error when attempting to declare a variable or function with the same name as an existing field.
> Fixes for calling a function on a nullable object with ?. throwing an error instead of returning null.

Instant 10/10

lemz1:
> Partial FULL support for abstracts. You can finally use FlxColor properly!
Note that you can't access instance methods of abstracts directly, but you can still use them by calling the static method and passing the instance as the first argument.

Keep in mind that operator functions have to be called explicitly.

Meaning, when adding two FlxColors together you cant just do: `color1 + color2`, instead you have to do `FlxColor.add(color1, color2)`

CrusherNotDrip:
type of shit my body felt upon discovering this pr IMG_9228

charlesisfeline:
the ultimate pr 🎉

EliteMasterEric:
Just fixed a bug with variable resolution in static functions.

EliteMasterEric:
New update fixes being able to modify final variables, complete with a custom error message:

463350505-462eb250-2d94-4604-bc25-d2d974cdfa83

macohi:
It got delayed??????????

AbnormalPoof (approved):
THATS FUCKING HSCRIPT RIGHT THERE. NONE OF THAT PANSY ASS `ScriptedClass.init()` BULLSHIT. PEOPLE `class.someVariable`, PEOPLE `get_` ON THE VARIABLES, PEOPLE DELIVER THEIR NEW MODS ON THE BEST POLYMOD BRANCH. FUCKING HARD CORE `import path.to.enum;` FUCK IT `EInvalidAccess` IT `new ClassName()` TIME SHIT. TAKE IT TO THE `ClassName extends Object`. STATIC FIELDS IN PLACES YOU DON’T EVEN REMEMBER. WE WIN TOGETHER WE CELEBRATE TOGETHER. HSCRIPT IS BACK BABY.

PR #5301, opened 2025-07-09
[BUGFIX] Use the actual character ID for `currentCharacterId`

TechnikTil:
## Description
Recently, a PR merged in `develop` made it so `currentCharacterId` would be the Freeplay Style ID, making it so it doesn't use the character specified directly. This PR fixes that issue by making `currentCharacterId` is the actual ID of the character.


Hundrec (approved):
I'm mergin' it

PR #5401, opened 2025-07-18
[BUGFIX] Fix 'New Variation' submit assigning charter to variation

### LINKED ISSUE #5400: Bug Report: Adding a variation fails and wipes the `default` one's difficulties
### Issue Checklist

### Platform

Compiled from GitHub Source Code

### Version

`develop` branch, also applies to 0.7.0

Title. Not only it does NOT create the variation it also completely wipes the default variation's difficulties. I've also tested it on the mobile version and this bug is also there, making this a 0.7.0 issue too

https://github.com/user-attachments/assets/66fb3c5e-ea25-43c7-b5e5-9aad40c99c22

### Steps to Reproduce

uhhhh can't be bothered to write how to repro i think the video is enough

NotHyper-474:
## Linked Issues
Fixes #5400 

## Description
A PR made a change to `SongMetadata`'s constructor to fix the charter not being saved but forgot to change many instances where it was being used, causing the variation to be `null` (which fallsback to `default`) and the charter to have the variation's name.

This PR changes it so all instances of `new SongMetadata` assign the arguments correctly which should fix some other unnoticed issues.

P.S: I've also noticed the New Song wizard window is missing a charter field, I would fix that but that's gonna require an assets PR.
Although this stills fixes a bug where the charter field ends up with the variation name (though now it'll be empty since it cannot be specified without the field for that)

## Screenshots/Videos
TBA (maybe)


Lasercar:
eric...................

NotHyper-474:
> eric...................

image

what do you mean by eric...

Lasercar:
> > eric...................
> 
> what do you mean by eric...

Ah, I was more on about how it took adding and saving a variation to notice that I completely forgot about the all the new uses of the song metadata.

PR #5408, opened 2025-07-20
[BUGFIX] Only destroy paused sounds on restart if they have `autoDestroy` enabled

### LINKED ISSUE #5371: Bug Report: BF's Vocals play 'twice' after restarting in Week 3
### Issue Checklist

### Platform

iOS/iPadOS

### Browser

None

### Mobile Device Model

iPhone 8+

### Mobile OS Version

iOS 16.7.11

### Version

0.7.1 and was on 0.7.0

Lowkey sounds good if it was right timing, but if like you keep pausing and restarting repeatly or I think pause and resume repeatedly the song "Pico" (though possible its in other songs -- You will hear BF's Vocals playing twice, one with you playing, and one out of nowhere without you playing.

It is possible it's in Android.

### Steps to Reproduce

1. Go to the First song of Week 3.
2. Pause and Restart over and over
3. If in a right moment during gameplay, while you hit the notes of Bf's for his vocals, his Vocals should be repeating again and out of place in the song.

https://github.com/user-attachments/assets/218f6000-4843-49df-8053-19e6ab31bf17

NotHyper-474:
## Linked Issues
Fixes #5371

## Description
I've realized the train sounds on Philly Train stages were broken, sometimes working, sometimes not. Turns out it was because the stages reuse the sound, and the sounds pause system was destroying it. I had failed to notice this when I originally coded this system lol. So this makes it so the sounds are only destroyed if they also auto-destroy.

~I feel like this might solve #5371 although I'm not really able to reproduce it on the `develop` branch, though I did seem to experience the same bug on Philly Erect (thought that was apparently because of an oversight when I was coding this PR where I was stopping the auto-destroying sounds and not the other way around lel)~
## Screenshots/Videos
~I don't think a video would suffice~

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1502,7 +1502,9 @@ class PlayState extends MusicBeatSubState
         musicPausedBySubState = false;
       }
 
-      forEachPausedSound((s) -> needsReset ? s.destroy() : s.resume());
+      // The logic here is that if this sound doesn't auto-destroy
+      // then it's gonna be reused somewhere, so we just stop it instead.
+      forEachPausedSound(s -> needsReset ? (s.autoDestroy ? s.destroy() : s.stop()) : s.resume());
 
       // Resume camera tweens if we paused any.
       for (camTween in cameraTweensPausedBySubState)


NotHyper-474:
It seems this doesn't fully fix the issue where the train sound sometimes doesn't work — at least on the non-Erect stage. That's because the stage's own logic for pausing and resuming the train sound conflicts with the global paused sound system. The fix would be to simply remove the stage-specific handling, I'll have wait until 0.7.x's sauce code is finally out though.

But from my tests this does seem to fix #5371 :D

AbnormalPoof (approved):
Screenshot 2025-07-26 at 2 39 34 PM
look at my cool waveform

Issue #5493, opened 2025-07-22
Crash Report: Saving a character in the Animation Editor crashes


kagolytrp21:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.7.3

### Description (include any images, videos, errors, or crash logs)

 When I save the character in the animation editor, it just crashes 

Null Object Reference
  in funkin/ui/debug/anim/DebugBoundingState.hx#226
  in funkin/ui/debug/anim/DebugBoundingState.hx#314
  in flixel/FlxState.hx#225
  in flixel/FlxGame.hx#693
  in flixel/FlxGame.hx#628
  in flixel/FlxGame.hx#507

anysad:
Did you save the Character Data or did you save the Offsets of the character?

kagolytrp21:
> Did you save the Character Data or did you save the Offsets of the character?

https://github.com/user-attachments/assets/882888ca-1aaf-48b1-be9d-c26d7a7534b2 i get like that

kagolytrp21:
> Did you save the Character Data or did you save the Offsets of the character?

i just save the character and just crashes

anysad:
do you click anything after saving?

kagolytrp21:
> do you click anything after saving?

no

PR #5519, opened 2025-07-23
[BUGFIX] Fix crash in Animation Editor after saving a character

anysad:
## Linked Issues
https://github.com/FunkinCrew/Funkin/issues/5493

## Description
* Before this PR, once a user has pressed `ESCAPE` button, the Dialog which displayed all of the information (including the modes, offsets, shortcuts, etc.) would close without having a proper ability to open it back. If a user tries to move or click on anything, it would crash the game. Now the issue is fixed.

## Screenshots/Videos
(TBA)


anysad:
True...

PR #5527, opened 2025-07-23
[ENHANCEMENT] Make Android Back button close the game in Title screen

### LINKED ISSUE #5349: Enhancement: Pressing the Back button on Title Screen should close the game
### Issue Checklist

- [x] I have properly titled my enhancement

Currently it does nothing so you have to exit the game via the Home button and then close it via the app overview thingIdontknowthenameof. It takes you back to the previous state on other screens (just like the in-game button does) so I don't see why not. 

Maybe this is already a thing on iOS, which would make this a bug I think, but I don't have any device of such to know.

I also know a in-game exit button would do the same job, but there's already an issue for that and this would still be nice to have for consistency with other mobile apps.

NotHyper-474:
## Linked Issues
Closes #5349 

## Description
My first mobile PR :D and I guess the second ever public mobile PR.
Just a lil' QOL change I suggested that's easily replaceable by an in-game Exit button that I decided to do anyway.

~TODO: There appears to be a small issue where you may have to touch on it more than once for it to work for some reason.~
Turns out that it's 'cause `justPressed` doesn't work too well in this case. Managed to fix it by replacing with `justReleased`.
## Screenshots/Videos
https://github.com/user-attachments/assets/f55c4129-0edc-47c9-ad9c-947c4d82f5cc

### DIFF
--- a/source/funkin/ui/title/TitleState.hx
+++ b/source/funkin/ui/title/TitleState.hx
@@ -228,11 +228,11 @@ class TitleState extends MusicBeatState
   {
     FlxG.bitmapLog.add(FlxG.camera.buffer);
 
-    #if desktop
+    #if (desktop || android)
     // Pressing BACK on the title screen should close the game.
     // This lets you exit without leaving fullscreen mode.
-    // Only applicable on desktop.
-    if (controls.BACK)
+    // Only applicable on desktop and Android.
+    if (#if android FlxG.android.justReleased.BACK || #end controls.BACK)
     {
       openfl.Lib.application.window.close();
     }


KoloInDaCrib:
fire wallpaper

NotHyper-474:
True...

AbnormalPoof:
The people have found out how to compile mobile, the world is ending.

KarimAkra:
Yeah never use justPressed (and for obvious reasons never use pressed either) for the android back button checks 
Only justReleased works properly and i think we can blame flixel for this

PR #5555, opened 2025-07-25
[TWEAK] Use FlxTween's and FlxTimer's global managers on pause/unpause

PurSnake:
## First Description
We use FlxTween's and FlxTimer's global managers stop/resume tweens and timers on game pause/unpause.

~That's it.~

TODO's:
- [ ] Implement as independent managers for Playstate

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -59,7 +59,7 @@ body:
     attributes:
       label: Version
       description: Which version are you playing on? The game version is in the bottom left corner of the main menu.
-      placeholder: ex. 0.7.3
+      placeholder: ex. 0.7.4
     validations:
       required: true
 
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -59,7 +59,7 @@ body:
     attributes:
       label: Version
       description: Which version are you playing on? The game version is in the bottom left corner of the main menu.
-      placeholder: ex. 0.7.2
+      placeholder: ex. 0.7.3
     validations:
       required: true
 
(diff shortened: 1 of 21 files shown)


KoloInDaCrib:
I personally disagree with this change for one reason:
What if a modder deliberately set a tween or a timer to inactive for a certain amount of time? Pausing and unpausing the game would re-enable them and cause an unintendional effect. Either that or the modder would have to write special cases if the `isGamePaused` variable is true, which in my opinion defeats the purpose of this PR.
I think a good alternative would be to have an `FlxTweenManager` and an `FlxTimerManager` that only update in PlayState when the song itself updates. That way a modder could attach their tweens and timers to it and avoid any kind of unintentional effect.

PurSnake:
> I personally disagree with this change for one reason: What if a modder deliberately set a tween or a timer to inactive for a certain amount of time? Pausing and unpausing the game would re-enable them and cause an unintendional effect. Either that or the modder would have to write special cases if the `isGamePaused` variable is true, which in my opinion defeats the purpose of this PR. I think a good alternative would be to have an `FlxTweenManager` and an `FlxTimerManager` that only update in PlayState when the song itself updates. That way a modder could attach their tweens and timers to it and avoid any kind of unintentional effect.

Hmm, quite interesting opinion
Need to implement

EliteMasterEric:
Aside from needing a good rebase/cleanup, this seems pretty nice, although yeah it would automatically unpause EVERY tween.

You could just do something like:

```haxe
FlxTween.globalManager.forEach((tween:FlxTween) -> {
	if (tween.active) {
	  pausedTweens.push(tween);
      tween.active = false;
    }
});
```

and then check if the tween is in the list when you unpause (making sure to clear the list).

PurSnake:
hmm, true

Starexify:
ts good, oh yuh

TechnikTil:
500 merge conflicts

PR #5565, opened 2025-07-27
Fix touch pointer positioning when in PlayState

SrtHero278:
## Description

While searching for a way to allow for the touch pointer to appear in PlayState, I realized this is not possible due to a positioning error. This pull request aims to patch said error in order to see the pointer plugin in gameplay.

(Originally marked as a draft for until I was able to compile and test a build of this branch, which has now occured)

### DIFF
--- a/source/funkin/util/plugins/TouchPointerPlugin.hx
+++ b/source/funkin/util/plugins/TouchPointerPlugin.hx
@@ -272,6 +272,7 @@ class TouchPointer extends FlxSprite
    * Used in TouchPointerPlugin's update method.
    *
    * @param touch The FlxTouch object containing the current touch input data.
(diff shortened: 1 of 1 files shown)


SrtHero278:
Took me a bit, but I was able to make a build and try it. Since I don't have a Mac or an android phone, I had to test and record on BlueStacks, but I got a friend to try it on their android device!

https://github.com/user-attachments/assets/127831af-5da8-4a4f-a832-3b48b955032f

SrtHero278:
And if it is worth noting, I had to revert "i love mergy conflicties" and surround NGSaveSlot with `#if FEATURE_NEWGROUNDS` in order to compile.

I will not include these changes myself incase it only works as a solution on my end.

image

moondroidcoder:
this change lowkey rad

Issue #5567, opened 2025-07-27
Crash Report: Spamming Q and E in the freeplay causes a crash


Michu0204:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [ ] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.7.3

### Description

https://github.com/user-attachments/assets/6dcb5bd5-8094-4560-9863-265aa937a65f

### Steps to Reproduce

1. Go to freeplay
2. spam Q and E, changing the categories

### Crash logs

[crash-2025-07-27-19-31-12.log](https://github.com/user-attachments/files/21455304/crash-2025-07-27-19-31-12.log)

Trofem:
this is recreatable, but a had a bit different error

<img width="1261" height="734" alt="Image" src="https://github.com/user-attachments/assets/916296e5-ee3a-4130-9ace-4aa8f6cc55b2" />

refferences this code

<img width="750" height="276" alt="Image" src="https://github.com/user-attachments/assets/04f9d4dc-6e19-40d0-8f8e-de72a059eb3d" />
line 174


 `spr.offset.x = offsetNum;`

It looks like while letter is removed but offset still occur. It looks not that hard to fix.
in class [LetterSort](https://github.com/FunkinCrew/Funkin/blob/72e0ff6c9bc70f0faf7ab1691caa15e225a89594/source/funkin/ui/freeplay/LetterSort.hx#L18) there are a lot of not null-save FlxTimer calls

JVNpixels:
<img width="633" height="284" alt="Image" src="https://github.com/user-attachments/assets/3a011a82-ebfc-493a-a5ff-d7cbe60b545c" />

I get this error when I reproduce the crash.

JVNpixels:
Looks like the controls don't get turned off when you leave the menu.

JVNpixels:
#5578, here i am to save you from Null Object References and Fatal Uncaught Exceptions

PR #5578, opened 2025-07-29
[CRASH-FIX] Fix Game Crashing After Spamming Through Letter Selection

JVNpixels:
## Linked Issues
#5567

## Description
When the player would spam the letter sorting keybinds (Q&E by default), and try to leave at the same time, the game would crash. This is due `LetterSort.hx` not checking if the controls were active, meaning you could still spam during the leaving transition of the freeplay menu. This would cause it to crash, and this P.R is here to check if the controls are on/off.

## Screenshots/Videos
Please wait patiently!


JVNpixels:
Merged! YES!

PR #5582, opened 2025-07-30
[ENHANCEMENT] Add Charter(s) as it’s own separate line in the pause menu

### LINKED ISSUE #5412: Enhancement: Add Charter(s) as it’s own separate line in the pause menu
### Issue Checklist

- [x] I have properly titled my enhancement

It’s a little bit annoying having to wait 10 seconds just to see who charted a song. I doubt adding it as a separate line would hurt the pause menu in any way.

JackXson-Real:
## Linked Issues
Fixes #5412 

## Description
The pause menu now displays the charter(s) of the song on a separate line instead of fading between artist and charter every 15 seconds. It doesn't take up much extra space on the pause menu, so I don't see a problem with it being like this. I know there have been other suggestions to reduce the timer to 5 seconds, but I personally do not want to have to wait any amount of time to see the charters.

## Screenshots/Videos

screenshot-2025-07-29-21-48-04

### DIFF
--- a/source/funkin/play/PauseSubState.hx
+++ b/source/funkin/play/PauseSubState.hx
@@ -194,9 +194,13 @@ class PauseSubState extends MusicBeatSubState
 
   /**
    * A text object which displays the current song's artist.
-   * Fades to the charter after a period before fading back.
    */
   var metadataArtist:FlxText;
+
+  /**
+   * A text object which displays the current song's charter.
+   */
+  var metadataCharter:FlxText;
 
   /**
    * A text object that displays the current global offset.
@@ -263,8 +267,6 @@ class PauseSubState extends MusicBeatSubState
     regenerateMenu();
 
     transitionIn();
-
-    startCharterTimer();
   }
 
   /**
@@ -287,8 +289,6 @@ class PauseSubState extends MusicBeatSubState
     // extension.admob.Admob.onEvent.remove(onBannerEvent);
     // #end
     super.destroy();
-    charterFadeTween.cancel();
-    charterFadeTween = null;
     dataFadeTimer.cancel();
     dataFadeTimer = null;
     hapticTimer.cancel();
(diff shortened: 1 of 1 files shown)


dombomb64:
I just had an idea, what if instead of there being dedicated fields for an artist and a charter, there could be a map of credits, with the keys being the role names and the values being lists of people? It'd probably be something like this:
```haxe
var songCredits:Map<String, Array<String>> = [
  "Artist" => ["Artist 1"],
  "Charters" => ["Charter 1", "Charter 2"],
  "Scripter" => ["Scripter 1"]
];
```
This method could already solve pluralization issues, but maybe instead of the keys being strings, they could be their own sort of "`PluralizableString`" class so that the base game could have those as ~~static constants~~ values in a static map that they could reuse across songs?
```haxe
var songCredits:Map<PluralizableString, Array<String>> = [
  PluralizableString.tryPluralString('Artist') => ["Artist 1"],
  PluralizableString.tryPluralString('Charter') => ["Charter 1", "Charter 2"],
  PluralizableString.tryPluralString('Scripter') => ["Scripter 1"]
];
```
```haxe
class PluralizableString {
  /**
   * A global list of `String`s that have `PluralizableString` counterparts.
   * People can feel free to add their own values, and anything that uses `tryPluralString()` will try to find the plural form in this `Map`.
   * 
   * Only use `globals.get()` if you need to access the `PluralizableString` value for some reason.
   * Otherwise, `tryPluralString()` is preferred.
   */
  public static final globals:Map<String, PluralizableString> = [
    'Artist' => new PluralizableString('Artist', 'Artists'),
    'Charter' => new PluralizableString('Charter', 'Charters'),
    'Scripter' => new PluralizableString('Scripter', 'Scripters'),
    'Composer' => new PluralizableString('Composer', 'Composers'),
    'Sprite Artist' => new PluralizableString('Sprite Artist', 'Sprite Artists'),
    'Background Artist' => new PluralizableString('Background Artist', 'Background Artists'),
    'Animator' => new PluralizableString('Animator', 'Animators'),
  ];

  public var singular:String;
  public var plural:String;

  public function new(singular:String, plural:String) {
    this.singular = singular;
    this.plural = plural;
  }

  /**
   * Tries to find a plural form of `str` in `globals` and return it, but returns `str` if none are found.
   */
  public function tryPluralString(str:String) {
    var result:Null<PluralizableString> = globals.get(str);
    if (result == null)
      return str;
    return result.plural;
  }
}
```
Let me know what you think of either of these implementations!

JackXson-Real:
> Let me know what you think of either of these implementations!

Idea is cool, but I feel like adding all of that would just clutter the pause menu too much. Furthermore, I am not a programmer, so this stuff is way beyond my skill level, so someone else would have to implement it.

dombomb64:
I might try to implement it while also having the lines with credits be smaller.
###### If I ever get around to that...

Issue #5430, opened 2025-07-21
Bug Report: Notes no longer scroll smoothly


JackXson-Real:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.7.2

### Description (include any images, videos, errors, or crash logs)

Before 0.7.0, notes were very smooth, but the update introduced a fix for music drifting away from the conductor which also caused this bug. Here is the code segment that causes it:

<img width="862" height="124" alt="Image" src="https://github.com/user-attachments/assets/24f7f78e-2c80-47b9-b9de-f101c314f018" />

In order for this to be fixed "we will need to find a precise way to measure time" according to @Hundrec. If anyone is able to find a fix for this, please PR it.

### Steps to Reproduce

Press and hold a sustain note

doggogit:
[Video of said sustain trail playing at a lower framerate](https://file.garden/aFBcDvmb4DlYDcr9/2025-07-21%2022-36-08(HB)2.mp4)

[Another video of Stress (Pico Mix) tankmen moving at a slower framerate(?) (could be placebo but still wanted to point it out)](https://file.garden/aFBcDvmb4DlYDcr9/2025-07-21%2022-22-36.mp4)

PR #5607, opened 2025-08-01
[BUGFIX] Smoother notes scrolling

KarimAkra:
Should be a fix for #5430 !!!

Would be epic for the people who are having this notes scrolling issue to try it out see if it does the job well 🙂

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -639,7 +639,7 @@ class PlayState extends MusicBeatSubState
   /**
    * The ratio for easing the song positon for smoother notes scrolling.
    */
-  static final MUSIC_EASE_RATIO:Float = 33.5;
+  static final MUSIC_EASE_RATIO:Float = 40;
 
   // TODO: Refactor or document
   var generatedMusic:Bool = false;
@@ -1080,7 +1080,7 @@ class PlayState extends MusicBeatSubState
       // The previous method where it "guessed" the song position based on the elapsed time had some flaws
       // Somtimes the songPosition would exceed the music length causing issues in other places
       // And it was frame dependant which we don't like!!
-      final easeRatio:Float = 1.0 - Math.exp(-MUSIC_EASE_RATIO * elapsed);
+      final easeRatio:Float = 1.0 - Math.exp(-(MUSIC_EASE_RATIO * playbackRate) * elapsed);
       Conductor.instance.update(FlxMath.lerp(Conductor.instance.songPosition, FlxG.sound.music.time + Conductor.instance.combinedOffset, easeRatio), false);
 
       // If, after updating the conductor, the instrumental has finished, end the song immediately.
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -636,6 +636,17 @@ class PlayState extends MusicBeatSubState
    */
   static final RESYNC_THRESHOLD:Float = 40;
 
+  /**
+   * The threshold for using normal song position updates.
+   * If the Conductor position deviates from the music by more than this amount, then the Conductor position will be set to the raw song time.
+   */
+  static final NORMAL_MUSIC_UPDATE_THRESHOLD:Float = 80;
+
+  /**
+   * The ratio for easing the song positon for smoother notes scrolling.
+   */
+  static final MUSIC_EASE_RATIO:Float = 30;
+
   // TODO: Refactor or document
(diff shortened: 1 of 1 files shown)


KarimAkra:
That's a vid of how the notes look like at 0.5x speed for people who just wanna see
https://github.com/user-attachments/assets/01c00b39-b74e-45d0-b9af-8c0a21d56376

KarimAkra:
And that's another video also at 0.5x speed but with more sustains!

https://github.com/user-attachments/assets/d9667d34-178c-4c8e-82df-779183b8127e

JackXson-Real:
Trying this out now, I'm a bit worried that this will bring back the offset hit windows (having to hit notes very early for them to be considered Sick!)

JackXson-Real (changes requested):
https://github.com/user-attachments/assets/54d111ab-d8fd-4440-a210-bbddd6ffc599

Yeah uhhhh, not only does this not fix the notes not being smooth, but this also happens. This one definitely needs a bit more work

KarimAkra:
> Yeah uhhhh, not only does this not fix the notes not being smooth, but this also happens. This one definitely needs a bit more work

Oh no it's 0.5.0 all over again...
It's bugged with offsets, should be easy to fix it

anysad:
> 8mb.video-GiU-JMEfPaFN.mp4 
> Yeah uhhhh, not only are the notes not scrolling smoothly, but this also happens. This one definitely needs a bit more work

hell yeah, swing mixes

realvirtu:
> 8mb.video-GiU-JMEfPaFN.mp4 
> Yeah uhhhh, not only are the notes not scrolling smoothly, but this also happens. This one definitely needs a bit more work

Welcome back strange input offset bug.

JackXson-Real:
> > Yeah uhhhh, not only does this not fix the notes not being smooth, but this also happens. This one definitely needs a bit more work
> 
> Oh no it's 0.5.0 all over again... It's bugged with offsets, should be easy to fix it

Re-request a review from me once you fix it, I'm still not convinced that fixing the swing mode bug will make the notes scroll smoothly

KarimAkra:
> I'm still not convinced that fixing the swing mode bug will make the notes scroll smoothly

We can tweak the lerp factor/ratio until we find something that actually works if `0.5` doesn't seem to work
that's honestly just a random number i threw and it looked good to me personally without any deeper investigation lol

realvirtu:
Does using lerp for this have a problem with framerate, or no because it lerps the song position to the music position?

KarimAkra:
> Does using lerp for this have a problem with framerate, or no because it lerps the song position to the music position?

There was an issue with the lerp on low FPS but got it fixed just now so it should work fine

KarimAkra:
> The notes are actually LESS smooth than in base 0.7.3 somehow, 

I adjusted the lerp ratio so it should be smoother

> the hit window feels a little bit offset, but I'm not entirely sure if it's just placebo. 

I think that's just you..? I set the playback rate to 0.7 and played through few songs while hitting bunch of notes early and late and the hit window seem completely fine

> Also changing your offset in the middle of a song can desync the song by a lot.

Not sure about this since i don't have wireless headphones to check it out myself but the offset seem to be applying properly

JackXson-Real (changes requested):
Ok so the arrows are smooth now, but there are still some issues. This time I made videos showcasing each one so they are easier to understand.
## Hit Window Offset
In this video you can see that you can get a sick by pressing the notes way below the strumline, but clicking them even a tiny bit late is a good. This issue was solved in 0.7.0 but came with the cost of less smooth notes. Whatever you changed seems to have brought it back. Personally I would rather have the notes be slightly less smooth in favor of accurate hit windows.

https://github.com/user-attachments/assets/de2b97b2-1a92-4ab2-8d52-f97100e3d1d6

## Song Skipping
This video shows an issue where the song is repeatedly correcting itself. Ignore second half of the video because Im an idiot and thought that was a bug for some reason 😭 

https://github.com/user-attachments/assets/a4175a63-e382-4772-b0f2-63dc38be4687

charlesisfeline:
> 8mb.video-GiU-JMEfPaFN.mp4 
> Yeah uhhhh, not only are the notes not scrolling smoothly, but this also happens. This one definitely needs a bit more work

cant run away from swing mode....

doggogit:
This seems to have fixed the issue!

Works just fine for me, you can also @ me for a re-review, if there is anything else to test

PR #5645, opened 2025-08-03
[ENHANCEMENT] Increase Freeplay Preview Volume To 80% To Match Random's Preview Volume

JackXson-Real:
## Linked Issues
None

## Description
Increases the volume of freeplay previews to 80%. They were previously 40% while Random was at 80% making it inconsistent. I think it's better to make the preview volume 80% instead of decreasing them to 40% since some people (including me) like listening to the previews once in a while, and it's a bit inconvenient to have to increase/decrease volume.

## Screenshots/Videos
Nah I'm good


JackXson-Real:
Preview volumes are now 70% as per Dave's request.

PR #5650, opened 2025-08-03
[ENHANCEMENT] Make both Start and Accept buttons work on Title Menu

MrMadera:
## Linked Issues

No

## Description

The title menu only worked with the start button on a gamepad, which is ok, but too many people would try to use the South button, notice that it doesn't work, and end up confused. This PR makes it so both the Start and South buttons work

## Screenshots/Videos

https://github.com/user-attachments/assets/79cd266c-d5a5-4129-b985-7bfec49365db


Hundrec:
Yeah this is a good idea

PR #5657, opened 2025-08-04
Add Charter field to New Chart dialog

NotHyper-474:
## Linked Issues
None

Adds a charter field to the song metadata dialog that shows up when you're creating a new chart.
This was only done for the add variation window in an earlier PR.
## Screenshots/Videos
image


Lasercar:
> ##### Also what the hell happened to the BPM field?

It's startled

Lasercar:
If you're wondering why I choose not to add it to the new chart dialog, well

I think it makes more sense if you edit the field once the chart is completed, as you would with say, a custom map, preemptively setting it just feels weird to me.

PR #5730, opened 2025-08-12
[OPTIMIZATION] Small Freeplay rendering optimizations

MaybeMaru:
## Description
I've been testing around with Nvidia Nsight to have more clear rendering debugging and found some issues that affect FreeplayState both in loading and performance. The main problem I noticed was with ``BGScrollingText``, since by creating multiple FlxText instances it also created seperate bitmaps and draw calls for each object with no batching.

I fixed this by making only one instance of text and making the rest of the sprites just copy the graphic of the main text sprite.
Also fixed the black overlay from the freeplay intro never getting set invisible, which saves some performance both with running the shader and rendering it's bitmap.

## Screenshots/Videos
Here's a small sample of the draw calls produced by ``BGScrollingText`` before and after this change.

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/db592d0d-35ca-409e-bc42-ca12110371d1" width="320" height="240" controls></video> | <video src="https://github.com/user-attachments/assets/eef386cf-47c2-44de-88e8-1c037ff22f95" width="320" height="240" controls></video> |

### DIFF
--- a/source/funkin/ui/freeplay/BGScrollingText.hx
+++ b/source/funkin/ui/freeplay/BGScrollingText.hx
@@ -4,70 +4,80 @@ import flixel.FlxObject;
 import flixel.group.FlxSpriteGroup;
 import flixel.text.FlxText;
 import flixel.util.FlxSort;
+import flixel.FlxSprite;
+import flixel.util.FlxDestroyUtil;
+import flixel.util.FlxColor;
 
 // its kinda like marqeee html lol!
 @:nullSafety
 class BGScrollingText extends FlxSpriteGroup
 {
-  var grpTexts:FlxTypedSpriteGroup<FlxText>;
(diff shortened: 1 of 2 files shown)


Lasercar:
Small optimisations?

But this is a `size: medium`

MaybeMaru:
Small as in a very slight performance improvement, not code size wise

Lasercar:
> Small as in a very slight performance improvement, not code size wise

So, `size: tiny` then?

MaybeMaru:
> So, `size: tiny` then?

`size: atom` if possible

NotHyper-474:
> > So, `size: tiny` then?
> 
> `size: atom` if possible

Your wish has been granted.

AbnormalPoof (approved):
very nice

Issue #5741, opened 2025-08-13
Bug Report: Random double notes in Lag Adjustments menu


anysad:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.7.3

### Description (include any images, videos, errors, or crash logs)

Whenever trying to test the input offsets in the `Input Offsets` test menu, there sometimes appear random doubles which are the UP and RIGHT arrows together at the same time.

https://github.com/user-attachments/assets/d8a19a75-a996-4175-b6f5-4b792ca51a79

### Steps to Reproduce

1. Open the game
2. Go to setting
3. Go to `Input Offsets` -> `Test`
4. Sometimes there are random double notes (up and right arrow together)

Hundrec:
Sorry anysad, this is an intentional feature requested by me personally

anysad:
<img width="375" height="500" alt="Image" src="https://github.com/user-attachments/assets/4df36aac-003d-4bfd-86ec-5a6d2bfc773f" />

PR #5742, opened 2025-08-13
[FIX] Fix random double notes in Lag Adjustment test menu

anysad:
## Linked Issues
https://github.com/FunkinCrew/Funkin/issues/5741

## Description
* Before this PR, for some reason there was left out code that was adding note data for UP direction for no apparent reason. Either it was for testing and forgot to be removed, or it was just a strange decision, who knows?


anysad:
grrr why does the status keeps deleting itself

PR #5743, opened 2025-08-13
[OPTIMIZATION?] Prevent script classes from being registered twice after hot-reloading

KoloInDaCrib:
## Linked Issues
N/A

## Description
Hot-reloading causes the function `forceReloadAssets` of `PolymodHandler` to be called, which loads back all mods and their script classes. This PR removes the line to register scripted classes.
On the surface, this change seems preposterous. However, reloading mods triggers the function `Polymod.init();` which already contains the line that is deleted in this PR. This means that this PR mitigates this monstrocity from occuring upon hot-reloading:
image


EliteMasterEric (approved):

EliteMasterEric:
Was wondering about this bug, thanks

PR #5745, opened 2025-08-13
LAST Fix for the song restarting bug!?!?!?!

KarimAkra:
I REALLY hope this WILL be the LAST time we ever see this damn cursed bug

i swear i sprouted some grey hairs strands while trying to figure it out

Anyways, the song restarting bug rn seem to only occur when you're lagging so i added a funni laginator that works by pressing space for who doesn't lag and wanna test this fix


KarimAkra:
Need public testing on this so we can FINALY get rid of this bug for certain

KarimAkra:
accidentally forgot to put one small important line earlier!

PR #5752, opened 2025-08-15
[OPTIMIZATION] Improve note batching

MaybeMaru:
## Description
Flixel batches quads when they all share the same rendering values (color, antialiasing, shader, etc). The problem with this is that, with each note having it's own hsv shader instance, batching is unavailable. Which can be pretty harmful for performance in instances with many notes on screen. To fix this I made it so the shader is disabled by default and only applied when neccesary (miss desaturation or hue changes).

## Screenshots/Videos
Here's the difference in draw calls needed checked through Nvidia Nsight.
| Before  | After |
| ------------- | ------------- |
| <video src="https://github.com/user-attachments/assets/fddddd8b-2d3d-4d28-857a-503b6b51b709" width="320" height="240" controls></video> | <video src="https://github.com/user-attachments/assets/4636800e-b052-4873-8bdd-cc2ac2cd0907" width="320" height="240" controls></video> |

### DIFF
--- a/source/funkin/play/notes/NoteSprite.hx
+++ b/source/funkin/play/notes/NoteSprite.hx
@@ -167,8 +167,6 @@ class NoteSprite extends FunkinSprite
   {
     noteStyle.buildNoteSprite(this);
 
-    this.shader = hsvShader;
-
     // `false` disables the update() function for performance.
     this.active = noteStyle.isNoteAnimated();
   }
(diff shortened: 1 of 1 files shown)


MAZ12211:
What's the song running in the background video called?

EliteMasterEric:
This optimization is cool, big thumbs up from me 👍

PR #5768, opened 2025-08-18
[BUGFIX] Fix New Rank being bigger than original ones

anysad:
## Description
* Before this PR, whenever you got a new rank on a specific song, the new rank would be a bit bigger than the ones that were already there. The problem was that in the original code each rank have a scale of `0.9`, but the new ranks got a scale of `1`. Now this is fixed.

## Screenshots/Videos

<div align='center'>
<table>
  <tr>
    <th>BEFORE (look at bopeebo rank)</th>
    <th>AFTER</th>
  </tr>
  <tr>
    <td>BEFORE</td>
    <td><video src="https://github.com/user-attachments/assets/248d0fe5-7d21-488d-bcc7-ddd77cff8aaa" alt="AFTER" width="250"/></td>
  </tr>
</table>
</div>

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -1098,7 +1098,7 @@ class FreeplayState extends MusicBeatSubState
       capsuleToRank.ranking.playAnimationEach(fromResults.newRank.getFreeplayRankIconAsset(), true);
     }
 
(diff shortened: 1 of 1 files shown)


moxie-coder:
This should be a size of nano lol

Hundrec:
I'm so proud of you for making a change to Freeplay that could not possibly break an unrelated Freeplay feature

PR #5773, opened 2025-08-19
[OPTIMIZATION] Multiple Freeplay batching fixes

MaybeMaru:
## Description
Looking through more of FreeplayState and its components I noticed a bunch more of instances where batching techniques could be used. Here's a list of all of the ones in this commit:
- Difficulty dots (by forcing the colored quad type to batch all dots together)
- Letter sort (same as difficulty dots, plus some organization using groups)
- Difficulty selector (by disabling the pure color shader when not necessary, this one is a bit hacky and not very important so I can remove it)
- Small offsets to capsules so the ones on the top and lower edges of the screen (which arent visible by the player) actually go completely offscreen and dont get unnecessarily rendered.
- Disable the BlueFade shader when not used (This one is very important to performance since, even though its not a complex shader, adding filters to a FlxCamera requires some buffers to be changed at render which can be very expensive.)
- Some general code cleanup like reducing Reflect, array getter calls, etc...

## Screenshots/Videos
Overall, these changes bring down the number of opengl calls by around 1300 calls on average. With no visual difference.
| Before  | After |
| ------------- | ------------- |
| image | image |


Lasercar:
Another Optimaru W

PurSnake:
:steamhappy:

PR #5811, opened 2025-08-25
[feat/flixel-animate] Fix cached CharSelect atlases not being destroyed when pressing F4

NotHyper-474:
## Description
Turns out pressing F4 to eject yourself from the character select sub-state causes some unforeseen consequences when you decide to go back to it and switch characters.....


AbnormalPoof (approved):
KRUEEEEE!!!!!!

PR #5820, opened 2025-08-27
[BUGFIX] Fix the `NEW` marking not appearing upon capsule generation

KoloInDaCrib:
## Linked Issues
N/A

## Description
If you (re-) generated the capsules (either just by entering Freeplay or changing the sorting of songs with Q/E), capsules that should be marked as "new" will not have that symbol until you scrolled directly to the song. This PR fixes it by making the `FreeplaySongData` use the base non-variated song to determine the current variation and then use that variation for other checks like if the song is new. 

## Screenshots/Videos

https://github.com/user-attachments/assets/b8ffad50-a5ce-4b00-af98-b35505a83623


NotHyper-474:
HE'S CHUDDING IT

Hundrec (approved):
`NEW` PR MERGED

PR #5841, opened 2025-08-30
[BUGFIX] Fix Freeplay Backing Card Beat Hit

VirtuGuy:
## Linked Issues
None

## Description
The backing card has a beat hit function that pretty much never runs because of the freeplay menu not properly updating the conductor. This PR changes it so that the conductor actually updates instead of only firing once like how it is currently. Due to this change, the backing card's beat hit function actually runs.

## Screenshots/Videos

Before

https://github.com/user-attachments/assets/8b38cdb1-8638-4de1-a280-e910f5ea3916

After

https://github.com/user-attachments/assets/855fbab9-4cf3-4bb2-9c6b-0b3e4f4282e2

Boyfriend also has the cool backing card flash thing as well. It's just that I have Pico selected because I love him a lot.

### DIFF
--- a/source/funkin/ui/freeplay/FreeplayState.hx
+++ b/source/funkin/ui/freeplay/FreeplayState.hx
@@ -1571,6 +1571,8 @@ class FreeplayState extends MusicBeatSubState
   {
     super.update(elapsed);
 
+    Conductor.instance.update(FlxG.sound?.music?.time ?? 0.0);
+
     #if FEATURE_TOUCH_CONTROLS
     if (backButton != null && !backTransitioning)
     {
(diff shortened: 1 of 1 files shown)


Hundrec:
Dude I just noticed this bug yesterday
Thank you for your service 💯

Michu0204:
this is NOT in 0.7.5

image

NotHyper-474:
This was ultimately reverted (https://github.com/FunkinCrew/Funkin/commit/c4ebd3175d1e80719ae80ce6fd79ea25f7e1e16a) because it was deemed too distracting.

PR #5847, opened 2025-08-30
[flixel-animate] Legacy bounds fixes

MaybeMaru:
## Description
Ran into some issues regarding the implementation of these bounds but I believe they should be fully fixed now.
Changed a bit the behaviour too so if ``legacyBoundsPosition`` is turned on itll override the check for ``applyStageMatrix `` too, for accuracy.
This PR also bumps the version of flixel-animate as the method it used to list assets wasnt playing nice with polymod, but now it should be fully fixed and working both for base game and modded assets.

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -59,7 +59,7 @@
       "name": "flixel-animate",
       "type": "git",
       "dir": null,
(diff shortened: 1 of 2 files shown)


Trayfellow:
I'm Tired Boss

AbnormalPoof (approved):
very nice

PR #5858, opened 2025-09-01
`[feat/flixel-animate]` Screen bounds fixes

MaybeMaru:
## Description
Continuation of https://github.com/FunkinCrew/Funkin/pull/5847
In the last PR I completely missed adding the extra necessary logic so flixel could check if the sprite was inside the camera bounds when using ``applyStageMatrix`` and/or ``legacyBoundsPosition``. Also bumped a bit the flixel-animate version since the library also had a small oversight related to the calculation of those.
It should be now completely fixed, but extra testing would be appreciated!

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -59,7 +59,7 @@
       "name": "flixel-animate",
       "type": "git",
       "dir": null,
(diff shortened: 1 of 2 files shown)


AbnormalPoof (approved):
There they are!
screenshot-2025-09-01-12-33-55

Issue #5905, opened 2025-09-04
Bug Report: Freeplay Rank Slam animation doesn't show the rank actually slamming


Hundrec:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.7.4 (and many previous versions)

### Description (include any images, videos, errors, or crash logs)

This bug has been here for as long as I can remember, but it wasn't always this way.

Here's a screenshot from 0.4.0 gameplay:

<img width="926" height="648" alt="Image" src="https://github.com/user-attachments/assets/cc134b63-3c4c-4e34-b795-43aafd7e4cb5" />

But now the rank never actually comes in and slams the capsule.

Can anyone find the old behavior and restore it?

This seems to have gotten lost alongside https://github.com/FunkinCrew/Funkin/issues/5605

### Steps to Reproduce

1. Start a song from Freeplay
2. Beat it with a new or higher rank
3. Watch the new rank magically appear

PR #5909, opened 2025-09-04
[BUGFIX] Make slamming rank visible

mikolka9144:
## Linked Issues
https://github.com/FunkinCrew/Funkin/issues/5905

## Description
Makes the rank visible while slamming

## Screenshots/Videos
I couldn't make any. The animation is too fast


Hundrec:
You know, this is a lot cleaner than what I did
I'm gonna test this out to see if it works anyway

Hundrec:
I'm impressed, this does make the slamming rank visible again!
The only slight issue was that it comes from the bottom right rather than the top center, but a little offsetting would make it perfect.

Thank you for opening this PR, it was just bad timing that I pushed a fix right before you did it.

PR #5924, opened 2025-09-05
[GITHUB] Add *the* checkbox to the Crash Template

VioletSnowLeopard:
#5678 for the crash template. If a mod can cause a bug, it could also cause a crash. I believe this checkbox should be added to the crash template too.


Hundrec:
Thought about this too, and I agree!

PR #5963, opened 2025-09-09
[OPTIMIZATION] Optimize BGScrollingText

FuroYT:
## Associated Funkin.assets PR
https://github.com/FunkinCrew/funkin.assets/pull/252

## Description
Fixes a major part of the Freeplay Memory Leak being associated with the `BGScrollingText` Class
This is a change for the scrolling text class and the new character backing card of the freeplay since in my PR I've remove the `funnyColor` variable in favor for `color` (since its not a group anymore but rather 1 text drawn over and over in the same frame)


TechnikTil:
no way?????

PR #5968, opened 2025-09-10
[GITHUB] Bump docs from 0.7.4 to 0.7.5

Nightfall-16:
also updates iOS 18 to iOS 26, and iPhone 16 to iPhone 17.

(also my first funkin pr so if anything wrong happens idk how to fix it lol.)

PR #5972, opened 2025-09-10
[GITHUB] Bump docs to 0.7.5

Nightfall-16:
redo of #5968, but hopefully better

again I hope this works cause im so bad at github

bumps iOS versions from 18 to iOS 26, and iPhone 16 to iPhone 17.


Hundrec (approved):
mmm 26

PR #5980, opened 2025-09-11
[ENHANCEMENT] Multi-directional singing animations

macohi:
## Linked Issues
N/A

## Description
This adds the ability for characters to play animations like "singDOWNLEFT" for when the down and left arrows are pressed and for other combinations

More specifically it means that every character that you'd want to perform this action wouldn't need a script file with the same function overwrite and only a few extra json lines of code and some extra animations are required.

## Screenshots/Videos
https://youtu.be/tESspqOKVQA


macohi:
How is this still open LMAO

PR #5985, opened 2025-09-12
[ENHANCEMENT] Option to disable Anti-aliasing in the Preferences menu.

ShadzXD:
## Description
Adds in an option to the preferences menu to disable global Anti-aliasing on sprites.

Should work on all sprites, tell me if there are any I missed.

## Screenshots/Videos
image

With Anti-aliasing off:

image

### DIFF
--- a/source/funkin/play/components/HealthIcon.hx
+++ b/source/funkin/play/components/HealthIcon.hx
@@ -459,7 +459,7 @@ class HealthIcon extends FunkinSprite
       loadAnimationOld();
     }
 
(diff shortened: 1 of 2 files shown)


ShadzXD:
Re-opened because it's now working properly.

ninjamuffin99:
I will decline this! I think we prefer to have art as smooth as possible, at expense of minor performance, than for someone to even have the option to change it

ShadzXD:
alright, all good 👍

Thanks for replying!

PR #6035, opened 2025-09-16
[ENHANCEMENT] FPS Independent health icon bopping

PurSnake:
## Linked Issues
Any issues about FPS Dependent HealthIcons

## Description
Now icons use their own tween (bopTween) to **bop**.
And also the speed of bopping accelerates at high BMP (noticeable at BMP > 190)

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1503,6 +1503,10 @@ class PlayState extends MusicBeatSubState
         cameraTweensPausedBySubState.add(cameraZoomTween);
       }
 
+      // Icons should not bop while the game is paused.
+      if (iconP1 != null && iconP1.bopTween != null) iconP1.bopTween.active = false;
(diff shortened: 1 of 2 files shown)


Trayfellow:
The best day of LeGooey's life.

PurSnake:
Well, enhancement then

Hundrec (changes requested):
This works great, except for one tiny detail that prevents me from approving:
> Testing FPS-independent health icons
I can't get a video right now, but I want to describe this to you
Does the health icon bopping seem to... snap to the wrong place for a single frame each time the camera bops?
I say this because when I watch the bopping on 0.7.4, it doesn't look like that
No matter what framerate I put it on, the health icons do that
What's good is the FPS doesn't change its behavior, but that behavior seems reeeally subtly bugged

PurSnake:
Saving your eyes from infinite frame-snap 
image

LeGooeyy:
its real... i cant believe it...

PR #6038, opened 2025-09-17
[BUGFIX?] Nest song capsule text squash and stretch timers

### LINKED ISSUE #5981: Bug Report: Squashed Text on Freeplay song capsules after selecting a song
### Issue Checklist

- [x] I did not have any mods installed when I encountered this issue

### Platform

iOS/iPadOS

### Browser

None

### Mobile Device Model

iPhone 16 Pro

### Mobile OS Version

iOS 18.6.2

### Version

0.7.4

#2659 again. That's all pretty much it. I don't know if some code was slightly modified in 0.7.4, but it worked fine in 0.7.3. 

Image

### Steps to Reproduce

I don't know, it is really inconsistent, when this happened to me, I had just finished playing Ugh on hard, and switched to Pico, and clicked random. So, no steps really, it's like a 1/100 chance.

Lasercar:
## Linked Issues
Fixes https://github.com/FunkinCrew/Funkin/issues/5981 (?)

## Description
My intial tests of this seem to suggest that this change works, as I was unable to encounter the squashed text at all, though I'd appreciate if other contributors test this change to prove or disprove this.

I've also restored the function call that makes it do the squash and stretch when the capsules refresh, just fyi.

### DIFF
--- a/source/funkin/ui/freeplay/SongMenuItem.hx
+++ b/source/funkin/ui/freeplay/SongMenuItem.hx
@@ -526,11 +526,10 @@ class SongMenuItem extends FlxSpriteGroup
     {
       songText.scale.x = 0.4;
       songText.scale.y = 1.4;
-    });
-
-    new FlxTimer().start(2 / 24, function(_)
-    {
-      songText.scale.x = songText.scale.y = 1;
+      new FlxTimer().start(2 / 24, function(_)
+      {
+        songText.scale.x = songText.scale.y = 1;
+      });
     });
   }
 
@@ -540,6 +539,8 @@ class SongMenuItem extends FlxSpriteGroup
     {
       spr.visible = value;
     }
+
+    textAppear();
 
     updateSelected();
   }


Trayfellow:
If Hungrec or anyone is reading this, the best way to test for fragile stuff like FlxTimers is by bringing FNF to a really bad framerate or a rly high framerate, freezing the game by holding the window bar is also a good way to test this

also, I feel like we can bring back this stretching also happening when you make the capsules do the squish animation when changing difficulties or loading into Freeplay, the text squishing bug that happened during those moments was caused by this FlxTimer function thingy not working properly sometimes

nesting it was prob the best move, but maybe everytime the timer happens, we should also have a function before it that cancels any existing FlxTimers the song text might've possibly had incase the user finds a way to spam the timer starting? idk i dont know anything about flixel there are 2763 ppl here that know better than me lol

Hundrec:
> I've also restored the function call that makes it do the squash and stretch when the capsules refresh, just fyi.

If I understand this correctly, wouldn't this bring back the possibility of the permanently squashed text bug we dealt with for about a year?

Lasercar:
> > I've also restored the function call that makes it do the squash and stretch when the capsules refresh, just fyi.
> 
> If I understand this correctly, wouldn't this bring back the possibility of the permanently squashed text bug we dealt with for about a year?

Yeah, also makes this easier to test as to whether this YOLO change actually does anything or not.

And this works, then great, the cool text animation on the flying capsules is back without the annoying bug, too!

Trayfellow:
> > I've also restored the function call that makes it do the squash and stretch when the capsules refresh, just fyi.
> 
> If I understand this correctly, wouldn't this bring back the possibility of the permanently squashed text bug we dealt with for about a year?

Yeah that would, but maybe my suggestions above could possibly be a good start to fixing that ever actually happening

Trayfellow:
> > > I've also restored the function call that makes it do the squash and stretch when the capsules refresh, just fyi.
> > 
> > 
> > If I understand this correctly, wouldn't this bring back the possibility of the permanently squashed text bug we dealt with for about a year?
> 
> Yeah that would, but maybe my suggestions above could possibly be a good start to fixing that ever actually happening

(especially cause if the user gets those flxtimers spammed, those are alot of timer chains being made, there should be something that stops the chain from building up)

Hundrec (approved):
My man @NotHyper-474 said this one is good

PR #6049, opened 2025-09-18
[OPTIMIZATION] Refactor: cleaner and faster time formatter

HeroEyad:
## Description

**What changed?**

- replaced repeated Math.pow(10, decimals) calls with a single precomputed factor.

- simplified the seconds formatting branch into a concise ternary expression.

i made this for **performance-wise** because `Math.pow` is relatively expensive compared to arithmetic. The old code recomputed it for every unit, even though it only matters for seconds. now it's computed once.

small gain but measurable in tight loops or frequent calls.

The ternary `(u.name == "second" ? ... : ...)` makes it clear that seconds are special-cased without spreading logic across multiple branches.

i've also change variable names (u instead of unit)
precomputed factor makes decimals handling self-contained and predictable.

### DIFF
--- a/source/Postbuild.hx
+++ b/source/Postbuild.hx
@@ -35,37 +35,28 @@ class Postbuild
 
   static function format(time:Float, decimals:Int = 1):String
   {
-    var units = [
-      {name: "day", secs: 86400},
-      {name: "hour", secs: 3600},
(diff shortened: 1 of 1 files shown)


Hundrec:
Excellent! How's this @MAJigsaw77?

PR #6104, opened 2025-09-23
[ENHANCEMENT] Restore strumline fade-out tween when transitioning to Results screen

TechnikTil:
## Description
When digging around the source code, I found a fade out tween that was supposed to play when transitioning into the results screen. This PR adds the tween back with some tiny changes, because why not.

## Video
https://github.com/user-attachments/assets/cdb8aa82-da15-4707-84fa-38cdf6dbeba8

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -3631,6 +3631,9 @@ class PlayState extends MusicBeatSubState
             });
          */
     });
+
+    playerStrumline.fadeOutArrows();
+    opponentStrumline.fadeOutArrows();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Hundrec:
Hey, this is cool!
I received a good number of notes:
* The arrows should fade out in place, not going upward.
* The arrows should all fade out at once, rather than disappearing in a wave.
* The arrows should fade more slowly. 
* The arrows should fade to be mostly transparent, but not fully gone. If it disappeared, it might be too eye-catching and distracting.

This would probably require a totally different implementation, but you are welcome to make those changes on this PR for me to get more feedback on.

Hundrec:
Gonna close this for now, since the scrapped behavior is probably too far from what Dave wants now.

PR #6110, opened 2025-09-23
[BUGFIX] Force the game to properly close instead of becoming a background task

Burokode:
No

I added a simple yet safe way to close the game properly because I remember checking my task manager one day and seeing like 5 copies of Funkin'

I probably made the new branch very poorly so there's like 5 commits when there should be one. But the actual fix is on the "BG Task Fix" commit lol


AbnormalPoof:
Cleaned up the commit history here :>

KaiCodesLol:
Thank you Abnormal I apologize for my dogshit Github usage

Hundrec:
Is there a reliable way to make Funkin' become a background task?
For example, is there some type of audio I should have playing when I close the game?

KaiCodesLol:
I don't think so. I just remember when playing around with the game sometimes (No mods dw) I would check my task manager a few days later and find like 5 copies of Funkin' each taking up 5% of my cpu

EliteMasterEric:
What's the latest version of the game that becomes a background task? This seems like a bandaid fix to a different underlying issue.

KaiCodesLol:
Latest I have found is 0.7.5, the whole reason I did this was because I had found copies of 0.7.5 in my task manager about 2-3 days prior to the PR.

I'm decently sure this is an issue with HaxeFlixel itself because I've had this issue with multiple of my own Flixel projects. So unless you guys want to test this out and try to fix it with your own flixel fork I feel like this is the best fix to the issue as of the current point.

Hundrec:
So it turns out this was an issue with OpenAL that got fixed recently
When we update our libraries (lime), this should be fixed for good!

PR #6118, opened 2025-09-24
[BUGFIX] Fix Chart/Stage Editor backup window retrieving incorrect file 

Lasercar:
## Description
A simple solution to avoid a issue where the it selects the first backup sorted alphabetically rather than the latest one by date.


Lasercar:
I've just come up with a proper solution.

PR #6119, opened 2025-09-24
[ENHANCEMENT/BUGFIX] Fix Chart/Stage Editor backup window retrieving incorrect file

Lasercar:
## Linked Issues
No / Answer

## Description
Get the last modified (and valid) editor backup rather than the one that came alphabetically first.

Also fixes a regression caused by removing the `autosave()` call in `quitChartEditor()` in https://github.com/FunkinCrew/Funkin/pull/5091

## Screenshots/Videos

Screenshot 2025-09-25 212603
Screenshot 2025-09-25 212957

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -697,7 +697,7 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
   var wasCursorOverHaxeUI:Bool = false;
 
   /**
-   * Set by ChartEditorDialogHandler, used to prevent background interaction while the dialog is open.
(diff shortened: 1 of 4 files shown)


Hundrec (changes requested):
This still doesn't work
Now it's just picking a random backup to load, or none at all
It's almost never the right one

Hundrec (changes requested):
This picked up the latest backup I had on the first launch, and then never opened the backup window again no matter how many times I exited without saving a modified chart.

Lasercar:
One more thing, add `&& !FlxG.keys.justPressed.F4` to the check in `quitChartEditor` to also make it autosave when using the main menu escape with unsaved changes.

I'd add it myself but I'm guessing this is too far into the review process to change now.

Hundrec:
Don't worry, you can make the change here and I'll pull it in

PR #6122, opened 2025-09-24
[ENHANCEMENT] Add back the FlxTransitions

KoloInDaCrib:
## Linked Issues
trayfellow's endless comments about it being omitted

## Description
Previously, every transition aside from the state enter was omitted. This PR brings them back, while also accounting for custom transitions like the Sticker Substate being open.
One thing I noticed is that if you're in a substate and switching to another menu, `onStateChangeBegin` actually gets triggered twice. I decided not to touch it, since it's out of scope for this PR but still worth mentioning because I am unsure if it's an oversight or intended behavior.

## Screenshots/Videos

https://github.com/user-attachments/assets/32148f70-4741-4e08-a383-cd013965a2e7

### DIFF
--- a/source/funkin/play/PauseSubState.hx
+++ b/source/funkin/play/PauseSubState.hx
@@ -262,7 +262,7 @@ class PauseSubState extends MusicBeatSubState
 
     regenerateMenu();
 
-    transitionIn();
+    transitionSpritesIn();
 
     startCharterTimer();
   }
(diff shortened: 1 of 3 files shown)


EliteMasterEric:
> One thing I noticed is that if you're in a substate and switching to another menu, onStateChangeBegin actually gets triggered twice. I decided not to touch it, since it's out of scope for this PR but still worth mentioning because I am unsure if it's an oversight or intended behavior.

I know the cause for this (since I've been tinkering with FlxTransition internally). It's because when `FlxTransitionableState` performs a transition, it creates an `FlxTransition` and loads it as a substate, in order to play the transition.

The solution is probably to just exclude states which extend `FlxTransition` from `onStateChange` events, since we can assume they are part of a transition and therefore not fully functional.

Issue #6129, opened 2025-09-25
Enhancement: Hide debug cursor when playtesting from the Chart Editor


minimehan:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [ ] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

When you test your chart in the editor, the mouse is still chillin there on your screen. I think it should be removed during testing just like in normal gameplay.

PR #6137, opened 2025-09-25
[ENHANCEMENT] Hide Cursor When Playtesting Chart

VirtuGuy:
## Linked Issues
Implements #6129

## Description
When using the chart editor, playtesting your chart will keep the cursor visible. This PR changes it so that the cursor hides when playtesting the chart and has it appear again when reopening the chart editor.

## Screenshots/Videos

https://github.com/user-attachments/assets/1a5e4c95-2933-49e1-b1d0-e2be6f13ea15

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -3516,7 +3516,7 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
 
     // Show the mouse cursor.
     // Just throwing this somewhere convenient and infrequently called because sometimes Flixel's debug thing hides the cursor.
-    Cursor.show();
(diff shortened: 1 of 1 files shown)


Hundrec:
I haven't tested this yet, but what happens if you exit the playtest through unconventional methods? (F4 or F5 come to mind)

realvirtu:
I haven't tested it with F4 or F5 just yet that but I imagine it keeps the cursor hidden. I don't think there's a reason for it to show up when those keys are pressed.

realvirtu:
Ok yeah I was right. F4 keeps the mouse hidden but hot reloading does reshow the mouse as it causes the player to reenter the editor. So everything works as it should.

realvirtu:
But thanks Hundrec for thinking about that.

Hundrec:
Thanks for testing!

PR #6155, opened 2025-09-27
[FEATURE] Osu!Mania Importing

FuroYT:
## Associated Funkin.assets PR
https://github.com/FunkinCrew/funkin.assets/pull/268

## Description
Adds the posibility to import osu!mania beatmaps into the chart editor

## Screenshots/Videos

https://github.com/user-attachments/assets/0393b09b-c4f5-4f67-9884-98b3b21d0f8b


MAZ12211:
OSU before step mania is wild but big claps anyway 👏👏👏

Kade-github:
would love to get this in as a start to importing other formats, will be reviewing internally. ty!

Kade-github:
4k charts do seem to work though

image
image

albeit putting notes on both sides (which probably should just be bf?)

FuroYT:
> 
> 
> image
> 
> 
> 
> image
> 
> 
> 
> 
> 
> does not work with 8k charts (too many notes)

Hi Kade,
Thanks for pointing that out I will fix it as soon as possible while making 8k charts not be flipped with the 2 strumlines

FuroYT:
> 4k charts do seem to work though
> 
> 
> 
> image
> 
> image
> 
> 
> 
> albeit putting notes on both sides (which probably should just be bf?)

The fact that it doesn't work well on 8K is because it copies to chart into the opponent Strumline because you may need a osu chart to be same for the player and opponent since I've seen lots of osu charters needing that so I added it by default

Kade-github (approved):
yup that works

Issue #5436, opened 2025-07-21
Bug Report: Stickers break on sides with widescreen


LeGooeyy:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Windows (Downloadable Build)

### Version

0.7.2

### Description (include any images, videos, errors, or crash logs)

https://github.com/user-attachments/assets/fead3203-3645-4ca4-aced-0b9a5b6c2bdd

The stickers on the sides of the screen dont fully work and line up incorrectly

PR #6156, opened 2025-09-27
[ENHANCEMENT] Render Transition Stickers on OpenFL Sprite

PurSnake:
## Linked Issues
#5436 

## Description
Now transition stickers are rendered on their own OpenFL container sprite, to render them on top of the entire game, as originally intended (right, Cam?)

Preview:

https://github.com/user-attachments/assets/ddbf938d-7e1d-49f2-8b10-a4f802d13713

### DIFF
--- a/source/funkin/ui/transition/stickers/StickerSubState.hx
+++ b/source/funkin/ui/transition/stickers/StickerSubState.hx
@@ -3,6 +3,7 @@ package funkin.ui.transition.stickers;
 import flixel.addons.transition.FlxTransitionableState;
 import flixel.FlxG;
 import flixel.FlxState;
+import flixel.FlxCamera;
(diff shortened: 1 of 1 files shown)


TechnikTil (changes requested):
There's a lot of formatting changes made, so I don't think The Funkin' Crew will like this PR.

PurSnake:
Hmm

ninjamuffin99:
If I were to redo this myself I would prob do it as a Flixel plugin, since we have those configured to render on-top of the game (see the in-game screenshot stuff) so that we can simply use Flixel/FlxSprite API rather than working with OpenFL

However this is a good PR thanks! We are just testing it internally to make sure it works as intended on all platforms!

Hundrec:
npesta-sped-up

PR #6171, opened 2025-09-29
[BUGFIX] Fix Crash When Using Static Pixel Icon In Character Select

VirtuGuy:
## Linked Issues

Issues Smissues.

## Description

Hm... well that's a lie.

image

This pr fixes a null object reference crash that occurs when using a static image as a pixel icon for the character select screen. This is not a problem for Freeplay.

## Screenshots/Videos

Below is a screenshot of the infamous null object reference crash that occurs in this particular screen for the particular reason stated above. How was this not reported (well at least I don't think it was)? This is an old issue I believe. Well... at least this pr fixes the crash.

image


AbnormalPoof:
lmao

Trayfellow:
Oh my god bruh

Issue #4810, opened 2025-04-26
Bug Report: Animation Editor shows a HaxeFlixel icon


Honton129:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.3

### Description (include any images, videos, errors, or crash logs)

When you change from "Spritesheet" to "Animation" when entering the menu for the first time, on the "Animation List", a HaxeFlixel icon appears on top.

Selecting or hovering the mouse on the dropdown solves the issue but it reappears everytime you re-enter the menu and repeat the steps to do it.

https://github.com/user-attachments/assets/c02e11f0-bca1-4093-86e7-b772752ea477

### Steps to Reproduce

1. Enter the Animation Editor.
2. Change to "Animations" mode.
3. Icon shows up on "Animations List" dropdown.

PSI3x:
For some reason, 0.6.4 still has this issue.

PR #6177, opened 2025-09-30
[BUGFIX] Fix the missing object icon spawing when switching from Spritesheet to Animations

KoloInDaCrib:
## Linked Issues
Closes https://github.com/FunkinCrew/Funkin/issues/4810

## Description
When changing to the animations mode in the animation editor, a lone haxeflixel icon will spawn on top of the animation dropdown. This is because the haxe ui dropdown doesnt get initialized properly when its visibility is set to false. This PR fixes this by making the animation dropdown use `hide()` and `show()` as opposed to toggling the visibility variable.

## Screenshots/Videos

https://github.com/user-attachments/assets/b6d9e191-aaa5-4dcb-8513-d9e7ba301c89


Trayfellow:
True...

PR #6181, opened 2025-09-30
[BUGFIX] Make MathUtil Lerp Functions not FPS-dependent

FuroYT:
## Description

Fixes the issue that `MathUtil::smoothLerpPrecision` and `MathUtil::smoothLerpDecay` could be fps dependant sometimes.
This makes these function really adapt to `deltaTime` so they show up the same in every framerate

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1149,11 +1149,13 @@ class PlayState extends MusicBeatSubState
     // Apply camera zoom + multipliers.
     if (subState == null && cameraZoomRate > 0.0) // && !isInCutscene)
     {
-      cameraBopMultiplier = FlxMath.lerp(1.0, cameraBopMultiplier, 0.95); // Lerp bop multiplier back to 1.0x
+      var lerpBumpRatio = 0.95;
+      lerpBumpRatio = 1.0 - Math.pow(lerpBumpRatio, FlxG.elapsed * 60); //Adapt it to the current fps
+      cameraBopMultiplier = FlxMath.lerp(cameraBopMultiplier, 1.0, lerpBumpRatio); // Lerp bop multiplier back to 1.0x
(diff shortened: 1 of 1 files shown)


Hundrec:
This might be the one that does it

Trayfellow:
Chat is this true

DustedDustin:
this fixes the lerping being strange for scoring at high framerates? (I assume it does for other lerping issues, but that was the one I noticed the most when playing at 500fps and the score not smoothly lerping)

lunacynne:
unfortunately, i think this pr (or at least the `MathUtil` change) is redundant
the current functions are already framerate independent

testing both the existing functions and the ones from this pr with different framerates (plus some random varying deltas just to make sure) doesnt show any real difference

image

they all still correctly converge to the function path with higher framerates, but the changed functions just do it in extra steps

image

FuroYT:
> unfortunately, i think this pr is redundant
> 
> the current functions are already framerate independent
> 
> 
> 
> testing both the existing functions and the ones from this pr with different framerates (plus some random varying deltas just to make sure) doesnt show any real difference
> 
> 
> 
> image
> 
> 
> 
> they all still correctly converge to the function path with higher framerates, but this pr just does it in extra steps
> 
> 
> 
> image
> 
> 

I've had multiple times where the function ran without fps independance and it wasn't that rare and I since don't have the problem

lunacynne:
> I've had multiple times where the function ran without fps independance and it wasn't that rare and I since don't have the problem

wherever this is occurring, im confident the issue lies in the implementation and not the lerp functions themselves then

Hundrec:
Fixed with https://github.com/FunkinCrew/Funkin/commit/6a71f95cf99964c25e07fa4ce1e3d1b8a0b3e427, which looks like it works.
If there's anything up with it, feel free to open a new PR!

PR #6211, opened 2025-10-04
[BUGFIX] Fix event dropdown field values passing to the next selected event 

NotHyper-474:
## Linked Issues
No one dared making a report on this yet...

## Description
Each time you select an event, the event toolbox is refreshed, and in turn it updates the values for each field with the values in `eventDataToPlace`, which is used in this case to store the newly selected event's data. However in doing so it triggers each field's `onChange` event, which will also take `eventDataToPlace` and apply it to the currently selected event.

The issue is that `onChange` will take the current value of the dropdown – which sometimes corresponds to the prior event's data – and apply that to `eventDataToPlace`, which causes the "persistence" you see.

So... this fixes the issue by simply preventing the fields' `onChange` to be dispatched when the toolbox is refreshed.

## Screenshots/Videos
[and i'm not gonna stop till i forget what we had.webm](https://github.com/user-attachments/assets/42147c22-a5a4-4b1e-98a3-7ce366c1d0cf)


Hundrec:
Really?? I really hope this one works

NotHyper-474:
> Really?? I really hope this one works

PR #6222, opened 2025-10-05
[ENHANCEMENT] Make Hot Reload key bindable

Starexify:
## Linked Issues
*-*

## Description
This PR makes the Hot Reload key configurable through the Controls menu, allowing players to rebind it to their preferred key instead of it being hardcoded to `F5`.

## Screenshots/Videos

image


EliteMasterEric:
Hot Reload is a DEBUG feature it should not be rebindable

PR #6225, opened 2025-10-06
[BUGFIX] Round `NumberPreferenceItem` floating-point values to prevent near-zero precision errors

### LINKED ISSUE #3888: Bug Report: createPrefItemNumber using a Float is complicated and buggy
### Issue Checklist

### Platform

### Browser

None

### Version

0.5.2

I try to make an custom Preference Option with Float instead of Int, and when I tried to go from 0.3 or higher down to 0, it returns an odd number.

This is my code:
```haxe
preferences.createPrefItemNumber("HUD Opacity", "Changes the opacity of the HUD", function (value:Float) {
save.data.hudOpacity = value; },
function (value:Float):String {
Std.parseFloat(save.data.hudOpacity); }, save.data.hudOpacity, 0.0, 1.0, 0.1, 1);
```

And this is the method that I use that is from the base game:
```haxe
/**
* Creates a pref item that works with general numbers
* @param onChange Gets called every time the player changes the value; use this to apply the value
* @param valueFormatter Will get called every time the game needs to display the float value; use this to change how the displayed value looks
* @param defaultValue The value that is loaded in when the pref item is created (usually your Preferences.settingVariable)
* @param min Minimum value (example: 0)
* @param max Maximum value (example: 10)
* @param step The value to increment/decrement by (default = 0.1)
* @param precision Rounds decimals up to a `precision` amount of digits (ex: 4 -> 0.1234, 2 -> 0.12)
*/
  function createPrefItemNumber(prefName:String, prefDesc:String, onChange:Float->Void, ?valueFormatter:Float->String, defaultValue:Int, min:Int, max:Int,
      step:Float = 0.1, precision:Int):Void
  {
    var item = new NumberPreferenceItem(0, (120 * items.length) + 30, prefName, defaultValue, min, max, step, precision, onChange, valueFormatter);
    items.addItem(prefName, item);
    preferenceItems.add(item.lefthandText);
  }
```

image
As you see in the image it shows basically 2.777777e^-17 etc, even tho the precision is set to 1 decimal after .

Also extra question, is there any way to set a float instead of int for the value and also save it even if you refresh the game and work properly?

Thanks for your help

### Steps to Reproduce

1. Try to create a createPrefItemNumber with Floats
2. Shows a very low number but not 0

Starexify:
## Linked Issues
Fixes #3888 

## Description
This PR fixes an issue where `NumberPreferenceItem` floating-point values could produce near-zero precision errors (e.g. 2e-17 or 1e-16).

Here is the script for testing:
```haxe
class Testing extends ScriptedModule {
  override function onStateChangeEnd(event:StateChangeScriptEvent) {
    if (!(event.targetState is OptionsState)) return;

    var preferences:PreferencesMenu = OptionsState.instance.optionsCodex.pages.get("preferences");
    if (preferences == null) return;

    preferences.createPrefItemNumber("Test - Number", "", value -> trace(value), null, 0,  -1.0, 1.0, 0.1, 1);
  }
}
```

## Screenshots/Videos

Before:

https://github.com/user-attachments/assets/cf3e66b9-2b22-44ae-98f6-d7074f6e83e4

After:

https://github.com/user-attachments/assets/658c12fc-dfd2-48c2-b6dc-2681ee92a77e


AbnormalPoof (approved):
very nice

PR #6240, opened 2025-10-08
[ENHANCEMENT/FIX] Sort notes by length

gamerbross:
## Linked Issues
Couldn't find any, but there could be.

## Description
After the sustain notes rework, note hits just play the sing animation once, so if there is a double note, one with length 0 and another one being sustained, depending on the note list sorting for automatic characters or hit time for players, it can look weird to just sing the short note instead of the long one.

## Screenshots/Videos
_Before (the chart has to be done WITHOUT the fix, also you need to add the short note AFTER the long ones)_

https://github.com/user-attachments/assets/4116803b-0fc5-4dd2-8b6e-8ef899809f6b

---
_After, with the same chart_

https://github.com/user-attachments/assets/7f39849b-fc68-490b-b69c-0f1db4b19ba5

### DIFF
--- a/source/funkin/data/song/SongDataUtils.hx
+++ b/source/funkin/data/song/SongDataUtils.hx
@@ -5,6 +5,7 @@ import funkin.data.song.SongData.SongEventData;
 import funkin.data.song.SongData.SongNoteData;
 import funkin.data.song.SongData.SongTimeChange;
 import funkin.util.ClipboardUtil;
+import funkin.util.SortUtil;
 
 using Lambda;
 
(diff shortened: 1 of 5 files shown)


gamerbross:
this doesn't really matter anymore because it's possible to manually fix it by setting `noanim` to the less relevant note, but maybe adding this anyway isn't bad ¯\\\_(ツ)_/¯

Hundrec:
I think this would still be nice to have, but it needs some conflict resolution first

gamerbross:
oops

gamerbross:
fixed 👍
code is slightly different but now doesn't have any breaking changes

Issue #6244, opened 2025-10-09
Enhancement: Enable back button in Character Select on mobile


JackXson-Real:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

0.7.5 added the ability to exit out of the character select menu by clicking backspace. This currently isn’t possible on mobile due to the back button being disabled in the menu. The button should be enabled so mobile users can take advantage of the new QOL feature.

ender69flock:
Hmmm, i was thinking about it

PR #6252, opened 2025-10-10
[ENHANCEMENT] Add Mobile Back Button to Character Select

VirtuGuy:
## Linked Issues
Implements #6244

## Description
This PR adds the back button to the character select for mobile. Pressing the back button on mobile does the exact same thing on desktop builds when you press back, the back button ~~hides~~ and becomes disabled when selecting a character and it will not reenable until you deselect that character.

## Screenshots/Videos

Sorry for the bad video quality, I had just figured out how to record my phone screen from my computer.

https://github.com/user-attachments/assets/29f0cc91-79f5-4b83-ab11-edf33c2373e6

Edit: I decided that the back button should stay visible after selecting a character. I changed this so that the back button would be consistent with the behavior of any other menu. If a developer wants this reverted or changed at all, I will do so. I also fixed a problem with this pr where the button wouldn't hide when going back to freeplay after selecting a character.

### DIFF
--- a/source/funkin/ui/charSelect/CharSelectSubState.hx
+++ b/source/funkin/ui/charSelect/CharSelectSubState.hx
@@ -729,6 +729,13 @@ class CharSelectSubState extends MusicBeatSubState
     allowInput = false;
     autoFollow = false;
 
+    #if mobile
+    if (backButton != null)
+    {
+      FlxTween.tween(backButton, {alpha: 0}, 0.4, {ease: FlxEase.quadOut});
+    }
+    #end
+
     FlxTween.tween(cursor, {alpha: 0}, 0.8, {ease: FlxEase.expoOut});
     FlxTween.tween(cursorBlue, {alpha: 0}, 0.8, {ease: FlxEase.expoOut});
     FlxTween.tween(cursorDarkBlue, {alpha: 0}, 0.8, {ease: FlxEase.expoOut});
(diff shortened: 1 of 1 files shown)


CEliuxJV:
I like this, but maybe the back button shouldn't disappear when you select a character?

FuroYT:
> I like this, but maybe the back button shouldn't disappear when you select a character?

It should because on the pc version when you press back it cancels the character but for mobile you just press the screen again to cancel the selection
So it is valid to hide the back button

CEliuxJV:
> > I like this, but maybe the back button shouldn't disappear when you select a character?
> 
> It should because on the pc version when you press back it cancels the character but for mobile you just press the screen again to cancel the selection So it is valid to hide the back button

I know I can touch anywhere on the screen to cancel the selection, but when I played the mobile port for the first time I didn't know that that's the way to see the funny animations for the characters.

Maybe It will be more intuitive for the new people to let them know that they can press back to cancel the selection and see a funny animation.

comatoasterstrudel:
I also agree, it would be more clear to have the back button cancel the character select instead of tapping anywhere on the screen in my opinion.

Hundrec:
We just merged a slightly different internal implementation of this. If there are any tweaks you think we should make, feel free to rebase and update this branch with those adjustments!

PR #6253, opened 2025-10-10
[ENHANCEMENT] Add back the "Video file does not exist" error

TechnikTil:
## Description
Inside of `funkin.play.cutscene.VideoCutscene.play`, there was going to be an error for when a video is missing. This was scrapped because HTML5 would always report it as missing.

After debugging on HTML5 for way too long, I figured out that the OpenFL library was never loaded into the game, causing the `exists` lookup to completely skip the library.

This PR adds back the error message and loads the video library in `LoadingState`, so that the `Assets.exists` function can be accurate.


AbnormalPoof (approved):
very awesome and cool

PR #6295, opened 2025-10-17
[OPTIMIZATION] Unify player/opponent voice list builders into one function using `CharacterType`

Starexify:
## Description
This PR unifies the functions `buildOpponentVoiceList` and `buildPlayerVoiceList` into a single `buildVoiceListForCharacter` function, which uses `CharacterType` to determine which vocal list to build.

Also removes the `$suffix` from the second checking, which prevented normal vocals from characters to be searched for, such as `Vocals-bf` if there is no `Vocals-bf-car-variation` or `Vocals-bf-variation`


AbnormalPoof:
FYI this is not a chore since it makes a functional change

PR #6298, opened 2025-10-17
[ENHANCEMENT] Add support for opening folders on Linux

Starexify:
## Description
This PR adds support for opening folders on Linux systems.
It first attempts to use the standard `xdg-open`  command
```
xdg-open <path>
```

If the command fails (exit code ≠ 0), it falls back to trying a list of common Linux file managers:
`dolphin, nautilus, nemo, thunar, caja, konqueror, spacefm, pcmanfm`

For `openSelectFile`, it currently falls back to `openFolder` since, based on testing, Linux File Managers don't have support for opening a directory with a specific file selected.

## Screenshots/Videos

https://github.com/user-attachments/assets/29cbebc0-ef5b-45e7-9629-eec3a91095a7

### DIFF
--- a/source/funkin/util/FileUtil.hx
+++ b/source/funkin/util/FileUtil.hx
@@ -1180,8 +1180,18 @@ class FileUtil
     // thats why the above comment is there!
     Sys.command('open', [pathFolder]);
     #elseif linux
-    // TODO: implement linux
-    // some shit with xdg-open :thinking: emoji...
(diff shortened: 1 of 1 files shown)


AbnormalPoof:
Looks like something for Kade to look at, will see what he thinks.

Kade-github:
~~This looks fine rn as a basic fix, but maybe the best fix would be a lime pr that adds a function to use std::filesystem (which handles os stuff)~~ ehh scratch that it looks like this is the only way.

Issue #2943, opened 2024-07-01
Enhancement: ability to set a static scroll speed across all charts


platonictf:
currently, scroll speed is set arbitrarily by whoever charted the song you are currently playing. this leads to various playability issues, such as scroll speeds varying wildly between songs/difficulties, which can throw off players not expecting it as well as newbies who cant play with a fast scroll speed. 
it also makes most songs, particularly ones with slow scroll speed generally uncomfortable to read for experienced players who are used to reading much faster scroll speeds, usually resulting in them hitting notes very early and/or misreading patterns entirely that they otherwise wouldnt have.

allowing the player to set a single scroll speed for every song would solve all these issues, as newbies won't be caught off guard with a scroll speed that is way too fast, as it now remains the same across all songs, and experienced players can use whatever speed is most comfortable for them to read.

minimehan:
Thank goodness. I thought about it as well when playing classic songs, but wasn't sure whether it's intentional or not.

PR #6300, opened 2025-10-18
[ENHANCEMENT] Ability to set the Scroll Speed with modes and change it dynamically during PlayState

Starexify:
## Linked Issues
#2943 

## Description
This PR adds a way for the player to set their own scroll speed when playing songs.

Includes:
- A Preference Option `Scroll Speed` with 3 Modes:
  - Off - No Scroll Speed Change from player applied
  - Static - The scroll speed is fixed based on the player's scroll speed
  - Adaptive - The scroll speed is set based on the player's scroll speed and can be altered by song events, (ex. `ScrollSpeedEvent`)

- 2 new Controls with their own header for changing the scroll speed mid-song
  - Increase - Increases the strumlines and player's scroll speed by 0.1
  - Decrease - Decreases the strumlines and player's scroll speed by 0.1
The change will affect differently the strumlines from the players based on the selected mode, `Static` will always be same for the strumlines/song/player scroll speed, instead `Adaptive` will change the current speed of the strumlines, meaning song events can alter it.

## Screenshots/Videos
Example:

https://github.com/user-attachments/assets/9e78e643-c3d2-4fc2-b6b6-930eec3652c3

https://github.com/user-attachments/assets/efbc4a22-e8d9-43ac-90ba-a30fb780c2c3

Preference:

image

Controls:

image

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -351,15 +351,8 @@ class Preferences
 
   static function set_scrollSpeedMode(value:ScrollSpeedMode):ScrollSpeedMode
   {
-    var result:String = switch (value)
-    {
-      case ScrollSpeedMode.STATIC: "Static";
-      case ScrollSpeedMode.ADAPTIVE: "Adaptive";
-      default: "Off";
-    };
-
     var save:Save = Save.instance;
-    save.options.scrollSpeedMode = result;
+    save.options.scrollSpeedMode = value.getName();
     save.flush();
     return value;
   }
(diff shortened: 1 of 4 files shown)


Starexify:
Since a new Pause Menu is on the way I won't continue this. I will just leave it as a concept if it's considered to be added in other updates 🫡

Hundrec:
We'd implement this differently if we decide to, so I'll close this. Cool concept nonetheless!

PR #6317, opened 2025-10-19
[MODDING] Use `Strumline.KEY_COUNT` instead of hardcoded 4-key logic for direction handling

Starexify:
## Description
This PR only replaces the hardcoded `= 4` parameter with `Strumline.KEY_COUNT`, allowing mods with extra key to handle the extra key animations and directions correctly.

**This doesn't change any gameplay in base game !**


Hundrec:
Eric says we will keep the key count hardcoded.
If you want more keys, you'll have to make your own engine.

PR #6320, opened 2025-10-19
[ENHANCEMENT] Extend fullscreen aspect ratio

dotevie:
## Description
I noticed some pillarboxing running the normal build on my phone, and I figured out this is because of `FullScreenScaleMode`'s aspect ratio maximum of 20:9. This is a bit below quite a few Android phones, like the Pixel 9/10 (20.5:9), all Galaxy Z Flip phones (21-21.5:9) and the outer screen of the 2 newest Z Fold phones (22:9). This PR extends the maximum aspect ratio to the widest available Android phone, which I believe to be the Z Fold 7's outer screen at 22:9.

While this does mostly just work, there are a few edge cases I saw where some things aren't far enough to extend to this ratio, being in weeks 2 and 7, and I'm sure there's also some I missed. I'm leaving this PR as a draft for now because I'm not sure how to handle these, and I would imagine there are preferred ways to do this, so I wouldn't want to do it with no other input. If anyone has any ideas for how to handle this I would very much appreciate it!

Screenshots below are of the game at 22:9.

## Screenshots
image

image

image

image

image

image

### DIFF
--- a/source/funkin/ui/FullScreenScaleMode.hx
+++ b/source/funkin/ui/FullScreenScaleMode.hx
@@ -35,7 +35,7 @@ class FullScreenScaleMode extends flixel.system.scaleModes.BaseScaleMode
   /**
    * The maximum aspect ratio a screen can have.
    */
-  public static var maxAspectRatio:FlxPoint = FlxPoint.get(20, 9);
+  public static var maxAspectRatio:FlxPoint = FlxPoint.get(22, 9);
 
   /**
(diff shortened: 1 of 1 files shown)


PurSnake:
At the moment, the ideal maximum aspect ratio should be 21:9

sector-a:
some bg assets just aren't long enough for 22:9, so i don't think its ok

Hundrec:
Cool concept, but this would require a lot of extra work.

PR #6335, opened 2025-10-22
[BUGFIX] Fix icons snapping away for one frame on song restart

NebulaStellaNova:
## Description
Fixed health icons position not setting when restarting song.

## Screenshots/Videos
Bug I Fixed:

https://github.com/user-attachments/assets/0d47b35c-213c-4376-a901-6e424154e4a8


charlesisfeline:
here before it merges

PR #6338, opened 2025-10-23
[OPTIMIZATION?] Don't process killed notes

NotHyper-474:
## Description
A tiny optimization I guess? Skips calculating all the hit window stuff for notes that were killed.

## Screenshots/Videos
Not applicable since the gameplay doesn't change... and probably not performance either.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2700,7 +2700,7 @@ class PlayState extends MusicBeatSubState
     // Process notes on the opponent's side.
     for (note in opponentStrumline.notes.members)
     {
-      if (note == null) continue;
+      if (note == null || !note.alive) continue;
       var r = GRhythmUtil.processWindow(note, false);
       if (r.botplayHit)
       {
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
Seems like a good change, IDK if there are any possible side effects from it but killed notes should be just skipped. Probably a teeny performance improvement if anything.

NotHyper-474:
>Seems like a good change, IDK if there are any possible side effects from it but killed notes should be just skipped.

I had tested this change a while ago and AFAIK the only thing it affected was that time-traveling forward would cause less damage, but well, it doesn't cause any now because of a recent change.

EliteMasterEric:
> I had tested this change a while ago and AFAIK the only thing it affected was that time-traveling forward would cause less damage, but well, it doesn't cause any now because of a recent change.

Time-travel should not cause damage ideally (it's a debug feature, and the intent is that any notes between the current time and the target time are treated as though they didn't exist), but was that change intentional or a side effect of another change?

NotHyper-474:
Yeah it was intentional. And to be more clear the "cause less damage" I described was how it used to behave, but after what I believe was the note recycling fix it started to always deal damage whenever you time traveled, so I think this fixed, or rather, would also fix it, since that change was a different fix for the same problem.

PR #6344, opened 2025-10-24
Fix wrong checks for NaN throughout the code

NotHyper-474:
## Description
Fixes wrong checks for NaN, the `Math.isNaN` function should be used instead of comparing the values because it's always gonna be `False...`

https://try.haxe.org/#6ea88699
^ You can see how `Math.isNaN` returns the correct result here


FuroYT:
LGTM!

PR #5123, opened 2025-05-21
[ENHANCEMENT] WIP: Animation Priority system

AbnormalPoof:
## Linked Issues
Fixes #3491, fixes #2367 (for real this time).

## Description
This PR adds a new animation priority system, replacing `ignoreExclusionPref`.
`priority` acts like `zIndex` for characters, where the higher the value is, the higher the animation is prioritized!

## TODO
- [x] Animate Atlas character support
- [x] Stage prop support
- [ ] Default priorities for `sing` and `idle`/`dance`
- [ ] Set priorities for existing animations
- [ ] Figure out how to do forced animations

PR #6370, opened 2025-10-28
[ENHANCEMENT] Implement animation priority

KoloInDaCrib:
## Linked Issues
True... or False... call it.

## Description
Characters had a very weird implementation of animation priority. Singing was set to have the highest priority, despite not even being set to ignore other animations, overriding any other animation that may play instead. It was a mess.
This PR introduces a json-configurable animation priority system, which should hopefully make the animations a lot more consistent. The new system supports adding asterisks to animations, in case multiple animations should have the same priority. Any animation that isn't in the configurable array assumes the index of -1 and therefore has the highest priority. The default animation priority queue is `["sing*", "idle"]` or `["sing*", "dance*"]`, meaning that sing and miss animations should happen before the idling animation(s).

> [!CAUTION]
> This is a breaking change for mods using `ignoreExclusionPref`, as well as for the base game scripts using it! I've opened a PR in the assets repo (https://github.com/FunkinCrew/funkin.assets/pull/295) that should be merged as well so that there's no script errors being thrown on the screen for every character initialization.

## Screenshots/Videos

https://github.com/user-attachments/assets/fc007340-6d34-4630-a975-d9084fc9d64e

### DIFF
--- a/source/funkin/data/character/CharacterData.hx
+++ b/source/funkin/data/character/CharacterData.hx
@@ -523,6 +523,12 @@ class CharacterDataParser
       input.startingAnimation = DEFAULT_STARTINGANIM;
     }
 
+    if (input.animPriorityQueue == null)
+    {
+      var canDance:Bool = [for (a in input.animations) a.name].contains("danceLeft");
+      input.animPriorityQueue = ["sing*", (canDance ? "dance*" : "idle")];
+    }
+
     if (input.scale == null)
     {
       input.scale = DEFAULT_SCALE;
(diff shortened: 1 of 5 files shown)


EliteMasterEric:
This implementation is kind of odd, and doesn't work with other types of sprites.

I was imagining that the `playAnimation` function, instead of having a `force` argument, would have a numeric priority with a default value, with that value determining whether one animation can override another.

CrusherNotDrip:
> This implementation is kind of odd, and doesn't work with other types of sprites.
> 
> I was imagining that the `playAnimation` function, instead of having a `force` argument, would have a numeric priority with a default value, with that value determining whether one animation can override another.

Hi #5123

TechnikTil:
You might wanna make fixing these merge conflicts a priority!

NotHyper-474:
noooooooooo

PR #6382, opened 2025-10-30
[BUGFIX] Fix animation editor frame outlines being improperly generated for multisparrow characters

KoloInDaCrib:
## Linked Issues
None that I can think of.

## Description
The animation editor uses the character's frames to generate the outline data. This works great most of the time, however for multisparrow characters, it can lead to generating the frame outlines for frames that aren't on the main spritesheet, considering the frames get appended to the character's frame collection.
The PR fixes it by forcefully creating a new frame collection instance for multisparrow characters (since Flixel reuses the frame collection instances, even if it has appended frames from other frame collections).

The memory usage shouldn't in theory be horrible, however if you find a better solution, let me know.

## Screenshots/Videos
Pico (Playable) Before:
image

Pico (Playable) After:
image

### DIFF
--- a/source/funkin/play/character/MultiSparrowCharacter.hx
+++ b/source/funkin/play/character/MultiSparrowCharacter.hx
@@ -2,6 +2,7 @@ package funkin.play.character;
 
 import flixel.graphics.frames.FlxAtlasFrames;
 import flixel.graphics.frames.FlxFramesCollection;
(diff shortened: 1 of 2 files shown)


AbnormalPoof (changes requested):
One note when reviewing the code

TechnikTil:
Merge Conflict Alert!

PR #6394, opened 2025-11-02
[BUGFIX] Fix Old Icon Being Flipped Incorrectly For Flipped Icons

VirtuGuy:
## Linked Issues
None

## Description
In the game, having it so that the player health icon is flipped makes it so that the old icon is flipped incorrectly. This PR (of course) fixes this by setting `flipX` to true for the old icon (because player icons are normally flipped). Yes, the icon does properly go back to how it was before if you were to change the icon back.

I probably made this sound confusing, but to help explain what I mean, characters can have `healthIcon` data where you can have `flipX` set to true, but for the player, having it set to true really means false.

Screenshot 2025-11-02 153534

Screenshot 2025-11-02 154048

## Screenshots/Videos

FNF if it was good... or if you're playing the game with this PR:
Screenshot 2025-11-02 153514

My organs failed trying to explain this. "Flipped" was used way too much in the making of this PR.

### DIFF
--- a/source/funkin/play/components/HealthIcon.hx
+++ b/source/funkin/play/components/HealthIcon.hx
@@ -171,6 +171,7 @@ class HealthIcon extends FunkinSprite
     {
       characterId = 'bf-old';
       isPixel = false;
+      flipX = true;
       loadCharacter(characterId);
     }
 


realvirtu:
I believe that flip is not the only thing that needs to fixed for the old icon. That would make a good free PR for someone.

realvirtu:
Closing as #6720 deals with everything that has to do with the inconsistencies of the old icon.

PR #6406, opened 2025-11-07
[ENHANCEMENT] ModOptions

VirtuGuy:
## Description

Adding your own custom preferences to the game requires quite a bit of work and knowledge, but what if... there was a class made just for creating preferences without having to actually mess with the preferences menu or all the checks involved in adding the preference menu items?

This pr adds a new class called `ModOptions.hx`. This new class comes with the ability to easily create four different types of preference options: checkbox, number, percentage, and enum. Not only can you easily create these with this class, but they automatically get saved to the save's `modOptions`!

These modded options have a parameter called `defaultValue`. This is of course the fallback for if the option wasn't saved ever, or if it had been saved as a different type of option (this means that your game won't crash if something like a checkbox tries to use a value that's a string or whatever).

Below is an example of how YOU can make a checkbox... as long as your testing the pr.

image

The parameters for this are the unique id, name, description, default value, whether the option is available, and a callback that runs when the option's value gets changed.

Obviously there's a bit more to this than being able to create a checkbox, but I just wanted to give an example on making a checkbox with this class because it would be the most basic type of option to create.

Just know that these options are **not** sorted in the order they're registered, but rather it's based on the alphabetical order of the unique ids.

## Screenshots/Videos

image

### DIFF
--- a/source/funkin/modding/ModOptions.hx
+++ b/source/funkin/modding/ModOptions.hx
@@ -1,0 +1,251 @@
+package funkin.modding;
+
+import haxe.ds.StringMap;
+import funkin.save.Save;
+import funkin.util.SortUtil;
+
+/**
+ * The params for a mod option.
+ */
+typedef ModOptionParams =
+{
+  type:ModOptionType,
+  name:String,
+  desc:String,
+
+  // Checkbox
+  ?available:Bool,
+
+  // Number
+  ?min:Float,
+  ?max:Float,
+  ?step:Float,
+  ?precision:Int,
+  ?valueFormatter:Float->String,
+
+  // Enum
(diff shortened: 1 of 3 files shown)


realvirtu:
Sometime I'll have to change how the preferences are actually loaded in. Yes it works fine, but checkboxes always load before anything else and I would also like to sort them in alphabetical order.

realvirtu:
> Sometime I'll have to change how the preferences are actually loaded in. Yes it works fine, but checkboxes always load before anything else and I would also like to sort them in alphabetical order.

Done and done.

realvirtu:
Gonna close this because although it's a really nice PR, pretty sure Kade said that this might not get added anyways.

PR #6414, opened 2025-11-09
[FEATURE] Macro to detect if a chart has been changed on Runtime

KoloInDaCrib:
## Linked Issues
None.

## Description
This PR introduces a macro that stores the chart text on compile-time to be later used to check if a chart file has been tampered with through mods that use `_merge` folder or override the chart file directly or through overriding a chart through the assets folder, effectively making cheating for scores more difficult.
Note that cheated scores are still possible by using some scripts to change the tallies however this PR makes cheating not baby-level.
Also the macro class is inside the `funkin/util/macro` folder so it's inaccessible to scripts.

Constants didn't seem to work during compilation so I couldn't use `Constants.DEFAULT_VARIATION` in the macro code, in case you're curious.

## Screenshots/Videos

https://github.com/user-attachments/assets/1cb9a2a4-204b-47dd-b50d-8eba74f4f47c

### DIFF
--- a/project.hxp
+++ b/project.hxp
@@ -1138,6 +1138,9 @@ class Project extends HXProject
     // These macros allow additions of new functionality to existing Flixel. -->
     addHaxeMacro("addMetadata('@:build(funkin.util.macro.FlxMacro.buildFlxBasic())', 'flixel.FlxBasic')");
     addHaxeMacro("addMetadata('@:build(funkin.util.macro.FlxMacro.buildFlxSprite())', 'flixel.FlxSprite')");
+
+    // This macro will go over every song in the assets folder and store them in an array to check for cheated scores.
+    addHaxeMacro("funkin.util.macro.SongDataValidator.loadSongData()");
   }
 
   function configureOutputDir()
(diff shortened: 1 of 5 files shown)


AbnormalPoof:
> Why not store a hash of the chart text instead
> MD5/SHA1

- Eric

EliteMasterEric:
Implementation looks good in general, although we should probably add something to make sure you can't modify the chart data map via a script.

EDIT: Actually all that should require is blacklisting the SongDataValidator in the PolymodHandler

KoloInDaCrib:
> EDIT: Actually all that should require is blacklisting the SongDataValidator in the PolymodHandler

The class itself is in the funkin/util/macro folder whose classes already are blacklisted due to the CompiledClassList exploit which had been patched in https://github.com/FunkinCrew/Funkin/pull/5185

mikolka9144:
Is the no in-game note about "The score won't be saved" intentional?

I can already see someone complaining about the game not saving score because it just didn't ???

Raltyro (approved):
I love the branch named "superman-image"

Image

TechnikTil:
Merge Conflicts!

Hundrec:
Would you like to bring this change to `preview/public-playtest`?

PR #6422, opened 2025-11-12
[BUGFIX] Fix Missing Album Title Asset Showing Flixel Logo

VirtuGuy:
## Description

When having an album that doesn't have a title asset, the Flixel logo appears and this PR is good enough to solve that problem. Just know that the logo disappears when reselecting the song after selecting a different song with a different album, but that shouldn't matter because like I said, this PR prevents the logo from appearing.

Also, this is not a problem with the actual album art as albums have a fallback for when its art is missing.

## Screenshots/Videos

Before:
screenshot-2025-11-11-20-10-12

After:
screenshot-2025-11-11-20-17-05

### DIFF
--- a/source/funkin/ui/freeplay/AlbumRoll.hx
+++ b/source/funkin/ui/freeplay/AlbumRoll.hx
@@ -183,7 +183,7 @@ class AlbumRoll extends FlxSpriteGroup
 
   public function showTitle():Void
   {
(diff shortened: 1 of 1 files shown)


realvirtu:
Thank you github-actions for adding the `size: tiny` label, then removing it afterwards.

PR #6424, opened 2025-11-12
[ENHANCEMENT] Provide Import Aliases for Typedefs

KoloInDaCrib:
## Linked Issues
No, sire, none at all.

## Description
This PR adds more functionality to ClassMacro - using it to list all typedefs and store them and their underlying class into a map to then be added as an import alias for HScript.
This works for both typedefs that point to a class (such as FlxSpriteGroup) and typedefs that point to an abstract of a class (such as FlxSignal).

If you're wondering why I'm comparing string versions of parameters instead of their regular enum selves, refer to this screenshot: 
image
As you can see, Type.enumEq returns false despite the two parameters being literally the exact same. Macros are infuriating.

## Screenshots/Videos
image

### DIFF
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -21,6 +21,7 @@ import funkin.util.FileUtil;
 import funkin.util.macro.ClassMacro;
 import polymod.backends.PolymodAssets.PolymodAssetType;
 import polymod.format.ParseRules.TextFileFormat;
+import polymod.hscript._internal.PolymodScriptClass;
 import polymod.Polymod;
 
 /**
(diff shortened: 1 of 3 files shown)


lemz1:
> As you can see, Type.enumEq returns false despite the two parameters being literally the exact same. Macros are infuriating.

The reason Type.enumEq returns false is because we are comparing arrays.
even though the arguments in TFun of params and checkParams are both an empty array, they dont live at the same memory address.
in other words, Type.enumEq does not work when the enum has arguments of type array, because it compares the references of the arrays not the actual values inside the array.

you can see what i mean here: [Try Haxe Code](https://try.haxe.org/#413FB618)

Maybe we should create a helper function that also works with enums that have arguments of type array?

KoloInDaCrib:
Importing typedefs was added to thePolymodrepo through another PR, so I'll be closing this.

Issue #6436, opened 2025-11-17
Enhancement: Adjust position of Chart Editor notification


Msguro:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

The notifications in the Chart Editor refer to windows that display messages like “Loaded song” or “Auto-Save.”
These windows appear over the playback bar, which can be quite obstructive when you want to jump to the end of a song. 
It would be much more convenient if the windows appeared in a blank area instead.

**current**

<img width="1919" height="1009" alt="Image" src="https://github.com/user-attachments/assets/58cfd42e-972f-4da9-9def-b85f37374381" />

**Improvement proposal**

<img width="1919" height="1009" alt="Image" src="https://github.com/user-attachments/assets/d1ce1480-d2e5-4c78-8786-a3022f2826a6" />

Trofem:
good suggestion, i tried do it once, but it was hard to do

Hundrec:
Good suggestion!
The current position of the notification is actually an improvement from the original position, which I made a few patches ago.

That was the best I could do given the way the notification's position is handled in the code, so if anyone wants to make this better, it would require some kind of refactor.

Anyone is welcome to take a shot at it!

PR #6445, opened 2025-11-20
[ENHANCEMENT] Move the Chart Editor Notifications Up a Bit

MightyTheArmiddilo:
## Linked Issues
#6436 

## Description
Now the notifications have been moved up! It does this by setting `NotificationManager`'s `AnimateFromBottom` field to a cutsom version that doesn't use the `GUTTER_SIZE` variable.

There's probably a better place to put the redefintion of `AnimateFromBottom` and might be a bit overboard with an entire redefine, but oh well, it works!

> [!NOTE]
> For some reason, the first notification to ever pop up is always off by about 10, but that was a thing before this PR.

## Screenshots/Videos

https://github.com/user-attachments/assets/9b873172-0a7d-43d1-8cd2-359d65de201e


Msguro:
This is incredibly beautiful…

PR #6471, opened 2025-11-22
[OPTIMIZATION?] Copy cache maps instead of assigning

KoloInDaCrib:
## Linked Issues
N/A

## Description
Unlike other basic types, arrays and maps aren't automatically copied when you asign them to a variable.

image
image

This means that in `FunkinMemory`, there are cases where permament cache maps get fully cleared, which hurts performance since the sounds/bitmaps would have to get cached again.
I can't say how much this affects performance, as this doesn't affect my computer that heavily, though what I can say is that with this the memory usage won't build up over time the more songs are played in a session, since the cache is cleared after exiting results. From my testing I've had memory usage roll back from 1gb to 0.75gb after exiting a song.

## Screenshots/Videos
N/A

### DIFF
--- a/source/funkin/FunkinMemory.hx
+++ b/source/funkin/FunkinMemory.hx
@@ -170,7 +170,7 @@ class FunkinMemory
     graphic.persist = true;
     permanentCachedTextures.set(key, graphic);
     forceRender(graphic);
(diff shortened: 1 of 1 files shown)


NotHyper-474 (approved):
So I'm not the only one who noticed this issue

PR #6476, opened 2025-11-23
[BUGFIX] Fix Results Screen Song Text Being Framerate Dependent

VirtuGuy:
## Description

This PR fixes an issue with the results screen where the song text scrolls at a speed that isn't framerate independent. This means that the text will scroll faster at a higher framerate and slower and a lower framerate, when really it should be the same scroll speed.

## Screenshots/Videos

Below are videos of what this problem looks like at 60 FPS and 200 FPS. Luckily, this PR fixes this problem!

https://github.com/user-attachments/assets/09018fff-7992-4b24-bc28-0053986d2062

https://github.com/user-attachments/assets/033ba37b-96b0-4e3c-9ee2-045c5aca3533

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -811,6 +811,10 @@ class PlayState extends MusicBeatSubState
     // This state receives draw calls even when a substate is active.
     this.persistentDraw = true;
 
+    // Make the player unable to pause if they're moving from the chart editor while the focus is still on since the input persists.
+    @:privateAccess
+    justUnpaused = isChartingMode && !FlxG.game._lostFocus;
+
     // Stop any pre-existing music.
     if (!overrideMusic)
     {


Trayfellow:
Elapsetic

NotHyper-474:
> The fix works but I have no idea what the logic behind multiplying the value by 60 is.

If I were to guess it's as if it kept the speed it moved at 60 FPS.

If we suppose the current FPS is 60 and, knowing that an elapsed frame time is technically `1 / FPS`, the equation would evaluate to `speed * (1/60) * 60` which would just become `speed` after cancelling out the values.

Doubling the FPS it'd become `speed * (1/120) * 60 -> speed * (1/2)` where then as a compensation it would move at half the speed per frame, which would also match what it would look like at 60 FPS.

realvirtu:
> If I were to guess it's as if it kept the speed it moved at 60 FPS.

Exactly right! If it wasn't multiplied by 60, the text would move much slower than it should be.

PR #6501, opened 2025-11-27
[ENHANCEMENT] Better backwards compatibility with `FlxAtlasSprite`

KoloInDaCrib:
## Linked Issues
None.

## Description
Due to the switch to flixel-animate deleting `FlxAtlasSprite`, an import alias that points to `FunkinSprite` was created. This, however, doesn't do the best job since FlxAtlasSprite's constructor function had a parameter for the atlas path, meaning that mods that used FlxAtlasSprite before would still break :ohmygodbruh:
This PR adds the missing arguments from FlxAtlasSprite's constructor to FunkinSprite's constructor, i.e. the path and the atlas settings, both of which being optional. I also made it so that if you provide a path ending with a .png in the constructor, it would load the graphic as if you called `.loadGraphic()`.

## Screenshots/Videos
N/A

### DIFF
--- a/source/funkin/graphics/FunkinSprite.hx
+++ b/source/funkin/graphics/FunkinSprite.hx
@@ -21,6 +21,7 @@ import animate.internal.elements.AtlasInstance;
 import animate.internal.elements.SymbolInstance;
 import animate.FlxAnimate;
 import animate.FlxAnimateFrames;
+import haxe.io.Path;
 
 using StringTools;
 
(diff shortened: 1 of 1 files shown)


KoloInDaCrib:
ive just now realized that mods that used FlxAtlasSprite would still break even with this because they used some functions from the Bopper class like playAnimation aaaaaaaaaaaaaaaaaaaaaa

would moving the functions like those from the Bopper class to the FunkinSprite be appropriate for this PR? Don't think it would break anything since Boppers extend FunkinSprite.

AbnormalPoof:
> ive just now realized that mods that used FlxAtlasSprite would still break even with this because they used some functions from the Bopper class like playAnimation aaaaaaaaaaaaaaaaaaaaaa
> 
> 
> 
> would moving the functions like those from the Bopper class to the FunkinSprite be appropriate for this PR? Don't think it would break anything since Boppers extend FunkinSprite.

That's basically a given with the switch. I'd rather have people replace those functions in their scripts.

PR #6502, opened 2025-11-28
[BUGFIX] Dispatch `onOpening` signal to actually set video cutscene volume

PurSnake:
## Linked Issues
Not actually from issue tab, but from some dude from discord

https://github.com/user-attachments/assets/db1ea886-a312-4f15-ba6d-aecba1f88c1e

## Description
Dispatches `onOpening` signal to set video's volume in moment it started playing (not in the next frame)

## Screenshots/Videos
Video's just starts now with correct volume :D

### DIFF
--- a/source/funkin/play/cutscene/VideoCutscene.hx
+++ b/source/funkin/play/cutscene/VideoCutscene.hx
@@ -198,6 +198,7 @@ class VideoCutscene
       if (vid.load(filePath, fileOptions) && vid.play())
       {
         onVideoStarted.dispatch();
+        vid.bitmap.onOpening.dispatch();
       }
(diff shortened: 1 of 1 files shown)


MAJigsaw77:
I think making a fix in `hxvlc` would actually be a better idea tbh.

PurSnake:
> I think making a fix in `hxvlc` would actually be a better idea tbh.

true

PurSnake:
Fixed in [new hxvlc commit](https://github.com/MAJigsaw77/hxvlc/commit/eaa2b1177d2086954cc86bf625e41c2b0ac6fbfc)

PR #6506, opened 2025-11-29
[BUGFIX] Fix FunkinAction Only Running checkPressed() Once.

VirtuGuy:
## Description

This PR fixes an issue with a recent develop branch commit where FunkinAction's checkPressed() would only run once until the player would release the key. This causes an issue with the character select menu where you have to release the key in order for the selection to change (because that's just how character select had been coded).

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -111,7 +111,7 @@
       "name": "hxcpp",
       "type": "git",
       "dir": null,
-      "ref": "6546fa5c3ad1bac065f144745122ab5a6d4195ff",
+      "ref": "5932340d095a7eea8635fe4d1355f1c0efd0b3c2",
       "url": "https://github.com/FunkinCrew/hxcpp"
     },
     {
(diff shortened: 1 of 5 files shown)


realvirtu:
I suppose I'll include this amazing bonus from when I was trying to fix this. Don't worry, it's not like this at all. This video's only here to show you something funny that luckily didn't make it into the PR.

https://github.com/user-attachments/assets/b85b5ec1-7b9f-4ce3-b782-b43076901c75

I'm sure all the Discord fellows are familiar with this as I already showed this not too long ago.

yeeterongithub:
we love simple fixes...

Hundrec:
Cam made a whole batch of fixes, including this one.
Thank you for providing a solution in the meantime!

yeeterongithub:
> Cam made a whole batch of fixes, including this one.
> 
> Thank you for providing a solution in the meantime!

wasn't cam the one that broke it 😭

PR #6528, opened 2025-12-04
[BUGFIX] Fix NOR when deleting save data

NotHyper-474:
## Linked Issues
... There's nothing ...

## Description
This one is quite interesting because at first it doesn't even make any sense until you discover it's related to SaveProperties.

Let's keep in mind `this.data` and `data` are different variables. The issue is that due to the equal name, the `data` argument is incorrectly referenced by the "body" of the properties (this becomes evident in the generated code) which is null when deleting the data. And so only `this.data` gets set, the other one will stay null, which causes the error.
image

The fix I used was making the argument also be set, which isn't an issue because it's passed by reference, which means that changes to `data` will also apply to `this.data`.
There was an other fix which involved forcing the macro to reference `this.data` but also required disabling null-safety for each property (either manually or automatically by adding the meta using the macro).

## Screenshots/Videos
Screenshot of the crash
Screenshot of the crash

### DIFF
--- a/source/funkin/save/Save.hx
+++ b/source/funkin/save/Save.hx
@@ -67,7 +67,7 @@ class Save implements ConsoleClass
   @:nullSafety(Off)
   public function new(?data:RawSaveData)
   {
-    this.data = data ?? Save.getDefaultData();
+    this.data = data ??= Save.getDefaultData();
     // Build macro will inject SaveProperty initialization here automatically
 
     // Make sure the verison number is up to date before we flush.


EliteMasterEric:
The fix works, but I don't think it's a good idea to try to be "clever" with syntax like this.

I'll keep your commit since you did the work of finding the problem, but I'm rewriting the code to remove the variable shadowing, fixing the issue while also keeping code clear and preventing future rewrites from causing more issues.

```haxe
@:nullSafety(Off)
public function new(?rawData:RawSaveData)
{
    data = rawData ?? Save.getDefaultData();
    // ...
}
```

EliteMasterEric (approved):

PR #6532, opened 2025-12-04
[BUGFIX] Make Rimlight Shader Better For Texture Atlases

VirtuGuy:
## Description

With characters being converted to texture atlases, there comes a problem where some characters have to be skipped due to the rimlight shader that's used in Week 7 Erect songs.

Well what's the problem? Uh um uh uh ummm uhh...

screenshot-2025-12-04-14-56-56

Yeah that's right.. the one and only piss colors. By the way, I'm using Boyfriend's car variant for this as it is the closest thing we got for a texture atlas BF so far.

So anyways, instead of using the frame's UV for applying the shader each animation frame, frameWidth and frameHeight are used instead.

Now this isn't perfect, but it's better than before.

screenshot-2025-12-04-15-31-57

However, using an alt mask helps. Setting the mask threshold to 1, you'll get a much better result. I'm not going to provide the alt mask due to it being a character that isn't used for Week 7 Erect, but I made it to see how these types of problems can be solved.

screenshot-2025-12-04-16-19-13

I know this is out of FunkinCrew's control, but if MaybeMaru/flixel-animate#15 gets merged, the alt mask likely won't be needed for texture atlases. Yes, this PR would still be needed.

Thanks to Abnormal for figuring out this problem and explaining a way to fix it. Really all I'm doing is making the PR.

!! FunkinCrew/funkin.assets#304 is required !!

Edit: MaybeMaru/flixel-animate#15 has been merged and now alt masks won't be needed!

### DIFF
--- a/source/funkin/graphics/shaders/DropShadowShader.hx
+++ b/source/funkin/graphics/shaders/DropShadowShader.hx
@@ -192,7 +192,7 @@ class DropShadowShader extends FlxShader
   function set_attachedSprite(spr:FlxSprite):FlxSprite
   {
     attachedSprite = spr;
-    updateFrameInfo(attachedSprite.frame);
+    updateFrameInfo(attachedSprite.frameWidth, attachedSprite.frameHeight, attachedSprite.frame.angle);
     return spr;
   }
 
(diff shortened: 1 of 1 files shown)


CheeseWithCake:
bro wait that rim looks sick though
like it doesnt look as flat

TechnikTil:
wouldn’t this break rotated frames?

realvirtu:
> wouldn’t this break rotated frames?

Well idk. Is there a reason why you think it would?

MaybeMaru:
> > wouldn’t this break rotated frames?
> 
> Well idk. Is there a reason why you think it would?

the reason why the shader took the frame data was because rotated frames need an extra rimlight angle offset to look correct, make sure to test with normal sprites too

realvirtu:
> the reason why the shader took the frame data was because rotated frames need an extra rimlight angle offset to look correct, make sure to test with normal sprites too

I did test other songs with regular characters, which it seemed fine to me, but now I realize that it's a problem. I'll have to update the asset pr to have this fixed.

MaybeMaru:
id recomend making it a draft while you do that just in case

realvirtu:
> id recomend making it a draft while you do that just in case

How come?

MaybeMaru:
Just so it's not in a confusing state of "ready to merge" but still a wip

realvirtu:
Alright. Rotated frames seem fine now.

AbnormalPoof:
Already did this internally w/ render texture, soggy!

PR #6540, opened 2025-12-05
[OPTIMIZATION] Use streamed audio for Freeplay Song Previews

KoloInDaCrib:
## Linked Issues
Nothing, heh, it's like that one soyjak gif.

## Description
This PR makes it so that on targets that have the `lime_vorbis` feature flag enabled, freeplay previews will attempt to be loaded through the vorbis file first before using the regular partial sound method.

I'd love to make more things use streamed audio for optimization, though there are some roadblocks to this:
1. Bugs, like a lot of them. From my testing, using too much streamed audio can sometimes cause the game to remove every audio, making the game silent, which I hope is an incredibly obvious problem for a rhythm game. The test was basically putting the vorbis loading method in FunkinSound's `load` and `loadPartial` functions, which also affected sound effects played through the `playOnce` function.
2. Nene and her malice break the game, since streamed audio doesn't have any data to use for the A-Bot visualizer. ACrazyTown has made a PR on the funkVis repository https://github.com/FunkinCrew/funkVis/pull/13 to fix the issue but it's been sitting there since april of this year.

Also if the code looks a bit ugly, that's because I had to apply `endTime` to the loaded sound so that it's actually a partial sound in a very specific position, since fading in the sound calls `play` without any arguments, which means that `endTime` is set to null and it makes the previews not partial anymore, spoiling the entire instrumental in the preview.............

## Screenshots/Videos

https://github.com/user-attachments/assets/a25e4b13-5d63-4a6b-931b-577b4f775171

### DIFF
--- a/source/funkin/audio/FunkinSound.hx
+++ b/source/funkin/audio/FunkinSound.hx
@@ -17,6 +17,9 @@ import lime.media.AudioSource;
 import openfl.events.Event;
 import openfl.media.SoundChannel;
 import openfl.media.SoundMixer;
+#if lime_vorbis
+import lime.media.vorbis.VorbisFile;
+#end
 
 /**
  * A FlxSound which adds additional functionality:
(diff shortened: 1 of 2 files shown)


ACrazyTown (changes requested):

TechnikTil:
Merge Conflicts!

PR #6545, opened 2025-12-07
[BUGFIX] Fix Alpha Not Being Applied To Animation Editor Onion Skin Upon Character Change

VirtuGuy:
## Description

This PR fixes an issue with the current version of the develop branch where the animation editor's onion skin doesn't have its alpha applied upon changing the character. It would instead only apply it when the character's animation gets changed.

image

I bet you can't tell which one is the real Girlfriend (Christmas).


realvirtu:
Doesn't happen anymore for some reason.

realvirtu:
Nevermind. I was flipping the character and I forgot that refreshes the onion skin.

PR #6547, opened 2025-12-07
[BUGFIX] Fix Chart Editor Difficulty Toolbox Displaying Any Variation Song Name

VirtuGuy:
## Description

This PR fixes a problem where the Chart Editor's difficulty toolbox displays the song name of the variation that was first loaded. This means that if you were to load a song such as Lit Up (BF Mix), the song name shown in the toolbox will be that instead of the song's default song name.

## Screenshots/Videos

Before:

image

After:

image


EliteMasterEric (approved):
Fixes the bug

PR #6557, opened 2025-12-10
[EXPERIMENTAL - BUGFIX] Prevent the song from progressing on lagspikes.

KarimAkra:
So this PR adds a new option called `"Frame Synced Music"`
It sort of makes the music time in PlayState based on frame progression so a lagspike should not caused the song to drift off and giving unfair penalties.

More details:
The music time is already sort of frame based before this PR.  But whenever the music shifts off from the conductor due to most likely a lagspike it just moves ahead the conductor to the music time (this logic is still there if the new option is disabled)
With the new music, the actual music time will be snapped back to the conductor time whenever it drifts away from what is most likely a lagspike
The song would still go forward during the hang,  but it'll snap back afterwards with the notes still inplace

We thought some people may not like this system so that's why it's an option

This has been made a public PR because conductor changes always seemed to have been sensitive and we don't want to accidentally cause public builds to be buggy like 0.5 again do we?

Anyways testing and stress testing on this will be much appreciated!


DustedDustin:
PEAK???? Huge for mobile or weaker devices if this works smoothly

source/funkin/play/PlayState.hx
```diff
@@ -1178,8 +1178,16 @@ class PlayState extends MusicBeatSubState
         {
           // Fallback to properly update the conductor incase the lerp messed up
           // Shouldn't be fallen back to unless you're lagging alot
-          trace(' WARNING '.bg_yellow().bold() + ' Normal Conductor Update!! are you lagging?');
-          Conductor.instance.update();
+          if (Preferences.frameSyncedMusic)
+          {
+            trace(' WARNING '.bg_yellow().bold() + ' Reseting song time to the conductor time!! are you lagging?');
+            FlxG.sound.music.time = Conductor.instance.songPosition;
```
KoloInDaCrib: Wouldn't this have a chance of offsetting the vocals, since you're only updating the time of the inst? I recommend calling `resyncVocals()` after this just to be sure

DustedDustin:
I'll start casually playing with it to test for bugs once someone compiles a build for android

DustedDustin:
After some playtesting, I can say it is nice but I see why it's an option. There's a lot of points where I'd rather have it off because the correction throws me off too much and inadvertently messes with the flow my inputs.

NotHyper-474:
There's a bug with it causing the music to constantly rewind to the same point on high offsets after a stutter whattheshit

Should be an easy fix since you just need to remove the offset when setting the music time

https://github.com/user-attachments/assets/0b2ad899-fba9-4e25-a0bc-98a90a89e532

KarimAkra:
> There's a bug with it causing the music to constantly rewind to the same point on high offsets after a stutter 

Should be fixed in the last commit!

TechnikTil:
Any plans on updating this soon? You might have merge conflicts...

KarimAkra:
Maybe one day...

PR #5734, opened 2025-08-12
(HELP WANTED) [BUGFIX] Fix Hit Window Being Offset By FPS

JackXson-Real:
## Linked Issues
Fixes #5667 

## Description
**This is a heavy WIP. This description will be updated with more info and videos once it is in a perfect state. Here are the currently known issues:**
- Sustain trail is affected by the offset causing it to appear above the hold note cover
- Downscroll is currently weirdly affected by the offsets
- There still seems to be some inconsistencies in the hit window, mainly noticeable on higher scroll speeds

## Screenshots/Videos
SoonTM

PR #6563, opened 2025-12-12
[BUGFIX] Hit Window Offset By FPS

### LINKED ISSUE #5667: Bug Report: Lower FPS Values Affect The Hit Window And Judgements/Score
### Issue Checklist

### Platform

Compiled from GitHub Source Code

### Browser

None

### Mobile Device Model

_No response_

### Mobile OS Version

_No response_

### Version

0.7.3 Develop

The lower the FPS is the more offset the hit window is. If you are playing at 60FPS, you can hit the notes WAY too early and still get a sick. I know it's very hard to get stuff like this to function consistently between FPS values, but other rhythm games are able to handle different FPS values, so there has to be a way to fix it. Here are some examples of how the hit window behaves on 60FPS vs 500FPS.
## 500 FPS

https://github.com/user-attachments/assets/a1767ec0-afac-49aa-b8f0-a5af59f0f2c0

## 60 FPS

https://github.com/user-attachments/assets/4b967013-e531-4735-93ac-7d7d1c95f92a

### Steps to Reproduce

1. Lower FPS Cap to a low number
2. Hit notes way before they reach the strumline.

Lasercar:
## Linked Issues
Fixes #5667

## Description

See previous PR for details: https://github.com/FunkinCrew/Funkin/pull/5734

> [!IMPORTANT]
> Requires https://github.com/FunkinCrew/funkin.assets/pull/331

### DIFF
--- a/source/funkin/InitState.hx
+++ b/source/funkin/InitState.hx
@@ -477,7 +477,7 @@ class InitState extends FlxState
    * @param songId
    * @param difficultyId
    */
-  function startSong(songId:String, difficultyId:String = 'normal'):Void
+  function startSong(songId:String = '', difficultyId:String = 'normal'):Void
   {
     var songData:Null<funkin.play.song.Song> = funkin.data.song.SongRegistry.instance.fetchEntry(songId, {variation: Constants.DEFAULT_VARIATION});
 
--- a/source/funkin/data/notestyle/NoteStyleData.hx
+++ b/source/funkin/data/notestyle/NoteStyleData.hx
@@ -244,6 +244,8 @@ typedef NoteStyleData_NoteStrumline =
   var rightPress:UnnamedAnimationData;
   var rightConfirm:UnnamedAnimationData;
   var rightConfirmHold:UnnamedAnimationData;
+  // Special FPS dependent Y offset for the strumline notes, for ensuring the visual hit window matches the calculated one.
+  var FPSYOffset:Float;
 }
 
 typedef NoteStyleData_NoteSplash =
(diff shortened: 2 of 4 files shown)


Kade-github:
Eric apparently has thoughts that he never put on the original, so im assigning him here as well

EliteMasterEric:
Why does this add data to the note style? And the value has two DIFFERENT default values?

I'm really cynical about PRs like this because the potential risks are so dire and the potential upsides seem like at best a placebo.

Lasercar:
> Why does this add data to the note style? And the value has two DIFFERENT default values?
> 
> I'm really cynical about PRs like this because the potential risks are so dire and the potential upsides seem like at best a placebo.

The pixel notestyle was still offset if the value for the default notestyle is used on it. Custom notestyles will probably have the same issue too, so the only solution is to create a value that can be modified until a perfect note hit matches the strumline at all fps caps, as much as I hate it.

For example, on downscroll, the pixel note style was looking like this:

Screenshot_2025-05-11_060706
Screenshot_2025-05-11_060932

While the default style looked like this:

Screenshot_2025-05-11_061156

Also, we still can't get an accurate current fps right now because of lime, so these changes are really all that can be done at the moment until that changes.

JackXson-Real:
> Also, we still can't get an accurate current fps right now because of lime, so these changes are really all that can be done at the moment until that changes

even if we could get accurate current FPS, it sounds like it would cause loads of lag since it would be constantly changing based on the FPS, and that can change a lot in a small amount of time. I’m not entirely sure how lime works though so maybe it wouldn’t cause any lag.

Hundrec:
This one's just too much of a bandaid solution to accept, sorry!
Maybe we'll find a more backend fix for this eventually.

PR #5229, opened 2025-06-13
Custom per-character death names

BubbledDEV:
Closes issue https://github.com/FunkinCrew/Funkin/issues/3759

Allows for players to be given custom death names on the pause screen to match their game over screens. These can be added to any player by adding the variable "deathName" to their .json file.
I personally chose to name the statistic "x Blams" for Pico, but if left unspecified for a player, it will just default to "x Deaths", the latter of which can be changed in PlayerData.hx.

image
image
image

PR #6574, opened 2025-12-17
[Enhancement] Improve death counter in pause menu

### LINKED ISSUE #3759: Enhancement: change pause menu blue balls text depending on the playable character
### Issue Checklist

- [X] I have properly named my enhancement
- [X] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested

as of right now while playing as pico it tells you that you x amount of blue balls even tho pico doesnt have a blue balls gameover
image
it would be cool if its changed to something like "x kniffed" or "x killed" or "x gameovers"

would be cool if there is a variable for it in the playable characters playable character json or its just changed to "x gameovers"

TerraFail451:
## Linked Issues
- #3759

## Description
This was originally just going to be a repeat of #5229 because the PR was stale, but then I had the idea to make the death counter singular when you have one death.

This PR makes it so that if the death count is at one, the death count is singular. If it's at any other value, its plural.

This PR also allows for player characters to be given unique deaths names on the pause menu that match up with their game over animations. This is done through the `deathName` and `deathNameSingular` variables in the player's `.json` file. I provide those files in the funkin.assets PR (FunkinCrew/funkin.assets#316).
I chose the dual variable approach because it's way easier than trying to account for words that end in "s" in the singular form (like loss --> losses) and irregular plurals (like goose --> geese, mouse --> mice, or sheep --> sheep).

btw the default `deathName` is `Deaths` and the default `deathNameSingular` is `Death`. i just gave Pico the defaults because I couldnt think of anything better and he does in fact die so it works. but if you can think of anything better its just two lines in his `.json` file that you have to change.

### Boyfriend with 0 Blue Balls
Boyfriend with 0 Blue Balls

### Boyfriend with 1 Blue Ball
Boyfriend with 1 Blue Ball

### Boyfriend with 2 Blue Balls
Boyfriend with 2 Blue Balls

### Pico with 0 Deaths
Pico with 0 Deaths

### Pico with 1 Death
Pico with 1 Death

### Pico with 2 Deaths
Pico with 2 Deaths


realvirtu:
absolute cinema

JackXson-Real:
Will you be fixing the merge conflicts?

TerraFail451:
yeah I'll do that

TerraFail451:
> yeah I'll do that

its been 2 months and I _haven't_ done that so here's an update:
i made a half-baked attempt at that pretty soon after I said that, but since I was expecting it to be quick and easy to solve the merge conflicts i wasn't ready to wrestle with VS Code. it wouldn't let me change branches until I uploaded the pending changes that magically appeared somehow and i couldn't figure out how to revert them or what i was supposed to do to fix it (i don't use VS Code or GitHub often so I don't really know what im doing and often forget how to do things), so i decided to deal with it later. i then procrastinated for 2 months due to lost interest (I have ADHD) and then school work started piling up and now its testing season

school ends at the end of the month for me, so I'll try to do it then. tests are almost over so I might even do it earlier than that. but i WILL fix those merge conflicts eventually

i also want to try to improve how plural names are handled since I'm not happy with how i did it. im thinking:
- `deathName` would be the singular name
- `deathNamePlural` would be the plural name used to override the `deathName` pluralization function
- `useDeathNamePlural` would be a boolean variable that determines whether to pluralize `deathName` or use `deathNamePlural` instead. as a personal note because I'll probably be referencing this, make sure to set the value to `0` by default
- check if `deathNamePlural` is defined or not. if I can't figure out how to do that without preventing crashes, it will instead check if it's set to some extremely specific default value like `this is the default deathNamePlural value that is used to determine that deathNamePlural should not override the deathName pluralization function`
  - if it's not defined / is set to the default
    - set `useDeathNamePlural` to`0`
  - if it's defined / not set to the default
    - set `useDeathNamePlural` to`1`
- if the death count is `1`
  - display `deathName` with no changes
- if the death count is not `1`
  - if `useDeathNamePlural` is `0`
    - if `deathName` ends in `s`
      - display `deathName` with `es` added to the end
    - if `deathName` does not end in `s`
      - display `deathName` with `s` added to the end
  - if `useDeathNamePlural` is `1`
    - display `deathNamePlural`

this would be good because I don't have to account for irregulars, and only one variable needs to be defined for non-irregulars.
doing the check for `deathNamePlural` beforehand and assigning it to the `useDeathNamePlural` variable is probably also good since it's probably an optimization

you know, typing this up is giving me some newfound motivation, so i might do this sooner than I previously expected. but like who knows when I'll do it. just know that i WILL fix those merge conflicts eventually

Hundrec:
We'll be doing an overhaul of the Pause Menu, so we'll close this.

Thank you for your contribution!

TerraFail451:
sounds good!

Issue #6558, opened 2025-12-11
Enhancement: Freeplay "OFFICIAL OST" text based on album art json file


ChickenSwimmer2020:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

simple idea, the "OFFICIAL OST" text in the top left should be changable depending on if a songs album art file changes the text. maybe somthing like
```json
{
    "OSTText": "something here"
}
```

(if this is an already submitted idea, then i didnt find it, i searched for keywords like: "official ost" and even just: "ost" and didnt find anything)

ZazaFi:
This is a feature i've been thinking about for a while. it would definitely benefit mod makers, especially with making compat less of a hassle between mods, QT Rewired for example overrides any song that is not from the mod with 'Official OST" even if the other mods have their own text, like for character mixes. a change like this would prevent compat issues as it will be a base game system, all the modders would have to do is just update to include the OST text value.

DatGamerYoloYT:
what if this could also be added to the player json data so the default text is always a custom version instead of "OFFICIAL OST" when in a custom character's freeplay

realvirtu:
I'd probably make a PR for this if it hasn't been done already. Makes sense actually. Modded songs aren't part of the official ost.

realvirtu:
> I'd probably make a PR for this if it hasn't been done already. Makes sense actually. Modded songs aren't part of the official ost.

Lmao I should stop saying things when I know I'm too busy to even bother.

Marcosayfi:
isn't the official ost text at the top right? either way, i really like this feature as i was confused why it said official ost even though the song i selected was modded in. maybe if the album doesn't specify the ost text or the song doesn't have an album assigned to it, it can just say "UNOFFICIAL OST" or "MODDED OST" or something like that?

Starexify:
Jarvis, add the `status: pending pull request`

PR #6578, opened 2025-12-18
[ENHANCEMENT] Moddable Freeplay OST text

NotHyper-474:
## Linked Issues
* #6558

## Description
Adds an optional `albumOSTName` property to – you've guessed it – album data, that replaces the "OFFICIAL OST" text in Freeplay.

The behaviour I've defined is like so:
- Base game albums will show "OFFICIAL OST" by default, and that is done automatically so. The JSON can still override it though.
- Modded albums will not display any text by default, but one can still be defined via JSON.

It can be useful because mods can try to override the text themselves but will force the default one when other mods attempt to do the same. This will not fix that issue though, only prevent it in future mods.

## Screenshots/Videos

https://github.com/user-attachments/assets/5f568a15-bb2a-4ca8-bcf2-1ec72d378179

### DIFF
--- a/source/funkin/data/freeplay/album/AlbumData.hx
+++ b/source/funkin/data/freeplay/album/AlbumData.hx
@@ -49,4 +49,10 @@ typedef AlbumData =
   @:optional
   @:default([])
   public var albumTitleAnimations:Array<AnimationData>;
+
+  /**
+   * An optional override for the Freeplay OST text.
+   */
+  @:optional
+  public var albumOSTName:String;
 }
--- a/source/funkin/ui/freeplay/Album.hx
+++ b/source/funkin/ui/freeplay/Album.hx
@@ -1,7 +1,6 @@
 package funkin.ui.freeplay;
 
 import funkin.data.freeplay.album.AlbumData;
-import funkin.data.freeplay.album.AlbumRegistry;
 import funkin.data.animation.AnimationData;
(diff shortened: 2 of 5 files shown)


realvirtu:
Absolute cinema.

lasystuff:
~~I'm not sure but can't we just call it "soundtrackType" or something?
Is it just me or "OSTName" sounds a bit weird.~~

(IM SORRY NEVERMIND THAT WAS ALREADY NAME OF THE TEXT INSTANCE)

NotHyper-474:
Considering Eric's response in https://github.com/FunkinCrew/Funkin/pull/6623#issuecomment-3707768104 and that this technically can also be easily done with Hscript I think this might be rejected too.

Well, considering there doesn't seem to be a way of doing that without other mods messing with it, we'll see how it turns out.

Raltyro:
Pretty sure it wont going to be rejected, the fact that there will be more mods that will overlaps changing this,
and it makes sense for it to be changeable through metadatas too

Raltyro (approved):

Marcosayfi:
i honestly thunk it's getting added in one of the next updates.

Marcosayfi:
what if the album doesn't specify the ost text or the song doesn't have an album assigned to it, it can just say "UNOFFICIAL OST" or "MODDED OST" or something like that?

PR #6580, opened 2025-12-19
[BUGFIX / ENHANCEMENT] Health Bar Layering Fix + Smoother Health Bar

ghost:
Fixed the layering of the health bar and health bar background, is not janky af anymore.
Also set `numDivisions` to make the health bar smoother.

image

https://github.com/user-attachments/assets/fdca0f66-f5f2-447e-9867-7d1645c1fc83


MaybeMaru:
darken blend is broken outside of desktop, use multiply instead, or no blends at all
this is a weird way of going around it

ghost:
ok noted

AbnormalPoof (commented):
What issue is this supposed to fix? Is there a before and after of the changes?

ghost:
The health bar is supposed to be stylized slightly, with the way it's currently implemented there is no reason to have the bg be a separate image. This fixes that and properly implements the health bar texture.

Before
image
After
image

MightyTheArmiddilo:
Why DOES there need to be a separate sprite for the background? The FlxBar class has borders built in, and borders are really the only thing I can see the bg sprite having a use for

ghost:
It's stylized

lunacynne:
this may be a stupid question but why not just have three 2x2 color sprites (`FunkinSprite.makeSolidColor`) and scale/position them like

red/opponent:
- scaled to target width/height of health bar

black/border:
- below red z
- red position - padding
- scaled to red width/height + 2x padding

green/player:
- above red z
- aligned to right side of red, scales to left side as health approaches max
- scaled to red height

it should use less memory than `FlxBar` because the sprite textures would be only 48b total instead of several kb
kinda insignificant yea BUT it being scale based means it would be super smooth on any resolution instead of quantized

NotHyper-474:
This now has conflicts, please update your PR in 90 days or it will be closed as stale.

PR #6584, opened 2025-12-20
[BUGFIX] Fix Welcome Music Playing After Unfocusing in Chart Editor Playtest

MightyTheArmiddilo:
## Linked Issues
Nerp.

## Description
Unfocusing in Chart Editor's Playtest makes it so the welcome music timer starts because it doesn't check to see if it has a substate open so we check that now.

## Screenshots/Videos
The fixed bug in question:

https://github.com/user-attachments/assets/604d3b85-9c33-4566-ae7d-3fb7a688ccc3


Hundrec:
Thanks for providing this solution! I'll compare the one we have internally to this one and see which one makes it in.

Hundrec:
This should be fixed now, but let me know if it isn't and we still need this fix

PR #6598, opened 2025-12-23
[SYSTEM] Add credits to the github contribuitors.

SantiagoCalebe:
Tecnically, not an issue, but an add.

It adds a simple system to fetch the Github API to link and thank the github contribuitors. (It may be shit-coded, but meh)

I don't know if this is useful in the terms of the Crew, but i found it cool, i gave it a shot.

### DIFF
--- a/source/funkin/ui/credits/CreditsDataHandler.hx
+++ b/source/funkin/ui/credits/CreditsDataHandler.hx
@@ -18,6 +18,8 @@ class CreditsDataHandler
 
   /* 
     Fetch the github contributors.
+
+    THIS IS INDEED IN BETA TESTING, MAY NOT WORK AS EXPECTED.
   */
   public static function fetchContributors(onComplete:Void->Void):Void
   {


EliteMasterEric:
The plan was to do this eventually, but with a different implementation. This is because:

1. The API would include contributors who only started working on the game AFTER the update released
2. The API method requires making web requests for something that wouldn't change anyway
3. Accessing the API directly means wouldn't be able to make corrections (for example, letting contributors choose how they want to be credited)

I was intending to create a script that calls `git shortlog` to build a contributor list, and adds the contributors to a section of the credits, with a file that maps commit authorship emails to names to display (shortlog lists me three different times for example), with fallbacks for newly added names and such.

SantiagoCalebe:
> The plan was to do this eventually, but with a different implementation. This is because:
> 
> 1. The API would include contributors who only started working on the game AFTER the update released
> 2. The API method requires making web requests for something that wouldn't change anyway
> 3. Accessing the API directly means wouldn't be able to make corrections (for example, letting contributors choose how they want to be credited)
> 
> I was intending to create a script that calls `git shortlog` to build a contributor list, and adds the contributors to a section of the credits, with a file that maps commit authorship emails to names to display (shortlog lists me three different times for example), with fallbacks for newly added names and such.

Oooh, got it! Ty!

PR #6604, opened 2025-12-25
[ENHANCEMENT] Auto Detect Rimlight Alt Masks For Characters

VirtuGuy:
> [!WARNING]
> FunkinCrew/funkin.assets#322 is required!

## Description

This PR makes it so that character rimlight alt masks are checked for automatically. That means instead of having to do a million character id checks, you simply just have to include an alt mask image that's named after the character id in the newly created `masks` folder.

This PR also removes Nene's alt mask as not only is it outdated, but a change was made that makes it so that the alt mask isn't needed at all.

Tankman (Bloody)'s alt mask was updated, but it doesn't work due to an issue with `useRenderTexture`. Sure that change was completely useless, but it should work fine if alt masks on texture atlases gets fixed. I did test Tankman's alt mask with `useRenderTexture` disabled and the alt mask works fine apart from the lack of `useRenderTexture` problems.

### DIFF
--- a/source/funkin/graphics/shaders/DropShadowShader.hx
+++ b/source/funkin/graphics/shaders/DropShadowShader.hx
@@ -2,6 +2,7 @@ package funkin.graphics.shaders;
 
 import flixel.system.FlxAssets.FlxShader;
 import flixel.util.FlxColor;
+import funkin.play.character.BaseCharacter;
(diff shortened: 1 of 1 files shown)


realvirtu:
Uh yeah this might be fine, but I don't wanna fix assets :(

PR #6606, opened 2025-12-25
[ADJUSTMENT/BUGFIX] Changed the order that events are dispatched in PlayState

ComedyLost:
## Description
Changes the order that the PlayState dispatches events from 
`Module, Stage, Character, Song, Conversation, Note` to `Module, Song, Note, Stage, Conversation, Character`

The current order in which events are dispatched currently allows Note Kind or Song scripts that cancel the onNoteMiss / onNoteHit event to still play the note animations for the characters. The new order, recommended by both Kade & KoloInDaCrib, should fix this issue

> [!CAUTION]
> This is most likely a breaking change for scripts that were written with the original order in mind.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1496,26 +1496,26 @@ class PlayState extends MusicBeatSubState
 
   public override function dispatchEvent(event:ScriptEvent):Void
   {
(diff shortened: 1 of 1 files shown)


Trayfellow:
image

TechnikTil (approved):
waiter, more breaking changes please!

Issue #5928, opened 2025-09-06
Enhancement: Add proper modding support for custom health bars


requazar:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

Right now, I'm pretty sure there's no way to implement health bars other than through script files, which is very inconveniencing. My suggestion is to just make it a part of the NoteStyle associated with the song, just to keep things nice and simple

PurSnake:
hmmm

PurSnake:
I feel like I want to rewrite half the gameplay hud stuff for this 👀

Hundrec:
Rejected in #6623

PR #6623, opened 2026-01-02
[ENHANCEMENT] Add support for modded healthbars

requazar:
## Linked Issues
This PR implements the suggestion from [Issue #5928](https://github.com/FunkinCrew/Funkin/issues/5928).

## Description
This PR adds a `healthBar` value to the `assets` field of a notestyle data file, allowing for modders to easily implement custom healthbars without the need of scripts. It also adds an optional `padding` field in the data for `healthBar` in case the need arises to modify the size of the bar and not just the background for it. 

In the attached example, I've reduced the size of the healthbar background and changed its color. I've also attached the data in the notestyle's data file as an example of how one would implement a modded healthbar.

## Screenshots/Videos

screenshot-2026-01-01-17-27-45

image

### DIFF
--- a/source/funkin/play/notes/HealthBar.hx
+++ b/source/funkin/play/notes/HealthBar.hx
@@ -69,7 +69,16 @@ class HealthBar extends FlxSpriteGroup
     this.add(healthBar);
 
     noteStyle.applyHealthBarBGOffsets(this);
-    noteStyle.applyHealthBarOffsets(this);
+    if (!isDownscroll)
+    {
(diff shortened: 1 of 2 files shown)


requazar:
A nicer-looking demonstration of what this does:

screenshot-2026-01-02-14-34-14

image

EliteMasterEric:
For a feature that will probably never be used by the base game, we'd rather this stay as something you just do with mods. HScript should be perfectly capable of doing this.

requazar:
I see. Thanks for considering.

PR #6624, opened 2026-01-02
[BUGFIX] Prevent Importing non-Osu!Mania beatmaps

FuroYT:
## Description
After the merge of #6155 i realized that the game doesnt restrict you from importing non mania osu! beatmaps so this pull request fixes that behavior
this pr also cleans the import process code so its also kind of a chore for that part


Kade-github:
Looks promising, will take a look later.

EliteMasterEric (approved):

PR #6625, opened 2026-01-03
[ENHANCEMENT] Scroll the grid while dragging on a hold note

T5mpler:
This PR makes it so when you're dragging down a hold note, the grid scrolls with it if the mouse goes out of reach.

This makes it easier for if you're charting long hold notes, making it so you don't have to constantly click and drag on the hold note changing its length.

Speed at which the grid goes isn't accurate iirc, pretty sure I edited it as I was making the recording and it goes a bit more faster but yeah.

https://github.com/user-attachments/assets/9dd6d180-6e56-49e4-93d2-9d6131e6c93a

Ignore the people talking in the background I was in a call and forgot to defend and I just realized that 💀


Villezen:
defending da bed

PR #6706, opened 2026-01-07
[MODDING] Add more default imports used by mods

Starexify:
## Description

This PR converts the repetitive `Polymod.addDefaultImport` for default imports into an Array loop and adds several commonly used default imports by mods for convenience:
`flixel.FlxG`
`funkin.util.Constants`
`funkin.Preferences`

### DIFF
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -247,8 +247,18 @@ class PolymodHandler
   static function buildImports():Void
   {
     // Add default imports for common classes.
-    Polymod.addDefaultImport(funkin.Assets);
-    Polymod.addDefaultImport(funkin.Paths);
+    static final DEFAULT_IMPORTS:Array<Class<Dynamic>> = [
+      funkin.Assets,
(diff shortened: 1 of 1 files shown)


Starexify:
> I think that, to minimize confusion, the list of default imports should be as short as possible.
> 
> I'd limit the list to classes imported by default in the main package:

Preferences is actually a good idea since people might check for them in scripts.

But I was thinking if the `import.hxc` PR from polymod gets merged, the defaults shouldn't really be needed anymore since modders can have defaults there 🤔

EliteMasterEric:
> Preferences is actually a good idea since people might check for them in scripts.
> 
> But I was thinking if the `import.hxc` PR from polymod gets merged, the defaults shouldn't really be needed anymore since modders can have defaults there 🤔

The ones I selected get used in a large majority of scripts, and the base game imports them by default, so it improves consistency. One of the goals I've had with Polymod's HScript is making the code as similar as possible to direct source code written for the game.

EliteMasterEric (approved):

Issue #6676, opened 2026-01-06
Bug Report: Pressing 9 Twice During GF's Part in SPAGHETTI Resets to BF's Icon


Braite1:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

Google Chrome

### Version

0.8.1

### When triggering the bf-old icon easter egg while GF is currently your icon, and then un-triggering it will reset the icon back to BF
https://github.com/user-attachments/assets/65cd2c09-195c-42e8-9d7c-8c1c21e917e3

### Steps to Reproduce

**1. Go to Freeplay or Story Mode
2. Play SPAGHETTI
3. Around 2 minutes into the song, press 9 (or tap the icon on mobile), then press it again**

Hundrec:
Wow that's a really good catch

Hundrec:
I'm thinking the Easter egg should check if your icon is Boyfriend before changing to the old Boyfriend icon.
Seems simple enough, does anyone want to give it a shot?

PR #6717, opened 2026-01-09
Fix toggleOldIcon inconsistencies

notkaif:
## Linked Issues
https://github.com/FunkinCrew/Funkin/issues/6676

## Description
Toggling the old BF icon easter egg correctly reverts to what the current icon should be.
When the icon change function is called, if the bf-old icon is toggled on, it caches the icon and its offsets.
When the bf-old icon is toggled off, the game will correctly revert to the new icon with its offsets (if the event fired with any icon offset data)

Flip X functionality isn't cached, I honestly tried and it didn't work. Toggling the easter egg on will force the player icon to have flipX enabled, even when toggled off.

### DIFF
--- a/source/funkin/play/components/HealthIcon.hx
+++ b/source/funkin/play/components/HealthIcon.hx
@@ -31,10 +31,22 @@ import funkin.util.MathUtil;
 class HealthIcon extends FunkinSprite
 {
   /**
-   * The character this icon is representing.
+   * The character this icon will represent and display.
    * Setting this variable will automatically update the graphic.
    */
+  public var displayCharacterId(default, set):String = Constants.DEFAULT_HEALTH_ICON;
+
+  /**
+   * The character this icon should represent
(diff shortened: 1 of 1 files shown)


SnivySquid65 (approved):

realvirtu:
I haven't tested this PR, but if icon flipping works how it should, I'll close #6394.

Also for the next time you make a PR, I recommend making it on a new branch that's based on develop.

notkaif:
Yeah this was my first one, i was testing out pull requests with this one but thanks 😀

notkaif:
> I haven't tested this PR, but if icon flipping works how it should, I'll close #6394.

Toggling old bf will set flipX to true, exactly the same as your PR. 
I've tried to cache the flipX value however either I did something wrong or it simply just doesn't work. I'll try add it again when i get home.

notkaif:
https://github.com/FunkinCrew/Funkin/pull/6720

Issue #6667, opened 2026-01-05
Enhancement: Decimal Offset Values for Camera Bop Event


MistrDJ:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

The set camera bop event has a parameter for offsetting when the bop occurs by x beats. But I find this too limiting, I've had trouble charting songs where I want to offset the camera bop by _less_ than 1 beat, which doesn't seem to be possible, it must be a whole number to be recognized from what I can tell. So, my suggestion is this: update the Set Camera Bop event to understand a decimal value in the offset box, or allow for an option to toggle between offsetting by beats or _offsetting by steps instead._ This would give much more control in where you can place camera bops throughout a song.

<img width="617" height="333" alt="Image" src="https://github.com/user-attachments/assets/7eb3b12a-b84d-48db-a7b8-b68413c50d27" />

Hundrec:
Huge agree! I think the increments should go by 0.25 if the units are kept in beats

This limitation really holds back the event's potential

PurSnake:
Gotcha

TechnikTil:
hey guys look at the last 2 digits

Issue #4716, opened 2025-04-13
Enhancement: Allow decimal rates in Set Camera Bop Event


Average-FNF-Modder:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

Recently, functionality was added to allow the typing of event fields to be as specific as desired, so that you wouldn't be limited to any preset increments, but unfortunately, the set camera bop event does not allow you to make use of this new functionality. I'm suggesting the ability to make partial beats an option, this allows you to account for particularly funky rhythms, and also to have, say, 2 or even 4 bops in a beat for a particularly intense build-up.

MAZ12211:
Bump

PR #6741, opened 2026-01-11
[ENHANCEMENT] Add decimal value support for camera bopping rate and offset

KoloInDaCrib:
## Linked Issues
Implements https://github.com/FunkinCrew/Funkin/issues/6667
Implements https://github.com/FunkinCrew/Funkin/issues/4716

## Description
This PR turns the camera bop zoom and rate into values that support decimals up to .25 precision, considering every beat has four steps, allowing for further customization of the camera bopping behavior.

## Screenshots/Videos

https://github.com/user-attachments/assets/3d56152f-a003-4f18-a8f4-cb545cec84f5

https://github.com/user-attachments/assets/16524967-11d5-4d50-aa1e-fe162128a247

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -347,14 +347,14 @@ class PlayState extends MusicBeatSubState
    * How many beats (quarter notes) between camera zooms.
    * @default One camera zoom per measure (four beats).
    */
-  public var cameraZoomRate:Int = Constants.DEFAULT_ZOOM_RATE;
+  public var cameraZoomRate:Float = Constants.DEFAULT_ZOOM_RATE;
 
   /**
    * How many beats (quarter notes) the zoom rate is offset.
    * For if you want the zoom to happen off-beat.
    * @default Zero beats (on-beat).
    */
-  public var cameraZoomRateOffset:Int = Constants.DEFAULT_ZOOM_OFFSET;
+  public var cameraZoomRateOffset:Float = Constants.DEFAULT_ZOOM_OFFSET;
 
   /**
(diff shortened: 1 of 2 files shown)


Starexify:
Kolo, you forgot to link 
- #4716 

😬

Hundrec:
Finally, a functionality this event has always needed

Does this also allow irregular rate/offset values like 0.33 or 2.11?

KoloInDaCrib:
> Finally, a functionality this event has always needed
> 
> Does this also allow irregular rate/offset values like 0.33 or 2.11?

Unfortunately not, as there's only 4 steps in a beat, so the decimal values are capped to the multiples of .25, similar to character bopping values.

Hundrec:
I can see cases where things get really irregular and irregular values would be necessary, but being able to access all 16 steps in a section is a huge improvement already

Starexify:
> > Finally, a functionality this event has always needed
> > 
> > Does this also allow irregular rate/offset values like 0.33 or 2.11?
> 
> Unfortunately not, as there's only 4 steps in a beat, so the decimal values are capped to the multiples of .25, similar to character bopping values.

But since its haxeui, cant you just set a custom number anyway so it allows 0.24 or such (because stepper doesnt check and fixes it) ?

KoloInDaCrib:
> But since its haxeui, cant you just set a custom number anyway so it allows 0.24 or such (because stepper doesnt check and fixes it) ?

That's not the problem here. The problem is that camera bopping is tied to hitting a beat (or in this PR, a step) and irregular decimals would require moving that behavior to the update loop. The big downside with that is that it's incredibly unprecise.

TechnikTil:
#6741

PR #6754, opened 2026-01-13
[CLEANUP] Remove redundant `Object` and `ScriptedObject` classes.

Starexify:
## Description

This PR removes the redundant `Object` and `ScriptedObject` classes from the base modding folder as there are no longer necessary in modding for initializing base classes.
Since ~0.7 polymod was updated to support initializing base classes without the need of using wrappers like `ScriptedObject`.

So instead of 
```haxe
class Test extends Object {
}
```
Mods can simply use these:
```haxe
class Test {
}
```


KoloInDaCrib:
honestly i see scenarios where the Object class and its scripted variant could be used, as the objects extending nothing in polymod tend to not work sometimes.

NotHyper-474:
In an ironic turn of events, this PR is now also redundant: https://github.com/FunkinCrew/Funkin/commit/0311c7373e8fc7915140acfec2fe7acd7010f720

PR #6765, opened 2026-01-14
[BUGFIX] Fix 2nd measure number not being rendered

NotHyper-474:
## Description
This fixes an issue where the 2nd measure tick would not be rendered when first entering the chart editor because, since a song has not been loaded yet, it will fallback to just generate measures for a song of a very short length. This fixes it by triggering an update when a song is loaded.

This could've been an one line banger but I've also addressed an issue where the previous to current measure checking was useless 'cause the previous value was never updated.

## Screenshots/Videos

https://github.com/user-attachments/assets/46c3ac9b-ab4a-4c87-b92a-47f69ddd06d4


realvirtu:
what the shi

PR #6833, opened 2026-01-28
[BUGFIX] Fix Chart Editor Waveforms desyncing on BPM change

gamerbross:
## Linked Issues
Couldn't find any.

## Description
Vocals waveform didn't stretch/compress when the song tempo/bpm changes midway, making them impossible to use.

## Screenshots/Videos

https://github.com/user-attachments/assets/3f93f3d0-6bfa-4bfe-8b13-da10d22cb4b5

_Before, audio waveforms being shorter than they should, turning almost useless_

https://github.com/user-attachments/assets/f4b81877-2758-41a7-a4ff-4c051e75cddf

_After, audio waveforms getting stretched to match the current tempo_

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -384,7 +384,7 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
 
         for (member in audioWaveforms.members)
         {
-          member.time = Conductor.instance.songPosition / Constants.MS_PER_SEC;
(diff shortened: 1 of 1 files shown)


Hundrec:
This is an awesome fix, and a one-line banger no less!

I just wonder if there's also a way to prevent the slight shifting around of the waveforms with some more changes?

gamerbross:
i think adding something like that would require a way to render the waveform with different widths, which i have no idea how to, i just messed with the existing things and got this
image
(really dumb image example)

Hundrec:
Thanks for the response!

I'll get this change in since it's a huge improvement anyway :)

PR #6860, opened 2026-02-01
[ENHANCEMENT] Clean up `GRhythmUtil.mirrorNoteDirection()` Along With A Few Other Changes

MightyTheArmiddilo:
This PR cleans up `GRhythmUtil.mirrorNoteDirection()` due to how monstrous it currently looks, along with adding checks for `Strumline.KEY_COUNT` instead of hard coding it into the function (so source coders don't have to rewrite the function themselves).

This PR also removes an unused temporary variable in `PlayState`, makes `Strumline.KEY_COUNT` public, changes a temporary `noteData` variable to use `songNote.data` instead of `songNote.getDirection()`, and formats `GRhythmUtil.hx`.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2486,10 +2486,7 @@ class PlayState extends MusicBeatSubState
         if (noteKind != null) scoreable = noteKind.scoreable;
       }
 
-      var noteData:Int = songNote.getDirection();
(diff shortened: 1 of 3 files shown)


Kade-github (approved):
this should be brought over yuh

EliteMasterEric (approved):

PR #6881, opened 2026-02-04
[CHORE] Typo in `ScriptEventType.hx`

MightyTheArmiddilo:
## Description
Typos all around!!!

## Screenshots/Videos
image


Hundrec:
While I'm not sure which one I use myself, I'd say we should keep it to the American usage (with one L)

MightyTheArmiddilo:
> While I'm not sure which one I use myself, I'd say we should keep it to the American usage (with one L)

The fact that I'm American and went for the British way

MightyTheArmiddilo:
###### Til you set me up

PR #6886, opened 2026-02-05
[BUGFIX] Don't dispatch conductor events if substate is active

KoloInDaCrib:
## Linked Issues
Nope.

## Description
When in a SubState, for example after dying in PlayState, the callbacks beatHit and stepHit of the parent state are still attached to the same Conductor instance and get dispatched. This leads to a weird effect where if you die in the week 2 stage, lightning can still strike. 
This PR fixes it by checking if the parent state has a substate attached to it and if it can update with it attached before dispatching the conductor callbacks.

## Screenshots/Videos
image
The trace was added manually to the script and doesn't exist in the original file.


EliteMasterEric (approved):
Code looks good to me.

PR #6902, opened 2026-02-08
[BUGFIX] Fix `FunkinSprite` atlas functions not working on Sparrow+Animate sprites

TechnikTil:
## Description
When playing an animation that is not an animate atlas. You cannot use any `FunkinSprite` atlas edit functions at all. It seems intentional until you think about any sprites that use both animation types. You will not be able to use those functions until you are in a atlas animation. This PR fixes this problem by replacing the `this.isAnimate` check with `this.library != null`, as that is what `animate.FlxAnimateController` uses under the hood.

### DIFF
--- a/source/funkin/graphics/FunkinSprite.hx
+++ b/source/funkin/graphics/FunkinSprite.hx
@@ -397,7 +397,7 @@ class FunkinSprite extends FlxAnimate
     {
       return true;
     }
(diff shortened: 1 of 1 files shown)


AbnormalPoof:
already did this internally 😵‍💫😵‍💫

PR #6912, opened 2026-02-10
[BUGFIX] Chart Editor - Ctrl click on hold note NOR fix (again) 

Lasercar:
## Linked Issues
None (and hopefully this issue continues to stay that way).

## Description
Ctrl clicking on a hold note causes a null reference error because it tries to check if a regular note is selected, rather than a hold note.

This simply changes it to check for a hold note instead.

Now also includes a comment specifying the issue this fixes so that the change doesn't get reverted again in the future.


Hundrec:
I did this internally a few days ago, but haven't gotten around to pushing it
So I'll merge this into dummy for now

PR #6915, opened 2026-02-10
[ENHANCEMENT] Add a directory parameter to other functions in `Paths` for better path handling.

Starexify:
## Description
This PR adds a `directory` parameter to other functions in `Paths`, similar to how `Paths.srt` contains an optional `directory` parameter with a default value.

This makes it easier to access assets like json files outside the `data` folder or other assets better than having to use a completely different approach just for that directory change.


Hundrec:
Would this change still be relevant on `preview/public-playtest`?

PR #6929, opened 2026-02-12
[BUGFIX] Hide the Chart Editor Event/Note tooltips when killing said objects

KoloInDaCrib:
## Linked Issues
None.

## Description
When removing notes/events in the chart editor, the tooltip region instance is still left in the region map, which causes the tooltips to show even when their parent object is killed for recycling. This PR fixes that by removing the tooltip region whenever `kill()` is called on the parent object. Tooltips get re-registered whenever the positions update anyways.

## Screenshots/Videos

https://github.com/user-attachments/assets/2c847790-32c1-46ac-869d-45f5f83137b5


Marcosayfi:
FINALLY THE MOST ANNOYING BUG EVER IS FIXED

PR #6954, opened 2026-02-19
[BUGFIX] Save Kevin and Michael

NotHyper-474:
## Linked Issues
The internal Linear issue made for this bug.

## Description
With the introduction of blend modes came the addition of the `FunkinCameraFrontEnd`, this, however, broke `TouchPointerPlugin`, because it is now adding the cameras that render our touch buddies to the old frontend.
This PR fixes the issue by moving where the frontend is changed to be before any plugins are initialized, which is when `FlxGame` is created.

Turns out Kevin and Michael were just hiding this whole time and were in no danger :D

## Screenshots/Videos
https://github.com/user-attachments/assets/b2558da8-fd23-478e-af25-1d5c88535af9

### DIFF
--- a/source/Main.hx
+++ b/source/Main.hx
@@ -158,6 +158,10 @@ class Main extends Sprite
 
     WindowUtil.setVSyncMode(funkin.Preferences.vsyncMode);
 
+    // Force a `FunkinCamera` to be the default camera.
+    // This allows the blend mode shader to work everywhere.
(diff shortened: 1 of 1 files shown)


TechnikTil (approved):
8 line banger

DustedDustin:
thank you for saving kevin and michael, maybe we can save week 8 next?

Hundrec (approved):
Image

PR #6963, opened 2026-02-21
[ENHANCEMENT] Play SFX When Undoing An Event

VirtuGuy:
## Description

This PR makes it so that an sfx plays when undoing an event, like how it is when undoing literally anything else that has to do with notes or events.

## Screenshots/Videos

Below is a video demonstrating what this PR does.

https://github.com/user-attachments/assets/22234615-0ae6-4ca8-9133-b8d81a4c571d


Hundrec (approved):
Genius

PR #6964, opened 2026-02-21
Enhanced MathUtil with constants and utility functions!!

HeroEyad:
Hey Funkin Crew! I gave `MathUtil` a lil tune up to improve performance during heavy update loops and added a few standard game dev tools.

### Why did I add a custom `PI` constant?

Instead of using Haxe's `Math.PI`, I added our own `inline final PI`. Here is why:

* **Consistency:** Haxe compiles to multiple targets (C++, JS, etc.). Hardcoding it ensures the exact same precision everywhere.
* **Pre-calculated math:** It let me cleanly define `TAU` (`PI * 2`), `HALF_PI`, `DEG2RAD`, etc. Since they are `inline final`, the compiler hardcodes the results so we aren't wasting CPU calculating `Math.PI / 180` every single frame.

### Performance Tweaks

* **Faster `gcd`:** Swapped out `Math.abs` and `Math.floor` for strict integer logic. This stops the game from constantly converting Ints to Floats and back.
* **Ditched `Math.pow`:** Swapped `Math.pow(x, 3)` for `x * x * x` in easing functions. Native multiplication is significantly faster than library calls.
* **Cleaned up Easing:** Removed a `Math.isNaN` check in `easeInOutCirc` that was mathematically impossible to fail anyway.
* **Inlining:** Added `inline` to most small helpers so the compiler pastes the math directly, eliminating function-call overhead.

### New Tools

* **`remap` & `inverseLerp`:** Perfect for UI (e.g., mapping a 0-100 health value to a 0-300 pixel width bar).
* **`wrap`:** Easily loop angles between 0 and 360 without messy modulo math.
* **`approximatelyEqual`:** Fixes floating-point errors (where `0.1 + 0.2 = 0.30000004`) using a tiny `EPSILON` tolerance.
* **`smoothStep`:** Standard GLSL-style smooth interpolations.

### DIFF
--- a/source/funkin/util/MathUtil.hx
+++ b/source/funkin/util/MathUtil.hx
@@ -1,16 +1,55 @@
 package funkin.util;
-
 /**
  * Utilities for performing mathematical operations.
  */
 @:nullSafety
 class MathUtil
 {
+  /** Euler's constant and the base of the natural logarithm. */
+  public static inline final E:Float = 2.71828182845904523536;
+
+  /** The ratio of a circle's circumference to its diameter. */
+  public static inline final PI:Float = 3.14159265358979323846;
+
+  /** Tau (2 * PI). Represents a full 360-degree rotation in radians. */
+  public static inline final TAU:Float = 6.28318530717958647692;
+
+  /** Half PI (PI / 2). Represents 90 degrees in radians. */
+  public static inline final HALF_PI:Float = 1.57079632679489661923;
+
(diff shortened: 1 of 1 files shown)


macohi:
heroEyad

RagdollClash:
Just curious, what tools did you use to make this PR?

HeroEyad:
wym what tools

PR #6965, opened 2026-02-22
[ENHANCEMENT] Smooth lerp scaling for Touch Play Button

AnimatingLegend:
## Description
Updated `touchHereToPlay` button in [`FunkinPreloader.hx`](https://github.com/FunkinCrew/Funkin/blob/main/source/funkin/ui/transition/preload/FunkinPreloader.hx) to make feel more polished, instead of scale jumping instantly.

### Changes:
- Added a lerp calculation so the sprite's scale catches up to the target scale.
- Swapped `MOUSE_OVER` and `MOUSE_OUT` events out for a `hitTestPoint()` check.
- Left old mouse functions blank since the scaling is now being handled in `updateGraphics()`.

### Before (itch.io build):
https://github.com/user-attachments/assets/82001302-cd26-434a-9c51-6ac17cd01325
### After:
https://github.com/user-attachments/assets/b2000790-7f25-4313-bc61-8fb5c870cb31

### DIFF
--- a/source/funkin/ui/transition/preload/FunkinPreloader.hx
+++ b/source/funkin/ui/transition/preload/FunkinPreloader.hx
@@ -52,6 +52,9 @@ class FunkinPreloader extends FlxBasePreloader
   // Ratio between window size and BASE_WIDTH
   var ratio:Float = 0;
 
+  // The target scale of `touchHereToPlay` sprite.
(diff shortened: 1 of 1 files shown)


AnimatingLegend:
some of my other branches overrode into this one... ignore this PR

Issue #6995, opened 2026-02-24
Bug Report: Stage Editor reapplies offsets when selecting character with mouse


Prod-42:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [ ] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.8.3

### Description (include any images, videos, errors, or crash logs)

While trying to move a character around on the stage editor, their X and Y position kept increasing everytime I clicked on them with the Character Properties window open, this only happens with mouse, the keyboard works totally fine.

https://github.com/user-attachments/assets/82d3f1dc-1e00-479c-9d07-0d44400520bc

### Steps to Reproduce

1. Open the Stage Editor
2. Open the Character Properties Window
3. Try moving the characters with a mouse
4. Now they're running away from you!!!

PR #7002, opened 2026-02-25
[BUGFIX] Fix characters with global offsets not getting properly moved in the Stage Editor

KoloInDaCrib:
## Linked Issues
Closes https://github.com/FunkinCrew/Funkin/issues/6995

## Description
The issue was basically that, because offsets get applied in `getScreenPosition()`, the calculation gets messed up every time a character with offsets was getting moved with the mouse in the Stage Editor.

## Screenshots/Videos

https://github.com/user-attachments/assets/6123dae2-d238-45e3-9d5f-7d8dc99527cd

### DIFF
--- a/source/funkin/ui/debug/stageeditor/StageEditorState.hx
+++ b/source/funkin/ui/debug/stageeditor/StageEditorState.hx
@@ -797,13 +797,13 @@ class StageEditorState extends UIState
         {
           this.createAndPushAction(CHARACTER_MOVED);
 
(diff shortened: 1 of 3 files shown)


Marcosayfi:
nice. i've had this bug for a while.

Hundrec (changes requested):
Strangely enough, this is still completely broken for me
It doesn't seem to behave differently between 0.8.3 release and `develop`
When I click on any character, it flies away from my mouse at lightspeed

realvirtu:
`FlxG.mouse` has `deltaX` and `deltaY`. Why not just add those variables to the character's position? That way, you won't have to worry too much about global offsets.

Issue #5988, opened 2025-09-12
Enhancement: Make Mod Dependencies read the Mod Title rather than Folder Name


ZazaFi:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

Current for setting up Mod Dependencies, the thing the dependencies table scans for is the Folder Name of the mod, rather than scanning for the polymod file's title for the mod.

this is rather ineffective, as it might cause certain mods for mods to not function because the update added a version number to the folder name, or had "  -  " added to them instead of spaces due to something like gamebanana's downloading system.

An alternative solution would be to add a "Internal Mod Name" that is scanned for by dependency tables if for some reason people really want to keep changing their mod titles.

moxie-coder:
I absolutely agree with this honestly

Trayfellow:
YES PLEASE

Hundrec:
I feel like Eric mentioned this at some point, so I'll pass this to him

PR #7003, opened 2026-02-25
[FEATURE/Modding] Add support for the new Polymod mod ID system.

### LINKED ISSUE #3991: Enhancement: more consistent mod dependency behavior

Starexify:
## Linked Issues
- Closes #3991
- Honorable Mentione #5988

> [!IMPORTANT]
> - Requires the following PR to be merged: https://github.com/FunkinCrew/polymod/pull/330

## Description
This PR updates the `PolymodHandler` to use the new `dirName` instead of `id` for loading mods.

**Note:** This is a breaking change for mods such as the Funkin' Mod Menu if they dont switch to using Dirs and `dirName` instead of Ids !

So let's take this file Tree as example:
```
.
└── mods/
    ├── Test
    └── TestTwo
```

The mod "Test" will be the dependency mod for "TestTwo". And its metadata doesn't contain an `id`, that means the dependency must be tied to the directory name:
```json
// TestTwo/_polymod_meta.json
  "dependencies": {
    "Test": "<VERSION>"
  }
```
But if a user modifies the directory name it will no longer work (check the polymod PR for reasons why).

That's why mods can add a "static" mod ID, which the dependencies could use and work even if the directory name changes:
```json
// Test/_polymod_meta.json
  "id": "Test"
```

> [!NOTE]
> This doesn't change the old behavior and mods don't need to update to this system immediately !

## Screenshots/Videos

https://github.com/user-attachments/assets/59ec6bd7-2494-4ff1-9405-9eeb35f2633f

### DIFF
--- a/source/funkin/modding/PolymodErrorHandler.hx
+++ b/source/funkin/modding/PolymodErrorHandler.hx
@@ -15,6 +15,12 @@ class PolymodErrorHandler
       //
 
       case MOD_MISSING_DIRECTORY:
+        // A mod directory was included in the list of mods to load, but it isn't installed.
+        trace(' WARNING '.warning() + 'Tried to load a mod that was not installed: ${error.message}');
+        // Notify the user via popup.
+        funkin.util.WindowUtil.showError('Mod Load Error', error.message);
+
+      case MOD_MISSING_ID:
         // A mod ID was included in the list of mods to load, but it isn't installed.
         trace(' WARNING '.warning() + ' Tried to load a mod that was not installed: ${error.message}');
         // Notify the user via popup.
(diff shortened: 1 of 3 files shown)


CrusherNotDrip:
Honorable Mentione

PR #7005, opened 2026-02-26
[SECURITY VULNERABILITY] Fix ReflectUtil by adding Polymod Blacklist

FuroYT:
## Linked Issues
N/A

## Description
This fixes a major bug where you could still access blacklisted classes through FlxSave.resolveFlixelClasses with ReflectUtil as it wasn't patched out to block certain fields but this Pull Requests fixes that and along side, it allows new functions onto ReflectUtil which are also protected via the same blacklist and import aliases

New Allowed but protected functions
- `callMethod`
- `createEmptyInstance`
- `createInstance`
- `resolveClass`
- `resolveEnum`

Fixed Functions due to accessors and aliases
- `deleteAnonymousField` (fixed accessing fields)
- `getAnonymousField` (now uses aliases)
- `getProperty` (now uses aliases)
- `hasAnonymousField` (now uses aliases)
- `setAnonymousField` (fixed accessing fields and using aliases)
- `setProperty` (fixed accessing fields and using aliases)

Please do intensive testing so bugs doesn't slip through this


source/funkin/util/ReflectUtil.hx
```diff
@@ -277,62 +284,98 @@ class ReflectUtil
...
+   * @param cls The class to create.
+   * @throws error When trying to create a blacklisted class.
    */
   @SuppressWarnings("checkstyle:FieldDocComment")
   public static function createEmptyInstance(cls:Class<Any>):Any
   {
-    throw "Function Type.createEmptyInstance is blacklisted.";
+    if (!isAccessAllowed(cls))
+    {
+      throw 'Attempted to call createInstance onto a blacklisted class "${getClassNameOf(getClassOrEnumFromName(cls))}"';
```
EliteMasterEric:
```
Class<Any> should be String
For function argument 'path'
```

EliteMasterEric (changes requested):
Code for this looks great, but I get several build errors from trying to use it myself!

FuroYT:
Oops, let me fix them asap

FuroYT:
Fixed the errors

EliteMasterEric (approved):
This is a great way of re-implementing this functionality in a way that still prevents vulnerabilities. I like the fact that it automatically pulls from the existing blacklist so we only need to update things once.

PR #7015, opened 2026-02-27
[BUGFIX] Fix Rank Slamming Animation

FuroYT:
## Description
This PR fixes the rank slamming animation that was mising the glow of the capsule and it also polishes it by applying transformations into the group itself instead of the capsule display sprite so it looks cleaner also added a option into the results debug to test the rank slamming from results

## Screenshots/Videos

https://github.com/user-attachments/assets/8685c11d-2c98-4ea3-872b-31a5cd9fe6e5


realvirtu:
0.8.furo when???

PR #7068, opened 2026-03-04
[BUGFIX] Fix sound tray playing multiple sounds at once and not appearing during lag

Raltyro:
## Linked Issues
None

## Description
Fixes various issues in sound tray which listed below:
- Fix sound tray playing the same volume indicator sounds 10 times.
- Fix sound tray not appearing next time if lag spikes occour during transitioning to dissapear.
- Fix sound tray not appearing sometimes first time used.
- Fix starting y position of the sound tray sprite instead of 0.

## Screenshots/Videos

Before:

https://github.com/user-attachments/assets/37b8eaad-eda9-454b-9707-10929671774a

After:

https://github.com/user-attachments/assets/dbeba470-05b2-482d-aea5-735d135c8866

### DIFF
--- a/source/funkin/ui/options/FunkinSoundTray.hx
+++ b/source/funkin/ui/options/FunkinSoundTray.hx
@@ -64,6 +64,7 @@ class FunkinSoundTray extends FlxSoundTray
     }
 
     screenCenter();
+    y = -height - 10;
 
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
This might just be the final bug we've found on SDL3.
Thanks!

Issue #7044, opened 2026-03-02
Bug Report: Character doesn't return to idle when continually hitting No Animation notes


MantisDGT:
[###]() Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.8.3

### Description (include any images, videos, errors, or crash logs)

If the player hits a constant string of notes labeled as the 'no-anim' notekind, the player character will not return back to their idle animation after they miss a note.

![Image](https://github.com/user-attachments/assets/47709c21-db95-41c9-b02e-6c1e4e9647fe)
(sorry for the low quality, this is the only way GitHub would allow me to post/show a video)

### Steps to Reproduce

1. Create a chart with a string of notes that is fast enough to stop the player from returning to their idle, with all notes as the 'no-anim' notekind.
2. Miss anywhere whilst in a section with the notes.
3. Return to pressing the rest of the notes.
4. The player does not return to their idle animation unless they are not hitting notes anymore (all notes still have the 'no-anim' notekind.

MantisDGT:
If anyone could tell me how to also include a video to help with explaining this bug, that would be great, as I could not find the option to also include a video/screenshot.

AnimatingLegend:
> If anyone could tell me how to also include a video to help with explaining this bug, that would be great, as I could not find the option to also include a video/screenshot.

streamable or youtube works

Starexify:
> streamable or youtube works

or just drag and drop the video (tho theres size limits)

AnimatingLegend:
i can reproduce this also, ill look into it.

Starexify:
hopefully today Ill make time to check the no anim stuff

AnimatingLegend:
> hopefully today Ill make time to check the no anim stuff

please do it will be a big help 🙏

PR #7078, opened 2026-03-04
[BUGFIX] Fix character stuck in miss pose during no-anim notes

AnimatingLegend:
## Linked Issues
#7044 - Character doesn't return to idle when continually hitting No Animation notes

## Description
This PR should fix characters being stuck in their miss-animations instead going back to their dance / idle poses.

### BEFORE:
https://github.com/user-attachments/assets/6e6a473e-d613-4d3d-9716-05630922e843
### AFTER:
https://github.com/user-attachments/assets/a43ace24-7c06-446b-b9c2-991d2c348a9c

### DIFF
--- a/source/funkin/play/character/BaseCharacter.hx
+++ b/source/funkin/play/character/BaseCharacter.hx
@@ -371,7 +371,11 @@ class BaseCharacter extends Bopper
     // Reset hold timer for each note pressed.
     if (justPressedNote() && this.characterType == BF)
     {
-      holdTimer = 0;
+      // If the note kind has `noanim` do not reset holdTimer.
+      if (curNoteKind != null && !curNoteKind.noanim)
+      {
+        holdTimer = 0;
+      }
     }
 
     if (isDead)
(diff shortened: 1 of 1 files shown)


PurSnake:
I think it should be more about HoldTimer inside BaseCharacter.hx for BF character's type

PurSnake:
I think just snapping to idle is kinda ehhh, it doesnt look good

AnimatingLegend:
> I think it should be more about HoldTimer inside BaseCharacter.hx for BF character's type
>
> I think just snapping to idle is kinda ehhh, it doesnt look good

Yeah, im revising currently, should be working properly soon!

Marcosayfi:
i like this change

AnimatingLegend:
> Getting closer! One last detail that will make the notes really seem like they're "not there" is that hitting a NoAnim note shouldn't snap the player back to idle, but rather keep the player holding the previous animation (whether that be a miss or some special animation)
> 
> I really want this notekind to be usable for situations where you need notes that must be hit, but have no animations attached to them

I can see what I can do 🫡

PR #7081, opened 2026-03-05
[ENHANCEMENT] Added an option called "Cutscenes on Freeplay" to the preference menu

SemizotluYogurt:
# This change depends on my funkin.assets repo PR here:
**[PR #377](https://github.com/FunkinCrew/funkin.assets/pull/377)**

## Description
This PR is adding an option that disables start cutscenes on freeplay song, like: Pico, Philly, Blammed pico mixes, Spaghetti, Stress Pico Mix, Senpai and Roses pico mixes.

## Screenshot
fenasal funkin

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -540,6 +540,25 @@ class Preferences
     return value;
   }
 
+  /**
+   * If enabled, cutscenes and dialogue sequences will be played on freeplay.
+   * @default `true`
+   */
+  public static var freeplayCutscenes(get, set):Bool;
+
+  static function get_freeplayCutscenes():Bool
+  {
+    return Save?.instance?.options?.freeplayCutscenes ?? true;
+  }
+
+  static function set_freeplayCutscenes(value:Bool):Bool
+  {
+    var save:Save = Save.instance;
+    save.options.freeplayCutscenes = value;
+    Save.system.flush();
+    return value;
+  }
+
   #if mobile
   /**
    * If enabled, device will be able to sleep on its own.
--- a/source/funkin/save/Save.hx
+++ b/source/funkin/save/Save.hx
@@ -107,6 +107,7 @@ class Save implements ConsoleClass
         debugDisplay: 'Off',
         debugDisplayBGOpacity: 50,
         subtitles: true,
+        freeplayCutscenes: true,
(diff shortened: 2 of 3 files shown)


realvirtu:
I think the reason why cutscenes play outside of freeplay is because it's called "story" mode. Cutscenes exist to tell the story. Pretty sure the point of freeplay is to revisit those songs and replay them either for the sake of playing them, or for a better score. I don't think people want to watch cutscenes every time they play a song.

Hundrec:
Cutscenes from the story should be kept to Story Mode.
Freeplay-exclusive cutscenes like Stress (Pico Mix) would play every time but be easily skippable.
This is sufficient for a good experience.

SemizotluYogurt:
> Cutscenes from the story should be kept to Story Mode. Freeplay-exclusive cutscenes like Stress (Pico Mix) would play every time but be easily skippable. This is sufficient for a good experience.

Yeah they are skipable but they are very annoying. Everyone I ask says that constantly skipping them is annoying.
I still think that this or a similar feature needs to be added.

SemizotluYogurt:
> I think the reason why cutscenes play outside of freeplay is because it's called "story" mode. Cutscenes exist to tell the story. Pretty sure the point of freeplay is to revisit those songs and replay them either for the sake of playing them, or for a better score. I don't think people want to watch cutscenes every time they play a song.

Yeah I know, and this PR was fixing that issue... No one wants to watch the dialouge cutscene or stress cutscenes on pico mixes CONSTANTLY.

SemizotluYogurt:
> Cutscenes from the story should be kept to Story Mode. Freeplay-exclusive cutscenes like Stress (Pico Mix) would play every time but be easily skippable. This is sufficient for a good experience.

I think the feature may have been misunderstood. This feature is ment to skip pico mix cutscenes and spaghetti.
Would it be okay if I open a new PR with a clearer explanation and change the name?

Hundrec:
I understand, but the team also wouldn't want this to be a setting by design.
Maybe we could make skipping cutscenes a bit easier later on, but this is how it will stay for now.

SemizotluYogurt:
> I understand, but the team also wouldn't want this to be a setting by design. Maybe we could make skipping cutscenes a bit easier later on, but this is how it will stay for now.

Okay then...

PR #7094, opened 2026-03-07
Big Batch of Checkstyle Rule Changes

EliteMasterEric:
I've wanted to make a bunch more changes to the repository's `haxe-formatter` and `haxe-checkstyle` rules, and I decided the best way to get some feedback was to make a public branch for people to review since I think one or two of the current rules might be a bit inconvenient right now.

## Code Style Rule Changes

- Add rules for line wrapping on arrays and maps.
  - If any one entry is >32 chars long, if there are at least 5 items, if there are any items that span multiple lines, or the total length of all wrappable items spans 160 characters, force each array element to its own line.
  - We can evaluate different arrays in the codebase and choose to add more specific rules for this if we want.
- Fix an issue where if you had an `enum abstract`, it would sometimes display a warning that the type wasn't documented even though it clearly is.
- Disable a style rule where it would request newlines for macro reification (like `macro $v{value}`)
- Enforce a newline after the type metadata. I wanted to enforce a newline for each metadata on a class type but couldn't find a rule for that.
- Enable a rule where modifiers (`public`, `static`, `inline`, `final`, etc.) must be in a specific pre-determined order. Please stop putting `override public` in your code it implicitly sounds wrong, like saying "green big house"
- Modify the naming regexes for local variables and member variables/functions, since I made a PR to support splitting rules up based on whether the variable is `final` or not.
- Require parenthesis around the switch condition (I didn't know they were optional but someone made a PR to add a rule for it).
- Disable enforcement of trailing commas on arrays and object literals (we could force them to be included, thoughts? there's no way to force them to be excluded right now).
- Inline functions are now required to be formatted as arrow functions.
- Disable the use of nested ternary operators (like `foo ? bar ? true : false : false)`.
- Add a rule requiring large decimal numbers to use separators (`1235678` must now be formatted as `1_235_678`, yes you can use underscores in the middle of numbers in Haxe)
- I made a PR just so I could implement a rule requiring the use of curly braces on any `if` block that has a corresponding `else` block.
- Doc comments now require `/**` on open, `*` on each line, and `*/` on close, matching the syntax used by both Java and HaxeFlixel.

# NOTE

To properly view and use these rules in your copy of VSCode, you'll need to use a fork of the Haxe Checkstyle extension that supports some of the new rules that are included: https://github.com/EliteMasterEric/vscode-checkstyle/releases

The PR is like 20 commits long but each commit only changes one rule so you can see exactly what each rule change is. Note that this PR will itself require a follow-up PR to apply the style changes to all the Haxe files in the project, and that will deliberately be a separate chore PR.

### DIFF
--- a/checkstyle.json
+++ b/checkstyle.json
@@ -270,7 +270,6 @@
           "SWITCH",
           "TRY",
           "CATCH",
-          "REIFICATION",
           "ARRAY_COMPREHENSION"
         ],
         "severity": "INFO"
--- a/checkstyle.json
+++ b/checkstyle.json
@@ -51,6 +51,14 @@
       "type": "AvoidStarImport"
     },
     {
+      "comment": "If enabled, prevents ternary operators from being nested.",
+      "props": {
+        "severity": "INFO",
+        "maxDepth": 1
+      },
+      "type": "AvoidTernaryOperator"
+    },
+    {
       "comment": "If enabled, prevents conditional compilation from crossing blocks. This prevents the case where conditional compilation could potentially create invalid syntax.",
       "props": { "severity": "INFO" },
       "type": "BlockBreakingConditional"
--- a/checkstyle.json
+++ b/checkstyle.json
@@ -99,7 +99,17 @@
       "props": {
         "ignoreExtern": true,
         "format": "^?:_[a-zA-Z0-9]+(?:_[a-zA-Z0-9]+)*$",
-        "tokens": ["INLINE", "NOTINLINE"],
+        "tokens": ["INLINE", "NOTINLINE", "NOTFINAL"],
+        "severity": "INFO"
+      },
+      "type": "ConstantName"
+    },
+    {
+      "comment": "Specifies a specific naming format for static final identifiers.",
+      "props": {
+        "ignoreExtern": true,
+        "format": "^[A-Z]([_A-Z0-9]+)*$",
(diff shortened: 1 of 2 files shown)


Huuuuuuuuunk:
good job as always eric

FuroYT:
i don't think the change of the modifier order is necessary

`override public` is fully fine because you say "i am overriding the public function" and not "i public override the function" or "i publicly override the function" (- Starexify)

Funkin' Already uses both in the codebase so making it a code stylecheck doesn't feel right (with the new order in here) in my opinion

EliteMasterEric:
> Funkin' Already uses both in the codebase so making it a code stylecheck doesn't feel right (with the new order in here) in my opinion

The fact that it uses both in the codebase is WHY we need a style check, to enforce one consistent style.

EliteMasterEric:
also I just pushed a fix so if you were testing this branch you'll have to re-download my checkstyle fork

Fixes include:
- Fix an issue where the DocStyleComment rule was broken so it'd complain about every comment for no reason.
- Refactor the LongNumber rule so that digit groups have to be a specific size (so `12_34_56_78` is no longer valid).

Keoiki:
> Please stop putting `override public` in your code it implicitly sounds wrong, like saying "green big house"

So you'd rather say that something is "a public override of a function"? What does "a public override" even mean???

PR #6135, opened 2025-09-25
Make the HashLink target work

NotHyper-474:
## Linked Issues
-. -..-. .-

## Description
This is quite an old branch of mine that's been sitting on my repository since 2024, finally becoming a PR now that there seems to be a growing interest in building for HashLink. It fixes issues that prevent someone from compiling to HashLink, which could be a pretty useful target for easily making and testing PRs like I did when making https://github.com/FunkinCrew/Funkin/pull/3574.

### Dependent on:
* https://github.com/FunkinCrew/flixel/pull/7
  - Has commit [FunkinCrew/flixel@`6f8ee63` (#7)](https://github.com/FunkinCrew/flixel/pull/7/commits/6f8ee637969a055fc0cfbc9458b5053088141da0) which fixes a crash with Flixel dispatchers.
* https://github.com/FunkinCrew/lime/pull/24
  - Fixes compile errors when rebuilding the HashLink binaries.
* https://github.com/FunkinCrew/FlxPartialSound/pull/4 
  - Optional, fixes Freeplay previews not loading sometimes
* https://github.com/FunkinCrew/openfl/pull/7
* https://github.com/larsiusprime/polymod/pull/258
  - Fixes pretty much almost all scripts throwing errors.

### Known Issues
* `DropShadowShader` is utterly broken.
* Advancing in a Conversation causes a crash.
* Chart Editor may randomly crash while messing around.

PR #7103, opened 2026-03-08
[CRASHFIX] Fix crash with events on HashLink

NotHyper-474:
## Description
This was a fix that I had already included in #6135, but that for some odd reason magically disappeared after being merged.
It makes the type of certain local variables explicit, since the code generator for HashLink will infer that the type for `result` should be the same as the return type, when it should be a `Dynamic` or `Any` instead.


AbnormalPoof:
This is already fixed internally due to internal changes to song data but I'll merge this on public as well.

PR #7104, opened 2026-03-09
fix timing windows being 1 ms tighter than they should be

Rudyrue:
timing windows are actually 1 millisecond tighter than they should be, because it checks if the note difference is LESS than the max window
instead of less than or equal to

so `45 < 45` would be false, which returns a good
but if it's `45 <= 45` that would be true, which returns a sick
but if the note difference is `46 <= 45` then it's false, like normal
which in that case is more proper

### DIFF
--- a/source/funkin/play/scoring/Scoring.hx
+++ b/source/funkin/play/scoring/Scoring.hx
@@ -176,15 +176,15 @@ class Scoring
 
     return switch (absTiming)
     {
-      // case(_ < PBOT1_KILLER_THRESHOLD) => true:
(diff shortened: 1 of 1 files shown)


SammyTheGoog (approved):
This will save the game

nebulazorua:
game will be playable

spark-the-dragon:
finally funkin will be more playable

Rudyrue:
image  




man what ironic timing i just got this same issue

requazar (approved):

Hundrec:
What the heck

Marcosayfi:
THE GAME IS FINALLY PLAYABLE

PR #7109, opened 2026-03-09
[FIX] Improve Preloader time on HashLink

NotHyper-474:
## Description
For some reason, on C++ `Date.getTime` gives out a more precise resolution of a second, whereas in HashLink it is a whole number instead. For instance an `elapsed` that would be 1.50 in C++ would become just 1 in HashLink, and this would mean the Preloader graphics and state would only update every second.

## Screenshots/Videos

https://github.com/user-attachments/assets/466a577e-bb55-44b0-9ff5-e537cc576246


Hundrec (approved):
Thank you HotNyperLink

PR #7114, opened 2026-03-10
[FEATURE] Add `BitmapFilter` support to `FunkinSprite`

TechnikTil:
## Description
This PR adds the ability for `FunkinSprite`s to use `BitmapFilter`s.

`FlxFilteredSprite`s implementation has been rewritten to be more like how `flixel-animate` approaches its filters, which is more optimized and brings both Sparrow and Animate Atlas support.
This can either be used for something like the character select outline, or multiple shaders.

This PR also removes `FlxFilteredSprite`, since it's functionality is now merged into `FunkinSprite`.

## Video Showcase

https://github.com/user-attachments/assets/83a99249-f545-4406-bc86-91f07ea4e143


CrusherNotDrip:
image

realvirtu:
I wonder if this could be applied to the rimlight shader for atlas, so that rendertexture isn't needed? Unless that isn't possible with this.

TechnikTil:
> I wonder if this could be applied to the rimlight shader for atlas, so that rendertexture isn't needed? Unless that isn't possible with this.

This forces on render texture to work properly.

realvirtu:
> This forces on render texture to work properly.

Ah

PR #7140, opened 2026-03-13
2 new events added

KT-Gameplay:
Im unable to take screenshot as of now but i have added two new events
they are set stage event and my own event - scroll switch event


anysad:
It does look like they stole it.

Hundrec:
Please do not PR other people's code. 
As for an event that changes the scroll direction, we wouldn't want this in the game because it's a personal preference and doesn't work well as a game mechanic.

KT-Gameplay:
I understand

KT-Gameplay:
i didnt know that there was already an pr for the set stage event, and it
was abnormalPoof's stage changer script but i made it compatible with this
current fnf version and then made the pr

PR #7147, opened 2026-03-13
[BUGFIX] Fix Pause Menu global offset adjustment being FPS-dependent

KolzeYT:
## Description
Fixes the global offset changing being extremly fast when on high fps. Really noticable when playing on unlocked fps.


Raltyro:
oh hello kolze

JackXson-Real:
Oh so it really has been adjusting at different speeds this entire time. I thought I was going insane

Issue #6981, opened 2026-02-24
Bug Report: Chart Editor Easing Preview displays wrongly on tweens that go past min/max values


echolotl:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.8.2

### Description (include any images, videos, errors, or crash logs)

For tweens like the Elastic tween, it's in their design that they go past the minimum/maximum values set by the tween. There's a graphical error on the easing preview graph in the chart editor that causes it to not reflect this, and just seemingly clamps the value to the min/max value on both the graph and little preview dot animation.

<img width="332" height="171" alt="Image" src="https://github.com/user-attachments/assets/5bc39495-cbc6-4722-8b6c-161c3dae7b6f" />

Here's an example from https://www.easings.net/ for what it probably should look like
<img width="176" height="148" alt="Image" src="https://github.com/user-attachments/assets/b7a178c9-5443-4850-883f-373abdf9e5a5" />

Affected tweens include Elastic and Back.

### Steps to Reproduce

1. Open the Chart Editor and select an event w/ easing
2. Choose an affected tween as the Easing Type
3. Observe the clipping of the graph

Hundrec:
That makes sense!
@PurSnake Got any idea how to display these kinds of tweens?

PurSnake:
I think i know how i can make this better

PR #7152, opened 2026-03-14
[BUGFIX] Fix max/min display of Chart Editor easing preview on certain easings and use a consistent line thickness

echolotl:
## Linked Issues
#6981 

## Description
This fixes the display of easings that go past min/max values on the easing preview graph in the Chart Editor, and also makes the thickness of the line consistent.

I know PurSnake commented on the original issue that they would probably revisit this, and I can close this PR if they do in the future!!

## Screenshots/Videos

Before
2026-03-1322-52-06-ezgif com-video-to-gif-converter

After
2026-03-1323-14-19-ezgif com-video-to-gif-converter

### DIFF
--- a/source/funkin/play/event/SongEventHelper.hx
+++ b/source/funkin/play/event/SongEventHelper.hx
@@ -53,6 +53,49 @@ class SongEventHelper
     return easeBitmapMap.get(key);
   }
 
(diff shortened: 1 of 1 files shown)


PurSnake:
Ehh yeah that can go like that

Issue #6989, opened 2026-02-24
Bug Report: Reduced Performance with songs that use many events (like SPAGHETTI)


Untorope:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Android

### Browser

None

### Mobile Device Model

Samsung Galaxy A04s

### Mobile OS Version

Android 13

### Version

0.8.3

### Description (include any images, videos, errors, or crash logs)

So while i was playing the 0.8.3 patch to see if stuterring issues have funally been fixed, i've stumbed upon a lower framerate in Spaghetti (the song that uses a shitton of events), previously it was around 60 fps once the second sserafim started singing, but now its a stable 25-35 fps (idk how to describe it)

(my video doesnt upload for some reason, ill upload it as soon as i can!)

### Steps to Reproduce

1. have a bugdet range phone and open fnf
2. play spaghetti
3. notice you have a lower but stable fps

Untorope:
https://github.com/user-attachments/assets/1c15e605-f683-436d-b234-590d9fa375fb

finally!!
in the video it may seem like everythings fine but its not really like that in the game!

Untorope:
(im sorry for the horrible compression i had to do it since github wont allow my file at the original size)

PR #7171, opened 2026-03-17
[OPTIMIZATION] Optimize how Events are queried

KoloInDaCrib:
## Linked Issues
Contributes towards closing https://github.com/FunkinCrew/Funkin/issues/6989

## Description
This PR utilizes a system from the Strumline, but applied to the event querying process, so that there is a minimal amount of events being looped over and tanking the framerate.


KoloInDaCrib:
According to NotHyper-474, this is an FPS increase, though isn't a full fix for the linked issue, in which case I'll undraft the PR and unlink the issue.

PR #7180, opened 2026-03-19
Fix assets not being cleared on HashLink in general

NotHyper-474:
## Description
Fixes what 56d576700407b78000c6f88f8792140e7ea954ed thought was only an issue on MacOS


Hundrec:
Jigsaw made the same change before you did, but why not accept it anyway?

PR #7216, opened 2026-03-22
[OPTIMIZATION] Optimize quick sort implementation

HeroEyad:
So these two functions had a comment saying they were written with ChatGPT, and honestly I couldn't just leave them sitting there. Rewrote them myself and snuck in two optimizations while I was at it. (heh i love optimizating)

**`quickSortPartition`**: switched from picking the first element as the pivot to median-of-three (first, middle, last). The old way tanks to O(n²) on sorted or reverse-sorted input, which isn't exactly an edge case. As a bonus, sorting those three elements in place gives you boundary sentinels so the inner loop **can't** overshoot.

**`quickSortInner`**:  added a fallback to insertion sort for partitions under 16 elements. Quicksort has too much overhead to be worth it at that size, insertion sort just wins there. This is what _Java_ and _Python_ do under the hood too, so feels like a safe call.

In Java:  [Arrays.sort()](https://www.geeksforgeeks.org/java/arrays-sort-in-java/) uses dual-pivot Quicksort for primitives, and insertion sort is used for small datasets, typically under 47 elements

In Python: [Timsort](https://www.geeksforgeeks.org/dsa/tim-sort-in-python/) is Python's default sorting algorithm (``sorted()``, ``list.sort()``). It sorts small runs shorter than a fixed size (usually 32) using insertion sort, since it's fast for small or nearly sorted data.

### DIFF
--- a/source/funkin/util/tools/ArraySortTools.hx
+++ b/source/funkin/util/tools/ArraySortTools.hx
@@ -44,21 +44,24 @@ class ArraySortTools
 
   /**
    * Internal recursive function for the quick sort algorithm.
-   * Written with ChatGPT!
+   * Recurses on the smaller partition and loops on the larger one
+   * to keep stack depth at O(log n) in the worst case.
    */
   static function quickSortInner<T>(input:Array<T>, low:Int, high:Int, compare:CompareFunction<T>):Void
   {
-    // When low == high, the array is empty or too small to sort.
-
-    // EDIT: Recurse on the smaller partition, and loop for the larger partition.
     while (low < high)
     {
-      // Designate the first element in the array as the pivot, then partition the array around it.
-      // Elements less than the pivot will be to the left, and elements greater than the pivot will be to the right.
-      // Return the index of the pivot.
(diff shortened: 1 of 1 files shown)


Trayfellow:
Funkin' is no longer vibe-coded medias

image

FuroYT:
As much as I don't really like AI code at least it was commented as made by AI which is better than making it your own code when you didn't

But now it should be gone for optimisations

AnimatingLegend (approved):
this is so Image

realvirtu:
Thank you for removing some of the AI stuff. FNF should be AI free honestly.

notkaif:
> Thank you for removing some of the AI stuff. FNF should be AI free honestly.

Yes remove Senpai from the game immediately

AbnormalPoof:
Please fix the conflicts!!!!!!!! WAAAAAAAHHHHHHHHHHHH!!!!!!!!!!!!!! WAHHHHHHHHHH!!!!!!!!!!!!!!

PR #7231, opened 2026-03-25
[FEATURE] Add a "Bop Camera Once" Song Event

TechnikTil:
## Description
This PR adds a new song event titled <i><b>Bop Camera Once</b></i>. It lets you bop the camera in at any point, no matter the step.
This can be useful if there is a specific point in your song that just makes your head bop, and you want the game to reflect that.

If you have ever used Psych Engine before, this is probably ringing a bell.
"No way Til..."
Yes way me! It's the "Add Camera Zoom" event!

It works relatively the same than it's Psych Engine counterpart, but with some small changes to make it feel more like "home" to other song events that this game has. 

## In the Editors
Let's take a look at it in the Chart Editor, shall we?

image

In my opinion, this should be very self-explanatory.
"Game" is the camera containing the Stage and characters, "HUD" is what contains the health bar and notes.
The intensity is multiplied by the base intensity, then applied to the camera. 

"What do you mean by base intensity Til?"
Good thing you're asking! The "base intensity" I'm talking about is the current intensity set for the regular Camera Bop, the camera bop that happens every measure. This is also known as the <i><b>Set Camera Bop</b></i> Intensity.

## Showcase
Here's an example of the event doing its thing:

https://github.com/user-attachments/assets/6f04d08d-8849-4d3c-95ce-eef1aca897ee

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1852,8 +1852,8 @@ class PlayState extends MusicBeatSubState
       && cameraZoomRate > 0
       && (Conductor.instance.currentStep + cameraZoomRateOffset * Constants.STEPS_PER_BEAT) % (cameraZoomRate * Constants.STEPS_PER_BEAT) == 0)
     {
-      // Set zoom multiplier for camera bop.
-      cameraBopMultiplier = cameraBopIntensity;
+      // Add zoom multiplier to camera bop.
+      cameraBopMultiplier += cameraBopIntensity - 1;
       // HUD camera zoom still uses old system. To change. (+3%)
       camHUD.zoom += hudCameraZoomIntensity * defaultHUDCameraZoom;
     }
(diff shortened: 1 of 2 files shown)


Starexify:
ah yes, bopeebo always makes my head bop when he hits that hold note

Honton129:
I'm afraid it's peak...
I always had an issue with 3/4 sections and the zoom frequency but now with this I can cook my cool zoom events!

Hundrec:
I think this name is too similar to the Set Camera Bop Speed event. 
It should be named "Bop Camera Once" to distinguish it from other similar sounding events.

ender69flock:
BEST PR

Hundrec:
True...

AbnormalPoof (approved):

ender69flock:
FINALLY!!!!!1211`111111!!!!!

Honton129:
YOOO WE WON

PR #7238, opened 2026-03-26
[OPTIMIZATION] Recycle script events

KoloInDaCrib:
## Linked Issues
None.

## Description
This PR implements a pooling system for script events, making the garbage collector get triggered way less frequently. There's a crap ton of changes because I've had to change **every single time** a script event is dispatched to accomodate for this.

This implementation might look a bit confusing, so let me explain:
A lot of read-only event fields had to be changed to `(default, set)` with an empty setter like this
```haxe
function set_field(value:Bool) return value;
```
This makes the field unable to be set, unless called with `@:bypassAccessor`, which negates the functionality of the setter. This is perfect, because in scripts we can't use it (and neither does reflection really work with it) but we can in source, making us be able to re-use the event instance while only changing the field values.

Also I've *slightly* changed whether some events are cancelable or not. It's mostly the `CREATE` events being cancelable, despite no logic being implemented to account for this.

### DIFF
--- a/source/funkin/play/Countdown.hx
+++ b/source/funkin/play/Countdown.hx
@@ -3,6 +3,7 @@ package funkin.play;
 import flixel.tweens.FlxEase;
 import flixel.tweens.FlxTween;
 import funkin.modding.events.ScriptEvent;
+import funkin.modding.events.ScriptEventType;
 import funkin.modding.events.ScriptEvent.CountdownScriptEvent;
 import flixel.util.FlxTimer;
 import funkin.util.EaseUtil;
@@ -107,23 +108,22 @@ class Countdown
    */
   static function propagateCountdownEvent(index:CountdownStep):Bool
   {
-    var event:ScriptEvent;
-
-    switch (index)
+    var type:ScriptEventType = switch (index)
     {
       case BEFORE:
-        event = new CountdownScriptEvent(COUNTDOWN_START, index);
-      case THREE | TWO | ONE | GO: // I didn't know you could use `|` in a switch/case block!
-        event = new CountdownScriptEvent(COUNTDOWN_STEP, index);
+        ScriptEventType.COUNTDOWN_START;
       case AFTER:
-        event = new CountdownScriptEvent(COUNTDOWN_END, index, false);
+        ScriptEventType.COUNTDOWN_END;
       default:
-        return true;
+        ScriptEventType.COUNTDOWN_STEP;
     }
+
+    var event:CountdownScriptEvent = CountdownScriptEvent.get(type, index, (index != AFTER));
 
     // Modules, stages, characters.
     @:privateAccess
     PlayState.instance.dispatchEvent(event);
+    event.put();
 
     return event.eventCanceled;
   }
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -548,7 +548,6 @@ class PlayState extends MusicBeatSubState
   var discordRPCAlbum:String = '';
   var discordRPCIcon:String = '';
   #end
-
   /**
    * RENDER OBJECTS
    */
@@ -1058,13 +1057,14 @@ class PlayState extends MusicBeatSubState
 
       prevScrollTargets = [];
 
-      var retryEvent = new SongRetryEvent(currentDifficulty);
+      var retryEvent = SongRetryEvent.get(currentDifficulty);
 
       previousDifficulty = currentDifficulty;
 
       currentStage?.resetStage();
 
       dispatchEvent(retryEvent);
+      retryEvent.put();
 
       resetCamera();
 
@@ -1280,8 +1280,9 @@ class PlayState extends MusicBeatSubState
         Events.logFailSong(currentSong.id, currentVariation);
         #end
 
-        var event:ScriptEvent = new ScriptEvent(GAME_OVER, true);
+        var event:ScriptEvent = ScriptEvent.get(GAME_OVER, true);
         dispatchEvent(event);
+        event.put();
 
         if (event.eventCanceled) return;
 
@@ -1372,16 +1373,18 @@ class PlayState extends MusicBeatSubState
       case Conversation:
         preparePauseUI();
 
-        final event = new PauseScriptEvent(false);
+        final event = PauseScriptEvent.get(false);
         dispatchEvent(event);
+        event.put();
 
         if (!event.eventCanceled) openPauseSubState(Conversation, camPause, lostFocus, () -> currentConversation?.pauseMusic());
 
       case Cutscene:
         preparePauseUI();
 
-        final event = new PauseScriptEvent(false);
+        final event = PauseScriptEvent.get(false);
         dispatchEvent(event);
+        event.put();
 
         if (!event.eventCanceled) openPauseSubState(Cutscene, camPause, lostFocus, () -> VideoCutscene.pauseVideo());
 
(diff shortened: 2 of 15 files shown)


DustedDustin:
Holy shit this may make lagspikes and stutters even less common than they were already especially in weaker hardware

source/funkin/play/Countdown.hx
```diff
@@ -107,20 +109,16 @@ class Countdown
...
-      case AFTER:
-        event = new CountdownScriptEvent(COUNTDOWN_END, index, false);
-      default:
-        return true;
+      case BEFORE: ScriptEventType.COUNTDOWN_START;
+      case AFTER: ScriptEventType.COUNTDOWN_END;
+      default: ScriptEventType.COUNTDOWN_STEP;
     }
 
+    var event:CountdownScriptEvent = cast ScriptEventDispatcher.recycleEvent(CountdownScriptEvent, type, (index != AFTER));
```
EliteMasterEric: It really bugs me that this is a cast. I wonder if we can somehow make this a generic function considering that I don't expect events to ever be dispatched from a script.

source/funkin/modding/events/ScriptEventDispatcher.hx
```diff
@@ -10,6 +10,42 @@ import funkin.modding.module.Module;
...
+   * Recycles the event from the event pool. This prevents the garbage collector from being called a ton!
+   * @param eventClass  The event class that should be recycled.
+   * @param type        The event type.
+   * @param cancelable  Whether the event can be canceled by scripts.
+   */
+  public static function recycleEvent(eventClass:Class<ScriptEvent>, type:ScriptEventType, cancelable:Bool = false):ScriptEvent
+  {
+    var result:Null<ScriptEvent> = null;
+
+    for (event in eventPool)
```
EliteMasterEric: I was going to ask if iterating over every event in the pool with each call would cause problems, but presumably there'd only be one event of each class in the pool, since right now we can assume events won't be dispatched in parallel. 👍

EliteMasterEric (changes requested):
Might be nice to make the code for this PR cleaner. I'm not the biggest fan of littering a bunch of classes with `@:bypassAccessor` calls.

Maybe we can add metadata that Polymod can see which makes the field effectively final?

KoloInDaCrib:
> Maybe we can add metadata that Polymod can see which makes the field effectively final?

Either that or we get rid of read-only fields. The reason for that is that, from my testing, Reflect ignores the `null` property, allowing scripts to set what would otherwise be read-only variables. Admittedly it's not ideal and I do agree that some sort of macro could be nice - maybe something similar to the defunct `PolymodFinalMacro` that instead checks for a metadata?

source/funkin/modding/events/ScriptEventDispatcher.hx
KoloInDaCrib: Yeah from some tracing it seems that there's about 8 events in the pool by the time the player enters the Freeplay Menu, which, while not perfect, is marginally less costly than constantly filling up the garbage with events.

Starexify:
> > Maybe we can add metadata that Polymod can see which makes the field effectively final?
> 
> Either that or we get rid of read-only fields. The reason for that is that, from my testing, Reflect ignores the `null` property, allowing scripts to set what would otherwise be read-only variables. Admittedly it's not ideal and I do agree that some sort of macro could be nice - maybe something similar to the defunct `PolymodFinalMacro` that instead checks for a metadata?

Why not just add the fields to blacklist ?

KoloInDaCrib:
> Why not just add the fields to blacklist ?

Because the field blacklist prevents field setting AND getting, which isn't the right solution, especially when a lot of mods depend on them.

Starexify:
> > Why not just add the fields to blacklist ?
> 
> Because the field blacklist prevents field setting AND getting, which isn't the right solution, especially when a lot of mods depend on them.

Then why not use never ? iirc reflect doesnt bypass that
Also would like an example of what you mean exactly

KoloInDaCrib:
I've made the `recycleEvent` function generic, removing a lot of casting. It worked well on Hashlink so I think it's safe to assume it would work well on other targets too.
As for the read-only field implementation I'll leave that for debate since I've brought up some ideas in a reply above.

EliteMasterEric:
If FunkinCrew/Polymod#415 gets merged, scripts will respect null setters on properties, meaning we can make these properties modifiable only through a private `recycle` method.

KoloInDaCrib:
Tweaked the system again to make every event class include a pool system through a macro. That way, the old method of creating events is intact and the event recycling process now accounts for some fields being read-only.

source/funkin/play/PlayState.hx
```diff
@@ -1044,13 +1044,14 @@ class PlayState extends MusicBeatSubState
...
 
-      var retryEvent = new SongRetryEvent(currentDifficulty);
+      var retryEvent = SongRetryEvent.get(currentDifficulty);
 
       previousDifficulty = currentDifficulty;
 
       currentStage?.resetStage();
 
       dispatchEvent(retryEvent);
+      retryEvent.put();
```
EliteMasterEric, line 1067: Now nearly every use of dispatchEvent has this `event.put()` call! What is this for and can we get rid of it?

EliteMasterEric (changes requested):

source/funkin/play/PlayState.hx
AbnormalPoof: This has the same functionality as `FlxPoint` where it just returns the event to the pool

KoloInDaCrib: > This has the same functionality as `FlxPoint` where it just returns the event to the pool

This ^, but the event putting also serves an important function of signifying if a recycled event can be used again within the same tick.

While testing with one of my mods, I found out that the script responsible for state redirection got borked because I was calling `FlxG.switchState` in `onStateChangeBegin` after cancelling its event. Normally this wouldn't be an issue since an event gets created for every time `startOutro` is called, however in this PR where the script events are recycled, the system would reuse the previously cancelled event, reset its properties (i.e. mark it as "not cancelled") and therefore essentially negate the redirection attempt performed by the script.

With the putting system, this never happens, as the game checks during the tick if the event is currently in the progress of being dispatched, and if it is and we need to dispatch it again, it creates a new replica of the event and adds it to the pool. 

I documented the `hasBeenUsed` variable of ScriptEvent to mention this, though obviously not in detail.

source/funkin/play/cutscene/dialogue/Conversation.hx
```diff
@@ -312,7 +318,16 @@ class Conversation extends FunkinSpriteGroup implements IDialogueScriptedClass i
...
+    var event:DialogueScriptEvent = DialogueScriptEvent.get(DIALOGUE_START, this, true);
+    dispatchEvent(event);
+    event.put();
+  }
+
+  public function endConversation():Void
+  {
+    var event:DialogueScriptEvent = DialogueScriptEvent.get(DIALOGUE_END, this, true);
+    dispatchEvent(event);
+    event.put;
```
NotHyper-474: I only just realized this line. It's apparently valid syntax, but it is not actually calling the function.

NotHyper-474 (changes requested):

source/funkin/play/cutscene/dialogue/Conversation.hx
AbnormalPoof: lmao????

AbnormalPoof (approved):

DustedDustin:
WWWWWWWWWWWWWWWWWW

PR #7265, opened 2026-03-28
[ENHANCEMENT/Modding] Move SongEvents to scripts.

Starexify:
## Associated funkin.assets PR
- https://github.com/FunkinCrew/funkin.assets/pull/402

## Description
This PR just moves the SongEvent from hardcoded to scripts (softcoded).
Mostly to allow modders to override the events easier via mixins and keep the original scheme.


EliteMasterEric:
For the purposes of optimization and for preventing breaking changes, the base game's built-in song events should all stay as source code.

If you are sure you want to override these events, and are fine with making changes that may break other mods, you can do something like the following:

```haxe
package mymod;

class FocusCameraSongEventOverrideModule extends Module {
  public function new() {
    super('mymod-override-focuscamera', 1000, {
      state: PlayState,
    });
  }

  override public function onSongEvent(event:SongEventScriptEvent):Void
  {
    var eventKind:String = event.eventData.eventKind;

    if (eventKind == 'FocusCamera') {
      // Prevent default behavior
      event.cancel();

      // Perform custom behavior here
      var eventValue:String = event.eventData.value;
      trace('FocusCamera event triggered with value: ' + eventValue);
    }
  };
}
```

PR #7275, opened 2026-03-29
[CLEANUP] Remove Useless Function in the MainMenuState

JVNpixels:
nvm so turns out i forgot that it was a getter and its 1:30 am and my brain is cooked


ExtraCode75:
reminds me of coolColors array. R.I.P old freeplay menu

PR #7319, opened 2026-04-02
[BUGFIX] Fix newly created charts improperly setting the voice list

KoloInDaCrib:
## Linked Issues
None.

## Description
The issue this PR aims to fix is that newly made charts that use players that aren't bf and opponents that aren't dad won't properly load vocal files in the chart editor. This is because the newly created data for characters in the song metadata files uses default characters, which then also fill the player/opponent voices array. In the chart editor, the aforementioned arrays are **never changed**, meaning that the metadata generated by the editor will always attempt to use dad and bf voices, even for charts that don't have them as characters.


TechnikTil:
Merge conflicts!

PR #7321, opened 2026-04-02
[FIX + OPTIMIZATION] Improve `onExit` cleanup & resource disposal 

### LINKED ISSUE #7235: Bug Report: Application onExit Event doesn't run when closing the game
### Issue Checklist

- [ ] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Mobile Device Model

_No response_

### Mobile OS Version

_No response_

### Version

0.8.4

In 0.8.4, if you try having a function run when the game closes using `WindowUtil.windowExit` or similar methods it will not run.

Traces in 0.8.4 / 0.8.3 Dev Build

Image

Traces in Release 0.8.3

Image

This is the code that both versions use

Image

This is most likely caused by the `Sys.exit(0)` in `Main.hx` (pointed out by Raltyro)

Image

### Steps to Reproduce

1. Have a script that adds to `WindowUtil.windowExit` or `Application.current.onExit` with traces
2. Run the game in a terminal and  try closing the game
3. Notice the missing traces that should've been called

AnimatingLegend:
## Linked Issues
- #7235 

## Description
I just wanted to make some minor improvements to the background processing PR i made for v0.8.4, aswell as fix some issues that involve Mods not being able to properly run `onExit`.

### Changes
- The game now disposes of cached audio / textures, alongside `openfl.Assets` to ensure a proper resource cleanup.
  * Disposing of cached audio / textures upon exiting the game will now be logged to the terminal.
- Set `onExit` handler priority to 99 to ensure cleanup runs after mod exit callbacks.
- Removed HTML5 builds from getting background processed to prevent compilation errors.
- Removed `hxvlc` disposal from background processing.

### DIFF
--- a/.github/actions/setup-haxe/action.yml
+++ b/.github/actions/setup-haxe/action.yml
@@ -5,7 +5,7 @@ inputs:
   haxe:
     description: 'Version of haxe to install'
     required: true
-    default: '4.3.6'
+    default: '4.3.7'
   hxcpp-cache:
     description: 'Whether to use a shared hxcpp compile cache'
     required: true
@@ -71,7 +71,7 @@ runs:
 
   - name: Restore cached dependencies
     id: cache-hmm
-    uses: actions/cache@v4
+    uses: actions/cache@v5
     with:
       path: .haxelib
       key: haxe-hmm-${{ runner.os }}-${{ hashFiles('**/hmm.json') }}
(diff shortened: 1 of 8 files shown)


ComedyLost (approved):
Fixed the issue I had 👀

KarimAkra:
The `Sys.exit` is blocking the mods' exit callback because this event is added before the `WindowUtil`'s event
Even if with removing it, this doesn't seem like a correct behavior to me if we're purging everything THEN dispatching the mods' exit function. The purge should be the last thing that gets executed before the whole apps exits.

You can fix this by setting the event's priority to something like 99 so it will come after all the other events
Basically something like this

`onExit.add(() -> {..}, 99);`

And i suggest adding back the `Sys.exit` call.

MAJigsaw77:
I think it would be better for LibVLC's disposal to be done inside hxvlc, not inside funkin

AnimatingLegend:
Just got around to updating this PR, everything should be good now. 👍🏽

NotHyper-474:
I feel like changing the check to only `!mobile` is gonna cause a build error on web since `Sys` does not exist there.

Issue #5793, opened 2025-08-22
Bug Report: Note hit input latency added when notes are hit late


Lasercar:
### Issue Checklist

- [X] I have read the [Contributing Guide](<https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md>)
- [X] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [X] I have properly titled my issue
- [X] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Version

0.7.3

### Description (include any images, videos, errors, or crash logs)

So, I noticed that the input latency is now added when a note is hit late:<br>[https://github.com/FunkinCrew/Funkin/blob/059b412e85c2f9ae528af57d638c79734886667f/source/funkin/play/PlayState.hx#L2781-L2793](<https://github.com/FunkinCrew/Funkin/blob/059b412e85c2f9ae528af57d638c79734886667f/source/funkin/play/PlayState.hx#L2781-L2793>)

Turns out, this is the exact same change I did a few months ago, that was rejected with this as the issue with it (well, that and changing when the conductor gets updated in `update()` but anyway):

[https://github.com/FunkinCrew/Funkin/pull/4297/files#diff-d735da8a36bf5a753af53c828f8a2213a4cd5bf5bad648e262c3391f9fb8f442R2527-R2530](<https://github.com/FunkinCrew/Funkin/pull/4297/files#diff-d735da8a36bf5a753af53c828f8a2213a4cd5bf5bad648e262c3391f9fb8f442R2527-R2530>)

Quote from eric:

> `inputLatencyMs` is the time between when the note was hit and the current time. For example, if the note was hit at 10500 ms and it takes 10 ms to get around to processing the note, the current time will be 10500 ms and `inputLatencyMs` will be 10 ms. You should ALWAYS subtract it, because otherwise the input delay will be DOUBLED.

So, I'm quite confused as to why this bad change is now in the base game...

Hundrec:
You'll need to provide evidence of this causing unintended behavior in the game
Looking at the code isn't enough to tell that something is wrong.

Lasercar:
Oh, forgot to @EliteMasterEric

PR #7367, opened 2026-04-12
[BUGFIX] Fix note input latency being doubled when hit late

Lasercar:
## Linked Issues
Fixes https://github.com/FunkinCrew/Funkin/issues/5793

## Description

The input latency is already included in every input hit. All we have to do is calculate and remove it to get the actual time the note was hit.

While adding input latency when a note is hit after it's timestamp does sometimes turn late bad note hits into better ones, it doubles the input latency and shifts the hit further away from where the player actually hit it (something we do not want to be doing or causing to happen).


Kade-github (approved):
yeah that makes sense

PR #7374, opened 2026-04-14
[CHORE] Improve null-safety for character handling in `PlayState`

NotHyper-474:
## Linked Issues
None, probably.

## Description
Marks some functions in `Stage` as nullable so that the compiler is able to yell at your freaking face if you fail to properly null-check those (at least in cases where null-safety is on). Fixes and prevents future crashes with `None` characters in PlayState.

## Screenshots/Videos
N/A


NotHyper-474:
Just realized https://github.com/FunkinCrew/Funkin/pull/7259 fixes the same crash :eric_whoopsie: 

Well, this would still be a good change anyway since it'd prevent future edits to the code from forgetting that characters can be `null` and creating new crashes

PR #7416, opened 2026-04-20
Fixed comments in NoteSprite.hx

Gesee-y:
## Description

Fixed comments in NoteSprites.hx for public variable `tooEarly`, `hasMissed` and `mayHit`, Those were saying that note hit window was frame depend which is false and misleading as they now use the constant `HIT_WINDOW_MS`


Hundrec:
Thanks for adjusting these comments, but Eric has made changes to the comments internally, now on the `develop` branch.

PR #7421, opened 2026-04-22
[BUGFIX] Fix Camera Editor window title starting as null

ComedyLost:
## Description
Changes how the Camera Editor sets the window title to be the same as how the Chart Editor does it.
This fixes an issue where when you first enter the state the title will show a null.

## Screenshots/Videos
Before:

image

After:

image


Hundrec (approved):
null...

Image

PR #7446, opened 2026-04-23
[OPTIMIZATION] Change `FunkinGroup.updateChildren` to use the FlxPoint pool

MightyTheArmiddilo:
## Description
For some reason I had it make a new FlxPoint every time updateChildren ran so...


TechnikTil (commented):
It would probably be better to reuse the same point throughout the entire instance instead of pulling from the pool every update.

You could also use `FlxSprite._point`, which is supposed to be used for rendering. Since we don't really render, why not use it for something else?

MightyTheArmiddilo:
Superceded by/merged into #7753

PR #7473, opened 2026-04-25
[BUGFIX] Fix `getScriptedEntryClassName` package oversight

TechnikTil:
## Description
This PR fixes an oversight done in `SongRegistry.getScriptedEntryClassName` that wouldn't return the proper path if the script mentioned included a package. In return, this would not let any packaged song scripts be playtested in the chart editor, and throw an error.


TechnikTil:
Duplicate of #7470
Can't disrespect the one line banger

PR #7476, opened 2026-04-27
[BUGFIX] Fix Nametag being mispositioned on widescreen resolutions

NotHyper-474:
## Linked Issues
https://discord.com/channels/1419268615605325907/1419276431032189038/1497990894521946252

## Description
This fixes an issue where, not only the nametag would not appear in the correct position, but it would also not play its intro animation.
This was caused by a refactor done in a `develop` commit to make the nametag use a singular variable for positioning instead of two. The problem is that it was done under the assumption that simply applying a setter for this variable would be enough for an assignment to a property (i.e `midpoint.x = ...`) to trigger `updatePosition`, but it'd actually require doing so to the whole variable (i.e `midpoint = ...`).
The fix? Throwing all this system off the window, we instead use `midpoint` to offset the position when we render the nametag.

### Before
https://github.com/user-attachments/assets/7a7281d7-4128-421a-9c96-7b6b4cab2335
### After
https://github.com/user-attachments/assets/841528b0-0ebc-4d4b-8d67-8948e53a16ca

### DIFF
--- a/source/funkin/ui/charSelect/CharSelectSubState.hx
+++ b/source/funkin/ui/charSelect/CharSelectSubState.hx
@@ -687,7 +687,7 @@ class CharSelectSubState extends MusicBeatSubState
     FlxTween.tween(cursors, {alpha: 0}, 0.8, {ease: FlxEase.expoOut});
 
     FlxTween.tween(barthing, {y: barthing.y + 80}, 0.8, {ease: FlxEase.backIn});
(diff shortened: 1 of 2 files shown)


Hundrec:
Aw hell yeah NotHyper-474 you da goat!

Hundrec:
Me when I merge directly into `develop`

NotHyper-474:
> Me when I merge directly into `develop`

Billions must now rebase :trollface:

PR #7497, opened 2026-04-29
[BUGFIX] Fix TextureSwap not setting the variables

ComedyLost:
## Linked Issues
MEEEEEEEE

## Description
This pr sets the variables correctly in the setters, which fixes the variables not giving the right values when trying to use them (like when trying to tween `amount`).


AnimatingLegend (approved):
2 line banger like usual :jigsaw~1:

PR #7508, opened 2026-05-01
[CHORE] Use using keyword in the CameraEditor to cleanup code in it.

T5mpler:
Utilizes the using keyword to cleanup some code in the CameraEditor. I have no idea why the keyword isn't used when it's properly set up to be used for the Camera Editor.


T5mpler:
K wait, have to re-open the PR again because I just realized this PR uses a different version of CameraEditor.hx from before develop & preview/camera-editor was force merge, sorry for the ranking closed PRs I keep doing this is annoying to deal with

PR #7510, opened 2026-05-01
[ENHANCEMENT] Allow ESCAPE key to exit editors when first opened

T5mpler:
This PR allows the user to use the ESCAPE key for if they want to exit any editors firsthand when opening them.

While the F4 key does exist to automatically return to the Main Menu, a lot of people generally don't know first hand that this keybind exists, leading you to believe that you're unable to exit the editor at all when opening it. This is especially bad for the Camera Editor as the user could believe they're softlocked if they don't have any .fnfc files to continue on, and DON'T know about the F4 key.


TechnikTil:
This should be closed due to the Camera Editor not accepting enhancements at the moment.
We also have merge conflicts here!

Hundrec:
We've added loading from song templates now, so this won't be a concern.

PR #7512, opened 2026-05-02
[VULNERABILITY] Blacklist `haxe.Http`

charlesisfeline:
## Description
self-explanatory, `haxe.Http` is literally just an alias for `sys.Http`, which is already... blacklisted...

(kinda regret doing this, but had to do it outta instinct im sorry💔)

### DIFF
--- a/source/funkin/modding/PolymodHandler.hx
+++ b/source/funkin/modding/PolymodHandler.hx
@@ -328,6 +328,10 @@ class PolymodHandler
     // Lib.load() can load malicious DLLs
     Polymod.blacklistImport('cpp.Lib');
 
+    // `haxe.Http`
+    // An alias for `sys.Http`, which is also a blacklisted package.
+    Polymod.blacklistImport('haxe.Http');
+    
     // `haxe.Unserializer`
     // Unserializer.DEFAULT_RESOLVER.resolveClass() can access blacklisted packages
     Polymod.blacklistImport('haxe.Unserializer');


VirtuGuy:
image

AbnormalPoof:
This needs to be rethought. `sys.Http` on its own isn't able to do anything meaningfully harmful. The reason the entire `sys` package is blacklisted because it includes many classes for the filesystem and running processes.

Accessing the internet isn't really harmful unless it downloads a malicious file, but even then you need something to actually **run** said file, and all of those methods have been blacklisted.

VirtuGuy:
image

Kade-github:
> This needs to be rethought. `sys.Http` on its own isn't able to do anything meaningfully harmful. The reason the entire `sys` package is blacklisted because it includes many classes for the filesystem and running processes.
> 
> Accessing the internet isn't really harmful unless it downloads a malicious file, but even then you need something to actually **run** said file, and all of those methods have been blacklisted.

"www.malcioussite.com/hack_yo_account.php?discordToken=mahpassword"

Kade-github:
if sys.http is already blocked, this should also be blocked! Will hand this to Eric for thought though

FuroYT:
> "www.malcioussite.com/hack_yo_account.php?discordToken=mahpassword"

My main concern is how would you get the token as the file system access is locked to the game's folder not full access to the disk and since discord token is stored in app data unless you install fnf there (if you do that's on you)

And also if you mention a external program to do it in the path, well you can't just execute a program from the game anymore from FileUtil

AbnormalPoof:
> "www.malcioussite.com/hack_yo_account.php?discordToken=mahpassword"

This might be parroting what Furo said, but for this to even work at all you'd need to get their Discord token!

From my understanding there's no realistic way to get someone's Discord token using the classes the game provides... at least without a blacklist bypass, but those are exceptionally rare nowadays.

NotHyper-474:
> if sys.http is already blocked, this should also be blocked! Will hand this to Eric for thought though

Overall I'm neutral about the whole situation, but it's possible that was only because whatever would allow for file access is part of the same package; [the comment for the blacklist](https://github.com/FunkinCrew/Funkin/blob/main/source%2Ffunkin%2Fmodding%2FPolymodHandler.hx#L476-L477) seems to corroborate this idea.

Hundrec (approved):
Eric said this should be blacklisted.

flashfreako:
image

PR #7518, opened 2026-05-03
[BUGFIX + OPTIMIZATION] Fix TimelineShader AMD Crash and optimize calculations

PurSnake:
## Linked Issues
image

## Description
Slightly rewrites TimelineShader to make it actually work for all amd users.
With a touch of optimization.

## Screenshots/Videos
nan


PurSnake:
Neat

PR #7528, opened 2026-05-07
[BUGFIX] Unfocus Chart Editor dropdown after selecting item

### LINKED ISSUE #2956: Bug Report: When opening the events window, playtest cannot be activated and the screen cannot be scrolled until an event is clicked, or you close the window.
## Describe the bug

If you're in chart editor and you open up the events window, you will not be able to scroll, place notes and events, or playtest the game until you click on an event. If no such event is present on screen already, you then have to close out of the menu, scroll down to find one, open it up again, click on the event, and then you can go about your event placing day. The thing is, it doesn't start happening immediately, it takes a while of messing around and being in chart editor before it starts happening.
## To Reproduce

I can't seem to figure out if there is some trigger that causes it to start happening or if it's just from me leaving the game open for a while in chart editor, but it will always start happening eventually if you are charting for long enough and the only way to stop it is by restarting the game.
## Expected behavior

I would hope that I could start placing events right away and not be forced to back out and go find a different one first.
## Screenshots/Video

Not sure how I could show that I'm _not_ able to scroll in video, but you get the idea. Nothing works.
## Desktop
 - OS: Windows 10
 
 - Browser: N/A
 
 - Version: 0.4.1

VICK-sys:
## Linked Issues
Fixes #2956

## Description
  Fixes the chart editor input lockout that occurred when a HaxeUI toolbox field had focus (blue outline). After interacting with a field, playtesting with the toolbox open,
  and reopening the toolbox, the field's focus was auto-restored, and `isHaxeUIFocused` returning true gated all chart editor input (scroll, note placement, playtest hotkey,
   etc.).

  **The fix:** in `ChartEditorToolboxHandler.showToolbox` and `.hideToolbox`, directly call `IFocusable.focus = false` on the currently-focused element via
  `FocusManager.instance.focus`. This bypasses HaxeUI's `set_focus(null)` API, which has an asymmetric implementation: `get_focus` walks all root components, but
  `set_focus(null)` only clears `_lastFocuses` for `Screen.instance.topComponent`, so focus living on a non-top root (like a dialog) can't be cleared through it.

## Screenshots/Videos

https://github.com/user-attachments/assets/d429120b-addd-4c78-9b8d-7d08d168bcbb

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -3793,6 +3793,13 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
     handleMusicPlayback(elapsed);
     handleNoteDisplay();
 
+    if (isHaxeUIFocused
+      && !isCursorOverHaxeUI
+      && (FlxG.mouse.justPressedRight || FlxG.mouse.deltaWheel.y != 0))
+    {
+      ChartEditorToolboxHandler.clearHaxeUIFocus();
+    }
+
     // These ones only happen if the modal dialog is not open.
     handleScrollKeybinds();
     handleCursor();
--- a/source/funkin/ui/debug/charting/handlers/ChartEditorToolboxHandler.hx
+++ b/source/funkin/ui/debug/charting/handlers/ChartEditorToolboxHandler.hx
@@ -35,6 +35,14 @@ class ChartEditorToolboxHandler
     }
   }
 
+  public static function clearHaxeUIFocus():Void
+  {
+    @:nullSafety(Off) {
+      final f = FocusManager.instance.focus;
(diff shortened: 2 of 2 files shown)


Hundrec (changes requested):
This bug has bothered me forever!

The PR's behavior is a big improvement, as it allows left clicks to unfocus the dropdown.
However, right clicking and scrolling is still gated until you left click. 

The PR would be perfect if you fixed this too!

VICK-sys:
> This bug has bothered me forever!
> 
> The PR's behavior is a big improvement, as it allows left clicks to unfocus the dropdown. However, right clicking and scrolling is still gated until you left click.
> 
> The PR would be perfect if you fixed this too!

I'm on it

Hundrec (approved):
It's finally fixed... after all this time

Issue #2262, opened 2024-05-03
Enhancement: Update Github version and make the game check for update using it


SomeoneUnknownForReal:
## What is your suggestion, and why should it be implemented?
1. Add the newest 2 updates to github.
2. Add a wiki for making mods using source and without (engines mostly write it without mentioning source so like you have to be better).
3. Make it so the game checks for updates using github (helpful source: Psych Engine and I think it was title state or flash state)
4. Make it so the game can update itself (check the yoshi engine archive on github, it has the code for updating itself).

SilverT4:
1. https://twitter.com/EliteMasterEric/status/1785731651839942819

Lasercar:
Duplicate of https://github.com/FunkinCrew/Funkin/issues/2496

Hundrec:
I don't think these issues are related.
Also, this one is older.

Starexify:
No joke, this would be annoying to see each time you open an older version

PR #4456, opened 2025-04-01
[FEATURE] Re-add out of date screen

TechnikTil:
## Does this PR close any issues? If so, link them below.
#2496, #2262

## Briefly describe the issue(s) fixed.
This adds the out of date screen mentioned in the issue above.

## Include any relevant screenshots or videos.
image
(the only reason it says z and x is because thats what getDialogueNameFromControl returns)

PR #7538, opened 2026-05-08
[ENHANCEMENT] Version check banner

VICK-sys:
## Linked Issues
Implements item 3 of #2262

## Description
Once per session on `TitleState`, the game pings `api.github.com/repos/FunkinCrew/Funkin/releases/latest`, parses
`tag_name`, and compares against `Application.current.meta.get('version')` via `thx.semver`. If the release is newer, a
semi-opaque dark banner with a yellow update message appears at the top of the screen after the title intro, holds for 5
seconds, then tweens off-screen and destroys itself.

Failures (no network, rate limit, parse error) are silent. Gated behind a new `Preferences.checkForUpdates` toggle in
Options → Preferences (default on, desktop only — `#if desktop`, hidden on web/mobile to respect the Newgrounds/Itch/App Store update channels that handle updates for those distributions).

## Screenshots/Videos
https://github.com/user-attachments/assets/63821e3d-313f-4905-b6df-655db2d39d86

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -301,6 +301,33 @@ class Preferences
   }
 
   /**
+   * If enabled, the title screen shows a banner when a newer GitHub release is available. Desktop only.
+   * @default `true`
+   */
+  public static var checkForUpdates(get, set):Bool;
+
+  static function get_checkForUpdates():Bool
+  {
+    #if !desktop
+    return false;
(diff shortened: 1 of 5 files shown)


NotHyper-474:
This is definitely something they themselves want to implement: https://github.com/FunkinCrew/Funkin/pull/4456#issuecomment-4049471794

Issue #7516, opened 2026-05-03
Bug Report: Modded song files are wiped when loaded from Freeplay


DatGuyAlex22:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.8.4

### Description (include any images, videos, errors, or crash logs)

When using the Chart Editor to create a new song (following the [modding documentation](https://funkincrew.github.io/funkin-modding-docs/title-page.html) and pasting the necessary files into a mod folder, entering the song in Freeplay seems to wipe any of the songs files in mod/data/songs/, reverting the the easy/normal/hard metadata and chart; this can only be temporarily fixed by restarting the game or pressing F5.

https://github.com/user-attachments/assets/6c1c6dfa-e968-4d8b-9ec5-ce1e4d09bef0

Apologies for the audio delay

### Steps to Reproduce

1. Enter the Chart Editor (~ on the main menu)
2. Press File -> New Chart in the top-left corner of the screen
3. Under Create New, click on Easy/Normal/Hard/Erect/Nightmare
4. Enter necessary information into the next
5. Go to File -> Save As... or press CTRL+S to save the .fnfc file somewhere convenient
6. Create a folder in <funkin' installation>/mods
7. Rename your .fnfc to a .zip file, then extract the contents and place them into the proper directories in your mod's folder
8. Create a .json in <funkin' installation>/mods/<coolmodname>/data/levels, and add your chart to it's  "songs":[] field
9. Enter the Freeplay menu and enter/exit the song's Erect or Nightmare difficulty

Hundrec:
I'm not sure I understand.
Are you sure you're not missing any metadata for your song?

Spikefall:
> I'm not sure I understand. Are you sure you're not missing any metadata for your song?

While I may not be the original issue creator, I am having the same problems when it comes to creating variations for songs.
Any song I try to make a variation for, whether it's an erect variation or a custom modded one, it only works when hot reloading freeplay, after that the variation blanks out if you play it or play any other song until you hot reload again.
Playing it before it disappears also plays the default instrumental instead of the variations instumental.
[...]

PR #7586, opened 2026-05-17
[BUGFIX] Fix PlayState using the incorrect script when loading playlist songs with non-default variations

requazar:
## Description
This Pull Request fixes PlayState only ever using default scripts when going to a new song in the playlist.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -3670,7 +3670,8 @@ class PlayState extends MusicBeatSubState
         }
         else
         {
-          var targetSong:Song = SongRegistry.instance.fetchEntry(targetSongId) ?? throw 'Could not find a song with ID $targetSongId';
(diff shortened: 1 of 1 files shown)


requazar:
please funkin i need this my whitty bonus weekend kinda broken........

image

realvirtu:
image

Hundrec:
Thank you!

Spikefall:
Does this fix #7516? Or is this for a different thing that happens to also involves song variations?
As no modded song with variants have been working outside of reload occurences.
Which promptly then breaks the instant you enter any song and exit it, requiring another reload to fix.
Made me give up on adding variations for the time being since I can't ever really access them outside of the chart editor.

requazar:
> Does this fix #7516? Or is this for a different thing that happens to also involves song variations?

I don't think this fixes that issue, this bug is only reproducible through scripts BUT is most definitely not intended behavior(which sounds weird, lol!).

PR #7591, opened 2026-05-17
[BUGFIX] Fix the song restarting instead of ending

### LINKED ISSUE #5449: Bug Report: Song restarts instead of ending
### Issue Checklist

### Platform

Windows (Downloadable Build)

### Version

0.7.2

As the title says, the final cutscene for Stress Pico Mix doesn't plays, instead of that the song play again, but I think this only happens when you skip the first cutscene, cause I tried to replicate the bug without skip the cutscene and the final cutscene works normally, the song doesn't play again.

https://github.com/user-attachments/assets/f093e039-ac8d-4519-9f1b-d14b3700bf46

### Steps to Reproduce

1. Select Stress Pico Mix.
2. Skip the first cutscene.
3. Play the entire song.
4. Listen to the song again but without you playing it.

VirtuGuy:
## Linked Issues
Fixes #5449

## Description
This PR (hopefully) fixes the bug where the song resets instead of ending. I'm saying hopefully because this bug is weird and also because I tested this using time travel, which probably isn't the best way to test. However, time traveling right to the end of the song is 100% guaranteed to reproduce this bug as if you reproduced it accidentally, which this PR fixes that problem.

Testing is appreciated.

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2724,7 +2724,7 @@ class PlayState extends MusicBeatSubState
     // Skip this if the music is paused (GameOver, Pause menu, start-of-song offset, etc.)
     if (!(FlxG.sound.music?.playing ?? false)) return;
 
-    var timeToPlayAt:Float = Math.min(FlxG.sound.music.length,
+    var timeToPlayAt:Float = Math.min(FlxG.sound.music.length - 1,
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
This might be the one (for real this time (actually))

CEliuxJV:
FINALLY!!! I hope this issue don't get re-opened... I hope we don't get stuck in the loop again

realvirtu:
True...

Issue #5224, opened 2025-06-11
Enhancement: Optimization Settings (anti-aliasing, low quality, shader toggles)


Zombiepunk10:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

I have a not-so-great laptop and I'm sure there are other out there that do too, I (personally) think that there should be settings for optimizing the games quality, along with possibly being about to turn off unneeded features (such as note-splashes maybe) like a lot of games have. Certain stages (specifically most of the Erect stages) are very taxing on my performance and also my enjoyment... I just wanna play Darnell BF Mix 😭 (which has a very intense stage when it comes to shaders and is basically unplayable on my device)

charlesisfeline:
btw i had tried to add an anti-aliasing option to the game one time

didnt go as well tho

MAZ12211:
Vouch

Raltyro:
It's probably just me but the shader toggle seems unnecessary if theres already a low quality toggle, like couldnt there be a low quality of shaders if low quality is toggled?

Lasercar:
Yes, we need a shader saturation setting

Zombiepunk10:
> It's probably just me but the shader toggle seems unnecessary if theres already a low quality toggle, like couldnt there be a low quality of shaders if low quality is toggled?

yeah thats why i put "maybe shaders", there definitely needs to be some way to make the shaders not as intense at least

wislswithwolvesJr:
this would help a lot with low end devices like school Chromebooks lol

Keoiki:
the hell you playing fnf on a school chromebook 💀

wislswithwolvesJr:
> the hell you playing fnf on a school chromebook 💀


the lag really hurts and yes it is possible. (proxies)

Zombiepunk10:
> the hell you playing fnf on a school chromebook 💀

wsg keoiki, and no im not. its just a really bad laptop, i recently re-installed windows and that version runs a little bit better but still i would like Opti. settings lol

Raltyro:
oh my goodness

Starexify:
Opinion on Anti-Aliasing option from ninjamuffin99:
[...]

PR #7606, opened 2026-05-21
Shaders option in preferences

KT-Gameplay:
## Linked Issues
One part of #5224 

## Description
This PR adds the option to toggle shaders in Preferences

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -604,6 +604,25 @@ class Preferences
     return value;
   }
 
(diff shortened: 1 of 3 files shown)


realvirtu:
Isn't this something the devs don't want?

KT-Gameplay:
that was anti aliasing

Hundrec:
We don't want a shader toggle either.

KT-Gameplay:
Or maybe add this for modding?

PR #4496, opened 2025-04-01
[ENHANCEMENT/FIX] Countdown reacts to offsets

xenkap:
Previously, the countdown wouldn't've reacted to the audio/visual offset at all.
Now it does! Amazing.

## Does this PR close any issues? If so, link them below.
None, as far as my eyes could see

## Briefly describe the issue(s) fixed.
The countdown would always be desynced if you had great audio/visual latency regardless of your offsets.
Here, the game will now check if you have it set to something other than 0, and adjust accordingly.
Which one plays first depends on if your offset is positive or negative (obviously)

## Include any relevant screenshots or videos.
[TBA]

PR #7608, opened 2026-05-21
[BUGFIX] Make Countdown react to offsets

zxcksharks:
Previously, the countdown didn't react to the global offset at all.
Now it does! Amazing.

## Linked Issues
Still none! But an old pull request https://github.com/FunkinCrew/Funkin/pull/4496

## Description
The countdown would always be desynced if you had great audio latency regardless of your offsets.
Here, the game will now check if you have it set to something other than 0, and adjust accordingly.
Which one plays first depends on if your offset is positive or negative (obviously)

## Screenshots/Videos
Before:

https://github.com/user-attachments/assets/8793179d-e35b-40b0-8e34-be5d800b67bc

After:

https://github.com/user-attachments/assets/b7ad716f-c2f9-4908-89df-10db35bad1ac

### DIFF
--- a/source/funkin/play/Countdown.hx
+++ b/source/funkin/play/Countdown.hx
@@ -67,12 +67,6 @@ class Countdown
     // Handle onBeatHit events manually
     // @:privateAccess
     // PlayState.instance.dispatchEvent(new SongTimeScriptEvent(SONG_BEAT_HIT, 0, 0));
-    var offsetTimer:FlxTimer = null;
-    if (Conductor.instance.globalOffset != 0)
-    {
-      offsetTimer = new FlxTimer();
-      countdownOffsetTimers.push(offsetTimer);
-    }
 
     // The timer function gets called based on the beat of the song.
     countdownTimer = new FlxTimer();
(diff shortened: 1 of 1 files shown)


zxcksharks:
Offtopic, but while I was building the newest branch of develop, it (might have) broken my pipewire installation :(
Not sure if I should make an issue request for this because I'm not sure if this is a Funkin' issue, upstream issue, or an issue on my Linux installation...

Starexify:
> Offtopic, but while I was building the newest branch of develop, it (might have) broken my pipewire installation :( Not sure if I should make an issue request for this because I'm not sure if this is a Funkin' issue, upstream issue, or an issue on my Linux installation...

I havent gotten that issue. Not sure how compiling funkin could even "break" the sound library

zxcksharks:
> I havent gotten that issue. Not sure how compiling funkin could even "break" the sound library

Seems like my pipewire was broken in general, and the develop build was the only app that forcibly used it.
I've been relying on pulseaudio since, and have removed most pipewire-related stuff (and it works!)

EDIT: Fixed my pipewire installation entirely! Funkin' wasn't the problem remotely :D

PR #7614, opened 2026-05-22
[CHORE] Remove the instrumental resync

### LINKED ISSUE #5449: Bug Report: Song restarts instead of ending

realvirtu:
>[!NOTE]
FunkinCrew/lime#94 is needed for this to work properly

## Description
This PR removes the instrumental resync. Why? Because it's unnecessary. The conductor ALWAYS tracks the instrumental time while it's playing, so the instrumental shouldn't be desynced from the chart.

Yes I tested this with a lag spike. It works just fine.
Yes I tested this with a different global offset.

image

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -1927,14 +1927,7 @@ class PlayState extends MusicBeatSubState
         }
       }
 
-      if (
-        !startingSong
-        && (
-          Math.abs(FlxG.sound.music.time - correctSync) > RESYNC_THRESHOLD
-          || Math.abs(playerVoicesError) > RESYNC_THRESHOLD
-          || Math.abs(opponentVoicesError) > RESYNC_THRESHOLD
-        )
-      )
+      if (!startingSong && (Math.abs(playerVoicesError) > RESYNC_THRESHOLD || Math.abs(opponentVoicesError) > RESYNC_THRESHOLD))
       {
         trace('VOCALS NEED RESYNC');
         if (vocals != null)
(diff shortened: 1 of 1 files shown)


realvirtu:
Thank you, Hyper. Your labeling is awesome.

KarimAkra:
I believe all these fixes we've been doing here are all bandaids and the real issue is still with-in lime's NativeAudioSource backend
I am looking into it right now, if i found nothing then i guess maybe this can be merged... if i do find smth then there's no need to do this change imo

realvirtu:
Understandable. Thank you, Karim.

Hundrec:
Let's try a change Karim made.

realvirtu:
So I know this is rejected, but knowing Karim's fix doesn't work according to Hyper, do we think this could be needed anyways?

Hundrec:
Sure, let's try it out

realvirtu:
This doesn't work when time traveling. However, I think this is needed still + I'm about to do a PR for Lime that makes this actually work.

realvirtu:
Maybe the name edit wasn't needed, but eh why not.

realvirtu:
something something i should probably rebase this onto public playtest branch

realvirtu:
almost forgot to actually switch the base branch and got so fucking scared

realvirtu:
this might need some work

Issue #7326, opened 2026-04-03
Bug Report: The game behaves strangely after running in the background for several hours


NotStampyC:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [ ] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.8.4

### Description (include any images, videos, errors, or crash logs)

For a while now i have noticed that leaving funkin open in the background for too long can cause it to be really buggy. (I may or may not have left it open for 6 hours without using it)

For example, note offsets arent registered properly so even when you hit the notes they count as misses and the score goes down but your health does go up just very slowly. 
Basicaly this can happen:
<img width="1912" height="841" alt="Image" src="https://github.com/user-attachments/assets/058333eb-e0a5-4407-a95a-ea955486abed" />

This part affects the girlfriend character as after a while gf would play the cry anim every time you hit a note and nene laughs every time you hit a note (As long as you have a Miss FC). 

<img width="1919" height="979" alt="Image" src="https://github.com/user-attachments/assets/affb6815-5886-434b-901a-01b96e8a912a" />

Also when you re-open the funkin window after like 4 hours it usualy lags for about 10 seconds before allowing me to actualy use it. 

This affects the game and any other mods (i only had a mod i made which just adds a song).

### Steps to Reproduce

Open the game
Leave it open and just do anything else
come back to it after 5 to 8 hours and you should see the lag

im not sure when you would start to see the note offsets bugging out but yea

Hundrec:
lol yeah I can confirm people ran into weird issues after leaving Windows to sleep for a very long time with the game open.
Since this can be resolved by restarting the game, this is low priority

NotStampyC:
cool

Haxogone:
Just experienced this too. Do you think you can get the L medal with this?

NotStampyC:
> Just experienced this too. Do you think you can get the L medal with this?

yes  u can. you can get a L medal while still getting a miss fc

spark-the-dragon:
wonder if this effects web build as well in some way? also had this happen to me when i first mentioned it no one believed me lol

kevinyyca:
[...]

PR #7631, opened 2026-05-26
[BUGFIX] Reset Conductor songPositionDelta on suspend/resume ticks

VICK-sys:
## Linked Issues

May mitigate one failure mode reported in #7326. Does not claim to fully resolve it.

## Description

`Conductor.update()` had no guard against pathologically large `FlxG.elapsed` ticks. On OS suspend/resume (e.g. Windows minimizing the window for hours), one resume tick carries the full suspend duration through `songPositionDelta`, inflating `getTimeWithDelta()` for one frame. Reset `songPositionDelta` when `elapsed > 1s` at the top of `update()`. Threshold is well above any normal frame, so this only fires on a true suspend/resume edge. Does not touch OpenAL state, where #7326's persistent corruption likely lives.

### DIFF
--- a/source/funkin/Conductor.hx
+++ b/source/funkin/Conductor.hx
@@ -421,6 +421,8 @@ class Conductor
    */
   public function update(?songPos:Float, applyOffsets:Bool = true, forceDispatch:Bool = false):Void
   {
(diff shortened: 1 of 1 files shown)


TechnikTil:
This is kind of useless now that `songPositionDelta` is gone.

PR #7634, opened 2026-05-26
[ENHANCEMENT] Note Quantization

### LINKED ISSUE #2933: Enhancement: Note Quantization
DDR, StepMania, and Etterna all color their notes based on when in a bar it happens (so 4th notes, 8th, 12th, 16th, etc.)

Would like to see this as an option. This likely depends on supporting full customization of the arrows though.

Examples of it I've seen in the wild are in Hit Single Real and in Forever Engine (as far as I can tell, Hit Single Real is built on Forever Engine)

Looks like Forever Engine for its Quant Notes just has it rendered as Red with the base color being exactly hue 0, so the color stuff works nice. Maybe this could serve as inspiration for custom note colors?

VICK-sys:
## Linked Issues

Addresses #2933

## Description

Adds a "Note Quantization" preference that colors notes by their quantization (4th, 8th, 16th, etc.). Applies to gameplay (including chart playtesting) and the chart editor grid, controlled by a single toggle in Options → Preferences. Default off.

## Screenshots/Videos

https://github.com/user-attachments/assets/8a37ed9f-f38e-44df-b12a-e7798705ac76

demopicture

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -95,6 +95,25 @@ class Preferences
   }
 
   /**
+   * If enabled, notes are colored by their quantization (4th, 8th, 16th, etc.).
+   * @default `false`
+   */
+  public static var noteQuantColors(get, set):Bool;
+
+  static function get_noteQuantColors():Bool
+  {
+    return Save?.instance?.options?.noteQuantColors ?? false;
+  }
+
+  static function set_noteQuantColors(value:Bool):Bool
+  {
+    var save:Save = Save.instance;
+    save.options.noteQuantColors = value;
+    Save.system.flush();
+    return value;
+  }
+
+  /**
    * If enabled, the strumline is at the bottom of the screen rather than the top.
    * @default `false`
    */
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -93,6 +93,25 @@ class Preferences
   }
 
   /**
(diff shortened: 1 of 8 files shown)


EliteMasterEric (changes requested):
This feature has been on my mind for a while so cool to see a contribution addressing it.

Aside from the issue that the colors are pretty ugly (this would have to be addressed in its own PR that provides better customization of note colors), you're creating a new shader instance for each note, which creates lots of objects that get quickly discarded, and additionally [disables an important optimization to note rendering](https://github.com/FunkinCrew/Funkin/pull/5752). Having a single shared shader object for every quarter note will draw them all at once, like in the video in the linked PR.

Starexify:
I think that firstly, the Note Customization option should be added, and then this feature :p

VICK-sys:
> This feature has been on my mind for a while so cool to see a contribution addressing it.
> 
> Aside from the issue that the colors are pretty ugly (this would have to be addressed in its own PR that provides better customization of note colors), you're creating a new shader instance for each note, which creates lots of objects that get quickly discarded, and additionally [disables an important optimization to note rendering](https://github.com/FunkinCrew/Funkin/pull/5752). Having a single shared shader object for every quarter note will draw them all at once, like in the video in the linked PR.

Anything in particular you would want from the customization of note colors as an option?

KT-Gameplay:
Note colors require RGB or some workaround right?

Hundrec:
Would you like to update this PR to base it on top of the new branch?

PR #7643, opened 2026-05-27
make discord-rpc use characters icon data

7oltan:
discord rpc uses the characters id rather then its icon id in its character json, this is an easy fix and it removes alot of the duplicate
icons in the funkin.art/discord folder and i think will help modders 

i have not playtested this pr (ik make fun of me but i cant build develop rn) so if someone can try if it works before that would be 
helpful

these are icons that do not exist in the funkin.art/discord folder so they will not show up:
all of lesserafims icon
pico pixel
week5 parents
tankmen (bloody)


AbnormalPoof (approved):
cool

PR #7645, opened 2026-05-28
[GITHUB] Bump some actions dependencies to use Node 24

MRK2009:
Since node.js 20 is deprecated, I had to update `dessant/label-actions` from v4 to v5 to fix the deprecation warning to use node.js 24.
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/actions/setup-haxe/action.yml
+++ b/.github/actions/setup-haxe/action.yml
@@ -5,7 +5,7 @@ inputs:
   haxe:
     description: 'Version of haxe to install'
     required: true
-    default: '4.3.6'
+    default: '4.3.7'
   hxcpp-cache:
     description: 'Whether to use a shared hxcpp compile cache'
     required: true
@@ -71,7 +71,7 @@ runs:
 
   - name: Restore cached dependencies
     id: cache-hmm
-    uses: actions/cache@v4
+    uses: actions/cache@v5
     with:
       path: .haxelib
       key: haxe-hmm-${{ runner.os }}-${{ hashFiles('**/hmm.json') }}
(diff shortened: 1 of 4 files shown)


TechnikTil:
i wonder if this would fix the random label unassignment 👀

MRK2009:
I'm not sure though.
But I feel like it fixed something related to that.

MRK2009:
Oh thx for changing the name of my pr to the right topic, Hundrec!

MRK2009:
Let me draft it for now.

EliteMasterEric (changes requested):
I think this needs to target `main` to apply properly, since the Github Actions runners pull their configuration from there.

Hundrec:
I'll move things around and make it work

MRK2009:
> I think this needs to target `main` to apply properly, since the Github Actions runners pull their configuration from there.

Oh mb, I thought it needs to be `develop` branch instead of `main`, but thx for pointing it out.

MRK2009:
Changed it

MRK2009:
Switched it back, because  it's so stupid
Also I don't really know how to force-push to fix it

MRK2009:
> I'll move things around and make it work

I hope you can fix my pr though.

NotHyper-474:
Just leaving an observation that the same might need to be done to funkin.assets

MRK2009:
The `ci-checkout` already uses Node 24, so yeah.

MRK2009:
> > I think this needs to target `main` to apply properly, since the Github Actions runners pull their configuration from there.
> 
> Oh mb, I thought it needs to be `develop` branch instead of `main`, but thx for pointing it out.

Yk what, I'll redo my PR, but in the `main` branch because doing this in `develop` isn't working out as planned.

PR #6914, opened 2026-02-10
[ENHANCEMENT] Make hold note script events more accessible to scripts

KoloInDaCrib:
## Linked Issues
None.

## Description
This PR adjusts the hold note script events in a couple of ways:
* Makes it so that a HoldNoteScriptEvent gets dispatched when a hold note is held, both by the player and by the opponent
* Moves the holdTimer setting to the HoldNoteScriptEvent that gets dispatched when a note is held. This is useful for charts with multiple opponents, because this allows for the main opponent to not hold the sing pose for a long period when a secondary opponent holds a note.
* Makes the HoldNoteScriptEvent for holding a note cancellable. It behaves similarly to cancelling the event of `onNoteHit`.
* Dispatches the event to the `onNoteHoldDrop` function when the opponent drops a note since a comment implies that this should be possible.

## Screenshots/Videos

https://github.com/user-attachments/assets/026d3cd8-05be-4183-9954-c72d58172723

PR #5114, opened 2025-05-19
Added "SongScore" class

An-enderman:
moved the "songScore" variable to it's own class called "SongScore"

I made this because it's a to-do

I also added the Void return type to the "applyScore" function because the function didn't have a specified return type

PR #7647, opened 2026-05-28
[FEATURE] Move `songScore` to a new class

zxcksharks:
> [!NOTE]
> Breaking change - all scripts using `songScore` must now reference `SongScore.instance.getScore()`.

## Linked Issues
N/A

## Description
In PlayState.hx, songScore used to have a TODO comment saying "Move this to a class."
Not sure if that's still needed, but I did it anyway! This allows us to add further logic into score logic when needed.

This used to be part of a bigger PR I was working on, regarding held note scoring consistency.
However, I'm holding off (giving up?) on that for now until some other PRs are pulled (namely #6914) and until my psyche recovers (my contributions to FNF last about as long as I am motivated, and I can't promise anything :sob: )

For the details on the incomplete held note scoring PR, if anyone wants to tackle it themselves if I die or just out of curiosity (it's worked on in my `accurate-sustains` branch):
- Add a field called `pendingPoints` and function `getWithPending()` to the new `SongScore` class
- Currently held notes will apply their score there as an integer, instead of at the real `points` variable
- Hold notes will also hold a new `appliedScore` field, tracking how much score they've applied to `pendingScore` so far
- Move player hold note logic to `processInputQueue` (maybe into a function called `goodHoldNoteHit()`)
- When held notes are released, calculate the diff like you would in `goodNoteHit()`, remove the `appliedScore` from the `pendingScore`, and apply the real score to `points`.
This would be a massive change in input handling, which is why I didn't do it :sweat_smile: 

## Screenshots/Videos
N/A

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -814,6 +814,9 @@ class PlayState extends MusicBeatSubState
     pauseCircle = FunkinSprite.create(0, 0, 'pauseCircle');
     #end
 
+    // Score
+    SongScore.instance.reset();
+
     // Don't do anything else here! Wait until create() when we attach to the camera.
   }
 
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -52,6 +52,7 @@ import funkin.play.notes.Strumline;
 import funkin.play.notes.SustainTrail;
 import funkin.play.notes.NoteVibrationsHandler;
 import funkin.play.scoring.Scoring;
+import funkin.play.scoring.SongScore;
 import funkin.play.song.Song;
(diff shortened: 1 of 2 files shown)


zxcksharks:
i targeted the wrong branch whoops

NotHyper-474:
Does this have any difference over https://github.com/FunkinCrew/Funkin/pull/5114? (Although it doesn't seem like that one will be updated anytime soon)

zxcksharks:
> Does this have any difference over #5114? (Although it doesn't seem like that one will be updated anytime soon)

Not much, but it does use a singleton pattern instead of a static class
(Though I don't know if I'm doing singletons right... I'm just using the interface :sweat_smile:)

PR #7648, opened 2026-05-28
[ENHANCEMENT] Make `Level.getDifficulties()` actually return the final list

TechnikTil:
## Description
This PR rewrites how Story Mode gets difficulties by making `Level.getDifficulties` return exactly what is shown to the user, which is usually the difficulties of the default variation.

By doing this, it removes the need of doing hard coded filtering in `StoryMenuState.changeDifficulty` entirely.

There is visually no difference with this PR, but it lets mods add custom difficulties to Story Mode by adding them to their default variation.

### DIFF
--- a/source/funkin/ui/story/Level.hx
+++ b/source/funkin/ui/story/Level.hx
@@ -184,45 +184,33 @@ class Level implements IRegistryEntry<LevelData>
   {
     var difficulties:Array<String> = [];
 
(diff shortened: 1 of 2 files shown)


EliteMasterEric:
It looks like this PR was made before the preview, then rebased without properly checking it.

This issue was already fixed in 4048e71. I made that commit specifically to resolve an issue with High Quality Funkin'.

Issue #7658, opened 2026-05-30
Enhancement: Allow mods to specify their own API version rule


zxcksharks:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

With the addition of pre-processing (conditional compilation/parsing) in the current release version, this only feels natural to follow.
It would greatly help mod developers and Funkin' contributors to adapt to breaking changes by forcefully allowing a wider set of versions to be available and using pre-processing to only alter the areas where a breaking change exists.
This would be somewhat similar to Minecraft's datapack versioning system, where they can specify a minimum version and maximum version to support and use overlays to bring different functionality to other versions.

PR #7659, opened 2026-05-30
[Enhancement] Add Funkin' version as compilation flag

zxcksharks:
## Linked Issues
This would not close https://github.com/FunkinCrew/Funkin/issues/7658, but is related.

## Description
In the far future where backwards compatibility may become an important focus in Funkin', allowing modders to query the current version in pre-processing would allow modders to only create one set of scripts that adapt to multiple versions at once.
Even now, when developing PRs, if you're implementing breaking changes but have scripts that rely on said breaking changes, if those scripts constantly break while switching between release and compiled versions then this may be an easy fix to use.
This relates to my enhancement issue at #7658, but it does not implement mod-specific API version rules. (Seems more like a Polymod thing, and I'm not well-versed in that codebase)

## Screenshots/Videos
N/A

### DIFF
--- a/project.hxp
+++ b/project.hxp
@@ -1011,6 +1011,9 @@ class Project extends HXProject
    */
   function configureCompileDefines()
   {
+    // Add Funkin's version as a compilation flag so Polymod scripts can use it.
+    setHaxedef("funkin", VERSION);
+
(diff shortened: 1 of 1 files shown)


EliteMasterEric (approved):
Why didn't I think of that? Smart.

Issue #7499, opened 2026-04-30
Crash Report: Opening certain stages crashes the Stage Editor


hucks5:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Compiled from GitHub Source Code

### Browser

None

### Version

0.8.5 Camera Editor Preview #1

### Description

I tried opening the Spooky Mansion stage in Stage Editor and it crashed 

<img width="1268" height="743" alt="Image" src="https://github.com/user-attachments/assets/ca3cb8e0-c12f-470f-8138-e6cfe9648b57" />

### Steps to Reproduce

1. Go to Stage Editor
2. Try to open Spooky Mansion stage
3. Crash

### Crash logs

[crash-2026-04-29-20-17-59.log](https://github.com/user-attachments/files/27224125/crash-2026-04-29-20-17-59.log)

Hundrec:
This happens with multiple stages

KT-Gameplay:
It happens whenever I load a stage with animated atlas props

VirtuGuy:
> It happens whenever I load a stage with animated atlas props

level's not being set for Paths.hx I think.

KT-Gameplay:
I'm not sure

AnimatingLegend:
interesting... i'll look more into this

PR #7674, opened 2026-06-07
[ENHANCEMENT] Implement Texture Atlas support in the Stage Editor

KoloInDaCrib:
## Linked Issues
Would fix the crash in https://github.com/FunkinCrew/Funkin/issues/7499

## Description
The PR implements support for the Texture Atlas props by doing the following:
* Completely reworking asset saving system. Previously, only BitmapDatas were saved, though now every file is saved.
* Adding the `Load Texture Atlas` option to the Object Graphic Toolbox, which also opens a small atlas settings dialog.
* Tweaking the Object Animations toolbox to account for symbol/frame label animations.
* Adopting the StageEditorObject class to extend Bopper. Admittedly this change is the least needed and was done primarily because of Bopper's better animation handling.

> [!NOTE]
> This PR requires the assets PR https://github.com/FunkinCrew/funkin.assets/pull/435!

## Screenshots/Videos

https://github.com/user-attachments/assets/103e42bf-e131-4ada-9811-de3182603e74

### DIFF
--- a/source/funkin/ui/debug/stageeditor/StageEditorObject.hx
+++ b/source/funkin/ui/debug/stageeditor/StageEditorObject.hx
@@ -1,19 +1,22 @@
 package funkin.ui.debug.stageeditor;
 
 #if FEATURE_STAGE_EDITOR
+import flixel.graphics.frames.FlxFramesCollection;
 import funkin.data.animation.AnimationData;
-import funkin.graphics.FunkinSprite;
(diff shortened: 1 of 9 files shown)


KT-Gameplay:
Please FunkinCrew i need this.
My stage kinda crashing

Hundrec (approved):
A treat for 0.8.6

PR #7676, opened 2026-06-08
[FIX] Sort `null` elements by `zIndex` correctly

### LINKED ISSUE #6671: Bug Report: Spirit's icon on Thorns is behind the icon bar. (Story Mode bug)
### Issue Checklist

- [x] I did not have any mods installed when I encountered this issue

### Platform

Linux (Downloadable Build)

### Browser

None

### Mobile Device Model

_No response_

### Mobile OS Version

_No response_

### Version

0.8.0 and 0.8.1

I was replaying through the base game and for some reason Spirit's icon got placed behind the icon bar instead of being in front of it. This only happens in the the story mode version of the song and not in the freeplay menu version of it. (This probably happens on the other builds as well but I have to wait and see about it though.)

### Steps to Reproduce

1.) Open the game

2.) Play through the story mode version of Thorns.

Image

NotHyper-474:
## Linked Issues
Alternative - and better in the long term - fix for #6671

## Description
The problem with the current sort method is that when either of the elements is `null`, they are kept intact. This might make sense, but it violates the **strict weak ordering** rule that the [sort function used in the C++ target under the hood adheres to.](https://github.com/HaxeFoundation/hxcpp/issues/1347#issuecomment-4646843949)
To fix this, we instead move `null` elements to either end of the array, depending of the specified order.

<details><summary>Test Code & Results</summary>
<p>
Let's consider this tiny snippet, that represents how elements could be in a `PlayState` game scene and how they are sorted on `refresh`.

```haxe
   var sortArr = [0, 0, 1000, null, 450, 100, 100];

    sortArr.sort((a, b) ->
    {
      var order:Int = flixel.util.FlxSort.ASCENDING;
      if (a == null || b == null) return 0;
      if (a < b)
      {
        return order;
      }
      else if (a > b)
      {
        return -order;
      }
      return 0;
    });

    trace(sortArr);
```
Running this will result in the array `[0,0,1000,null,100,100,450]`. Note that `1000` is not in the place it should be.
Modifying the checks to how the PR did it will result in an array that instead is `[null,0,0,100,100,450,1000]`, which better fits the expectancy.
</p>
</details> 

## Screenshots/Videos
N/A

### DIFF
--- a/source/funkin/util/SortUtil.hx
+++ b/source/funkin/util/SortUtil.hx
@@ -36,7 +36,9 @@ class SortUtil
    */
   public static inline function byZIndex(order:Int, a:FlxBasic, b:FlxBasic):Int
   {
-    if (a == null || b == null) return 0;
+    if (a == null && b == null) return 0;
+    if (a == null) return order;
+    if (b == null) return -order;
     return FlxSort.byValues(order, a.zIndex, b.zIndex);
   }
 


TechnikTil (approved):
one line banger

EliteMasterEric (approved):

PR #6832, opened 2026-01-28
[ENHANCEMENT] Stop strumline hold splash on Bot Play / Opponent, Allow character metadata to control icon "bop"

pahaze:
## Description
I noticed a few things that I thought would be nice to be enhanced, namely disabling the opponent's strumline (and, your strumline during Bot Play) hold note splash, and allowing character metadata to control the ability to have their icon bop or not.

The former is something I just had a pet peeve with since it doesn't play the final splash after the hold notes are finished. The latter is something I personally saw fit to be fixed as I was playing around with porting some older mods, and having the ability to use animated icons OOTB now is incredibly nice, but the bop missing just felt wrong.

### Opponent strumline not showing hold note splash

https://github.com/user-attachments/assets/4bef8181-a1a8-4ba0-8185-288ddcda7223

### Bot Play strumline not showing hold splash

https://github.com/user-attachments/assets/96f19111-73bd-4252-9136-b1441bb290ef

### Set Health Icon event showing "Bop Enabled" option

https://github.com/user-attachments/assets/947363f2-b94a-4b8c-8180-459446b03d1b

## Sidenote
There's also an absolutely tiny change in the FunkinMemory that I just thought looked nicer. It just changes how the combo numbers are loaded to a for loop rather than doing 1, 2, 3, etc etc manually.
Also sorry about my horrid commit from a few years back, I was 14 and barely knew anything about coding, let alone the idea of animated sprites lol

Issue #3476, opened 2024-09-28
Enhancement: Add hold note splashes for the opponent


epokgamr:
### Issue Checklist

- [X] I have properly named the enhancement
- [X] I have checked the issues/discussions pages to see if the enhancement has been previously suggested

### What is your suggestion, and why should it be implemented?

there was a teaser for the new chart editor featuring gameplay of the new engine, with the new hold splashes. But in this teaser the enemy's hold splashes actually splash while in base game it just has the hold animation
![image](https://github.com/user-attachments/assets/3cf50c3d-147e-444d-9c3e-5d08182130bf)
picture for what im yapping about

Hundrec:
This suggestion is very similar to this one (https://github.com/FunkinCrew/Funkin/issues/2862) but with hold note splashes rather than regular note splashes.

I still think splashes in general should exclusively be used as visual feedback for the player's inputs. This means hitting a regular note near-perfectly or holding a hold note for its full length.

Hundrec:
Just like #2862, splashes are reserved for the player.

PR #7689, opened 2026-06-13
[ENHANCEMENT] Bring back strumline end splash on Bot Play / Opponent

pahaze:
## Description
Redo of #6832 since I somehow managed to create merge conflicts :exploding_head: 

I noticed something that I thought would be nice to be enhanced, disabling the opponent's strumline (and, your strumline during Bot Play's) hold note splash. I just had a pet peeve with since it doesn't play the final splash after the hold notes are finished and it looks a little weird without that final splash.

### Opponent strumline not showing hold note splash

https://github.com/user-attachments/assets/f7476b4e-56ab-427d-b01b-70b2a8375da7

### Bot Play strumline not showing hold splash

https://github.com/user-attachments/assets/949f82f4-9663-4039-bfbc-3cc8700fa461

## Sidenote
There's also an absolutely tiny change in the FunkinMemory that I just thought looked nicer. It just changes how the combo numbers are loaded to a for loop rather than doing 1, 2, 3, etc etc manually.
Also sorry about my horrid commit from a few years back, I was 14 and barely knew anything about coding, let alone the idea of animated sprites lol

### DIFF
--- a/source/funkin/FunkinMemory.hx
+++ b/source/funkin/FunkinMemory.hx
@@ -309,9 +309,16 @@ class FunkinMemory
     cacheTexture(Paths.image(style.buildJudgementSpritePath('bad') ?? 'bad'));
     cacheTexture(Paths.image(style.buildJudgementSpritePath('shit') ?? 'shit'));
 
-    for(i in 0...10) {
-      cacheTexture(Paths.image(style.buildComboNumSpritePath(i) ?? '$i'));
-    }
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(0) ?? '0'));
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(1) ?? '1'));
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(2) ?? '2'));
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(3) ?? '3'));
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(4) ?? '4'));
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(5) ?? '5'));
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(6) ?? '6'));
+    cacheTexture(Paths.image(style.buildComboNumSpritePath(7) ?? '7'));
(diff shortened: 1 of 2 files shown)


JackXson-Real:
No.

ComedyLost:
IMO the note hold covers not splashing for opponents fits better than just not having them appear at all. The current way also works similarly to note splashes since both splashes don't appear if you (Note splashes: hit anything that's not a sick, Note hold covers: release before the hold is finished) or if an opponent / botplay were to hit it.

The FunkinMemory change is a very good change tho

NotHyper-474:
I gotta agree, I don't see why remove the cover at any circumstance at all. 
I also feel like the combo change should be its own PR.

pahaze:
With your alls' thoughts in mind I did the opposite and just made it actually do the splash for Bot Play / Opponent lol

https://github.com/user-attachments/assets/b834b8ee-0f06-4cb8-9f9c-57f422ee86b7

Also moving the FunkinMemory change to its own branch / PR

JackXson-Real:
> With your alls' thoughts in mind I did the opposite and just made it actually do the splash for Bot Play / Opponent lol
> 
>  2026-06-15.15-04-25.mp4 

this is how it used to be before weekend 1 came out (it was shown in the chart editor showcase). So the decision to remove the splashes for opponent/bot was intentional. I doubt this will get accepted because adding the splashes back would just cause inconsistencies; the Sick note splashes don’t appear for opponents. And no I wouldn’t enable the Sick splashes for opponents either because they are meant to just be feedback for how well you hit a note rather than only a style choice.

pahaze:
> So the decision to remove the splashes for opponent/bot was intentional

Interesting.. IMO it looks a little weird just being cut off for the opponent
For a player it makes sense since you can let go of it early, but an opponent (/Bot Play) always hits it all the way through. But, as you said, it is feedback rather than style choice

pahaze:
> I gotta agree, I don't see why remove the cover at any circumstance at all. I also feel like the combo change should be its own PR.

Moved the change to #7697

Hundrec:
The way it works right now consistently provides the right feedback to the player.

https://github.com/FunkinCrew/Funkin/issues/3476

Issue #7653, opened 2026-05-29
Enhancement: Let decayRate in PlayState be a public var.


KT-Gameplay:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my enhancement has already been suggested
- [x] I have properly titled my enhancement

### What is your suggestion, and why should it be implemented?

Its for modding purposes.

AbnormalPoof:
... Okay...?

KT-Gameplay:
Should I make a PR for this?

realvirtu:
Probably not if it's reviewing internally and self assigned.

KT-Gameplay:
Ok

KT-Gameplay:
Will this be considered?

PR #7690, opened 2026-06-13
Allow decay rate to be customisable

KT-Gameplay:
## Linked Issues
#7653 

## Description
It uses constants too
The reason for this PR is in the linked issue itself


realvirtu:
So why would you want this?

Hundrec:
If we want this, we'll do it our own way.

PR #7697, opened 2026-06-16
[CLEANUP] Use loops in FunkinMemory over manually iterating

pahaze:
## Description
Uses for loops in FunkinMemory to cache the number graphics and the miss sounds rather than doing
```
permanentCacheSound(Paths.sound('missnote1', 'shared');
permanentCacheSound(Paths.sound('missnote2', 'shared');
permanentCacheSound(Paths.sound('missnote3', 'shared');
```

### DIFF
--- a/source/funkin/FunkinMemory.hx
+++ b/source/funkin/FunkinMemory.hx
@@ -87,9 +87,9 @@ class FunkinMemory
     permanentCacheSound(Paths.music('freakyMenu/freakyMenu'));
     permanentCacheSound(Paths.music('offsetsLoop/offsetsLoop'));
     permanentCacheSound(Paths.music('offsetsLoop/drumsLoop'));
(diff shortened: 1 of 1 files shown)


MRK2009 (approved):

PurSnake:
i wonder how it looks in raw cpp code

NotHyper-474:
> i wonder how it looks in raw cpp code
> 

Haxe usually "optimizes" this kind of loop into manual placements of the loop body (this is called "loop unrolling", that can be seen in the generated JS Source [here](https://try.haxe.org/#2c00e836)), so probably not much difference.

PurSnake:
so i dont think game really need's this chore type chnage
nothing personal but its not that much garbage code, like in chart editor
and because i have ptsd from hxcpp

AbnormalPoof:
Already done internally!

Issue #7698, opened 2026-06-16
Bug Report: Mobile Freeplay scroll speed is still FPS-dependent


CEliuxJV:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue

### Platform

Android

### Browser

None

### Mobile Device Model

Galaxy S23+

### Mobile OS Version

Android 16

### Version

0.8.5

### Description (include any images, videos, errors, or crash logs)

This bug was supposed to be fixed because #6582 was closed, but I accidentally encountered it again in 0.8.5 when I enabled Power Saving mode. which limits my FPS from 120 to 60.

https://github.com/user-attachments/assets/304ef536-c8da-4234-9065-a08cc365e831

I also tried replicating the bug without any mods installed, and it still occurs.

### Steps to Reproduce

1. Enter Freeplay
2. Swipe through the songs
3. Limit your FPS
4. Swipe through the songs again

Hundrec:
I actually noticed this on an Android after the bug was "fixed" a while back
The good news is that the inconsistency seems to be a bit better than before

PR #7746, opened 2026-07-01
[BUGFIX] fix framerate dependent freeplay scroll speed

kiryvajiravuth:
## Linked Issues
Closes https://github.com/FunkinCrew/Funkin/issues/7698

## Description
It fixes the problem that makes the freeplay scroll speed fps dependant
Where if you're on 120 fps it goes faster and if 60 fps it goes slower
By adding `framerateMultiplier` in the line 2059 it causes freeplay scroll speed to be the same across all fps

## Screenshots/Videos)
None because I currently don't have my laptop to recompile to android


kiryvajiravuth:
I do not trust about this fix so I'ma convert to draft and try to look a little bit more

kiryvajiravuth:
I cannot reproduced the bug since my phone is locked at 60fps so I am going to close this PR sadly and move on 😞

PR #7753, opened 2026-07-04
[ENHANCEMENT] Rework how `FunkinGroup` changes its children's variables

MightyTheArmiddilo:
## Description
When I first created `FunkinGroup`, I made it the way it was as I thought the way `FlxNestedSprite` did it was the only good way to do so. I found the local variables annoying, though, and so did other people I talked to that used it. So I decided to change that.

Now, in `draw()`, `FunkinGroup` saves the transformations of its children, applies new transformations like `updateChildren()` did, draws the children, and reapplies the original saved transformations. This allows for a similar functionality to be used without local variables.

I also added `members` as an alias for `children` because some other contributors expressed their want for it.

> [!IMPORTANT]
>
> The crossed out section is now obsolete as backwards compatibility has been added as per Abnormal's request! It's still recommended that mods move over, though.
> ~~This PR is a breaking change for any mods that use `FunkinGroup`!! The main thing that mods would need to change would be `localX` -> `x` (and those alike).~~ They may also need to change overrides of `updateChildren()` to `positionChildren()` or `positionChild()` or replace the setting of `customChildUpdate` with the setting of `customChildPositioning`

> [!NOTE]
>
> While testing this PR, I noticed a bug with the `clipRect` implementation Kade made, where scale and angle would break the rendering. This isn't caused by this PR as I could cause it without this PR. (Not sure if this is still broken now.)

### DIFF
--- a/source/funkin/data/character/CharacterData.hx
+++ b/source/funkin/data/character/CharacterData.hx
@@ -676,7 +676,9 @@ class CharacterDataParser
     char.debug = debug;
 
     // Call onCreate only in the fetchCharacter() function, not at application initialization.
-    ScriptEventDispatcher.callEvent(char, new ScriptEvent(CREATE));
+    var event:ScriptEvent = ScriptEvent.get(CREATE);
+    ScriptEventDispatcher.callEvent(char, event);
+    event.put();
 
     return char;
   }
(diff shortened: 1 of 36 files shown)


AbnormalPoof (approved):

TechnikTil:
This PR has merge conflicts! Please fix them.

PR #7754, opened 2026-07-05
[ENHANCEMENT] Refactor scoring system to improve moddability

nykwono:
This PR refactors how scoring is handled, separating the multiple scoring systems in-game & un-restricts only using PBOT1.

The game still defaults to using PBOT1 however users can script in their own systems for if they want to use their own special scoring system for a mod. Small but fun idea I had in mind.


EliteMasterEric:
I've concluded that there's actually no good reason to have alternate scoring systems in the first place so I just made a PR internally to remove them which is the exact OPPOSITE of what this PR does 😈

nykwono:
🫡

PR #7773, opened 2026-07-18
[ENHANCEMENT] Remove the offset hack in DropShadowShader

iLauncherDev:
## Description

I'm removing this offset hack because it doesn't replicate the old shader behavior very well making it look off on some mods

### DIFF
--- a/source/funkin/graphics/shaders/DropShadowShader.hx
+++ b/source/funkin/graphics/shaders/DropShadowShader.hx
@@ -449,12 +449,7 @@ class DropShadowShader extends FlxShader
 
         float delta = lwidth_manual(color3_light, uv, px);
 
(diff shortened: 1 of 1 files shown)


Hundrec:
Does this remove any functionality from the drop shadow shader, or is it just the same?

iLauncherDev:
> Does this remove any functionality from the drop shadow shader, or is it just the same?

no it doesn't, it just change the behavior (i will make PR for funkin.assets too)

iLauncherDev:
from what i have tested here it didn't change too much so maybe it doesn't need an change on funkin.assets

iLauncherDev:
building funkin from develop branch was a hell of compiler errors cus i'm on linux

NotHyper-474 (approved):

PR #7790, opened 2026-07-24
[BUGFIX] Use song position of exact input frame when calculating note timings

nykwono:
Currently, the game calculates note timings by comparing the milliseconds of the conductor's current position with the note's millisecond time while also compensating for input latency.

However, due to there being latency between the exact frame the input was pressed and the frame at which the input is actually being processed, the song position the game uses to calculate the ms difference is not accurate. 

This PR stores the milliseconds of the current song position for the precise input frame and uses that to calculate the note timing, which should help with bringing more accurate inputs.

### DIFF
--- a/source/funkin/input/PreciseInputManager.hx
+++ b/source/funkin/input/PreciseInputManager.hx
@@ -298,8 +298,7 @@ class PreciseInputManager extends FlxKeyManager<FlxKey, PreciseInputList>
       onInputPressed.dispatch({
         noteDirection: getDirectionForKey(key),
         timestamp: timestamp,
-        keyCode: keyCode,
-        position: Conductor?.instance?.songPosition ?? 0.0,
+        keyCode: keyCode
       });
       _dirPressTimestamps.set(getDirectionForKey(key), timestamp);
     }
@@ -317,8 +316,7 @@ class PreciseInputManager extends FlxKeyManager<FlxKey, PreciseInputList>
       onInputReleased.dispatch({
         noteDirection: getDirectionForKey(key),
         timestamp: timestamp,
-        keyCode: keyCode,
-        position: Conductor?.instance?.songPosition ?? 0.0,
+        keyCode: keyCode
       });
       _dirReleaseTimestamps.set(getDirectionForKey(key), timestamp);
     }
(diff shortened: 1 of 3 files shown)


iLauncherDev:
why not use timestamp?

nykwono:
`timestamp` is already used to help with input latency, however since `timestamp` and `Conductor.instance.songPosition` update differently (with one being nanoseconds and another being milliseconds) there is latency for the current song position as well that can't be accounted for with `timestamp`, which is why you need to account for that as well.

(Image showing that this latency does actually exist, I presume it's small for me because I run on a really good PC, but nonetheless it exists)
image

Edit: Tested it out since I was curious and yeah I was correct, the latency seems to be framerate dependent, on 60 FPS I was getting around seems around ~15-20ms while on 30 FPS it was above 30ms, so if you're on a low end device this does affect your inputs lol)
image
image

Kade-github:
> `timestamp` is already used to help with input latency, however since `timestamp` and `Conductor.instance.songPosition` update differently (with one being nanoseconds and another being milliseconds) there is latency for the current song position as well that can't be accounted for with `timestamp`, which is why you need to account for that as well.
> 
> (Image showing that this latency does actually exist, I presume it's small for me because I run on a really good PC, but nonetheless it exists) image
> 
> Edit: Tested it out since I was curious and yeah I was correct, the latency seems to be framerate dependent, on 60 FPS I was getting around seems around ~15-20ms while on 30 FPS it was above 30ms, so if you're on a low end device this does affect your inputs lol) image image

Yeah, that makes sense, since it would be the difference of how far the song goes after you press the note, and then when the note input actually gets processed.

Kade-github (approved):

Hundrec:
Can you rebase this branch to remove conflicts?

nykwono:
Got it

TechnikTil:
This PR has merge conflicts, please fix them!

TechnikTil:
This PR has merge conflicts, please fix them!

nykwono:
bro needed to say it twice 🥀

Hundrec (approved):
Feels right, let's see if anyone feels a difference

PR #7791, opened 2026-07-25
[BUGFIX] Improve input latency consistency

iLauncherDev:
## Description

This patch make sure to not call PreciseInputManager.getCurrentTimestamp for every input potentially causing inconsistent input latency if something delays some ms while the inputs are being processed


iLauncherDev:
Well i ported it to `preview/public-playtest` now

TechnikTil:
This PR has merge conflicts, please fix them!

zxcksharks:
Doesn't this defeat the point of calling `PreciseInputManager.getCurrentTimestamp()` at all?
This makes it so that inputs are frame-dependent again instead of having unique, accurate timestamps for each input.

The inputs themselves are distinctly processed _outside_ of the input callback (in the update loop), so I’m confused what you also mean by "if something delays some ms while the inputs are being processed"

PR #7792, opened 2026-07-25
[BUGFIX] Apply updated input system to Lag Adjustment menu

nykwono:
Updates Offset Menu so that the input system in it works the same as in PlayState


Hundrec:
Requested by me, based on #7367.
Thanks!

PR #7832, opened 2026-08-04
[BUGFIX] Fix music metadata backwards compat not finding the right file

KoloInDaCrib:
## Linked Issues
N/A

## Description
Previously, the backwards compatibility for the music metadata would point to `assets/music/song-metadata.json`, however in older versions, music metadata was stored in `assets/music/song/song-metadata.json`, whoopsie! This PR fixes that and also removes the data filepath from the `assets/music` lookup since I don't think that was ever supported.

## Screenshots/Videos
Dunno man im lowk kinda lazy

### DIFF
--- a/source/funkin/modding/compat/Paths.hx
+++ b/source/funkin/modding/compat/Paths.hx
@@ -1,5 +1,6 @@
 package funkin.modding.compat;
 
+import haxe.io.Path;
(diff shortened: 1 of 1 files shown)


EliteMasterEric (approved):
lgtm

PR #7842, opened 2026-08-04
[BUGFIX] Fix merging not working with new paths

### LINKED ISSUE #7835: Bug Report: _merge does not work with new asset paths
### Issue Checklist

- [ ] I did not have any mods installed when I encountered this issue

### Platform

Windows (Downloadable Build)

### Browser

None

### Mobile Device Model

_No response_

### Mobile OS Version

_No response_

### Version

preview 3

Image

Image

While the old merge file paths still work, the new ones seem to be fully broken and not merging properly (or i could be doing it wrong if so my bad) 
https://github.com/FNF-Gooey-Mix/FNF-Gooey-Mix-Test-Builds
this is what I am testing with atm

### Steps to Reproduce

try merging a remix or variation using the new file paths

KoloInDaCrib:
## Linked Issues
Closes https://github.com/FunkinCrew/Funkin/issues/7835

## Description
Funkin's new assets structure completely skips the lime one, in which the text merging was implemented. This PR copy-pastes the merging process from there into `funkin.assets.Assets`.

## Screenshots/Videos

https://github.com/user-attachments/assets/e0448a98-77fc-4a7b-9ebc-891011092486

### DIFF
--- a/source/funkin/assets/Assets.hx
+++ b/source/funkin/assets/Assets.hx
@@ -341,7 +341,11 @@ class Assets implements ConsoleClass
   {
     if (assetPath == null) throw 'Input is not a valid AssetPath, did you call Paths.txt()?';
 
-    return FunkinAssetCache.instance.getText(assetPath.toString());
+    @:privateAccess
+    var assetLib:polymod.backends.PolymodAssetLibrary = polymod.Polymod.assetLibrary;
+    var text:String = FunkinAssetCache.instance.getText(assetPath.toString());
+    
+    return assetLib == null ? text : assetLib.mergeAndAppendText(assetPath.toString().substring(6), text);
   }
 
   /**


EliteMasterEric (changes requested):
This seems like a janky way of fixing the problem.

`FunkinAssetCache.getText()` calls `openfl.utils.Assets.getText()` which should call Polymod and correctly fetch the asset, complete with replacement and merge/append.

If it's not doing that, it's a POLYMOD bug. We shouldn't try to fix it in Funkin unless we figure out what the root cause of the problem is.

KoloInDaCrib:
> This seems like a janky way of fixing the problem.
> 
> `FunkinAssetCache.getText()` calls `openfl.utils.Assets.getText()` which should call Polymod and correctly fetch the asset, complete with replacement and merge/append.
> 
> If it's not doing that, it's a POLYMOD bug. We shouldn't try to fix it in Funkin unless we figure out what the root cause of the problem is.

From my understanding it doesn't *always* call getText from openfl's assets, but rather only when the texts haven't been staged (which some are at the start of the game), so getText isn't called but rather loadText, which skips calling getText entirely.

Looking at the LimeBackend's `loadText` implementation it seems that only custom assets have support for merging/appending and that fallback assets haven't been accounted for.

image

Because I'm personally not sure if practically every text is loaded like this, I find the proposed solution in the PR to be better as it accounts for both the text that was already there and the text that is yet to be loaded. Though I'm open for other arguments against this. Not to mention that LimeBackend also always applied json merges every time the text was fetched so I don't think there's any performance loss to be found.

KoloInDaCrib:
Moved to https://github.com/FunkinCrew/polymod/pull/460

PR #7862, opened 2026-08-05
[ENHANCEMENT] Improve navigation behavior between mod lists 

MrMadera:
## Linked Issues

Nope.

## Description

This PR changes a bit how user navigates through the Mod Menu to give a more comfortable way of use it by modifying how changing lists work. Before, changing from disabled to enabled would make you be always at the first mod, even if you were selecting the mod 20th from the list. Now, you go to the mod which is exactly next to the mod you are selecting (and if there are less mods, you will select the last mod)

## Screenshots/Videos

Before:

https://github.com/user-attachments/assets/afec2203-0099-4b04-9612-697a38f5f636

After:

https://github.com/user-attachments/assets/b73637ff-45eb-4525-a65f-1e76bfad45bf

### DIFF
--- a/source/funkin/ui/modmenu/ModMenuState.hx
+++ b/source/funkin/ui/modmenu/ModMenuState.hx
@@ -1754,9 +1754,20 @@ class ModMenuState extends MusicBeatState
   }
   #end
 
+  var lastModIndex:Int = 0;
   function handleSelection():Void
   {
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
Nice, this is better than the current "Send to Top" behavior we've got!

PR #7877, opened 2026-08-06
Load characters asynchronously

nykwono:
_And finally...._

This PR adds being able to load characters asynchronously for the game to hot reload them.

Should be the last of registries (for now) that need to be loaded asynchronously, surprisingly this wasn't as hard as I thought it'd hard be.


yf-hmx:
wow

PR #7879, opened 2026-08-06
[ENHANCEMENT] Automatically add instrumental field in metadata if song is a variation

LeaNimatics:
## Description
This PR makes that, when saving a song, if it's a variation, it will automatically add the "instrumental" field into the json with the variation name, and if's not a variation of a already existing song, it will not write it to the JSON metadata.

## Screenshots/Videos

Here is a video of the feature in action !

https://github.com/user-attachments/assets/88fd390a-1559-4098-aada-ed8673bd40dd


LeaNimatics:
Wow sorry I got the wrong branch at first I was super scared sorry again for the conflict 🥹
I'm super stressed 😭

EliteMasterEric (approved):

PR #7151, opened 2026-03-14
[BUGFIX] Apply haxedef `openfl_pool_events` to decrease stuttering in GC (OLD)

charlesisfeline:
## Linked Issues
no

## Description
it's in the title. openfl's event allocation is broken, which is one of many reasons of the memory leaks. 

this may decrease the chances of memory leakage, since it reduces the memory allocated and garbage collected when using event listeners...

image

thx @Raltyro for this btw

requires https://github.com/FunkinCrew/openfl/pull/25 to compile

PR #7898, opened 2026-08-08
[BUGFIX] Apply haxedef `openfl_pool_events` to decrease stuttering in GC

charlesisfeline:
PORT OF #7151 (i forgot about this AAAAAAA)

## Description
it's in the title. openfl's event allocation is kinda bugged, which is one of many reasons of the memory leaks. 

this may decrease the chances of memory leakage, since it reduces the memory allocated and garbage collected when using event listeners...

image

_thx @Raltyro for this btw_

### DIFF
--- a/project.hxp
+++ b/project.hxp
@@ -1087,6 +1087,9 @@ class Project extends HXProject
     // Enable OpenFL's error handler. Required for the crash logger.
     setHaxedef("openfl-enable-handle-error");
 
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
This one's interesting, we'll have to experiment with it. I wonder if there's any thread safety issues tied to pooling.

Raltyro:
henlo

PR #7930, opened 2026-08-13
[BUGFIX] Fix the Stage Editor file extension adding unnecessary symbols

JVNpixels:
## Linked Issues
No Linked Issues.

## Description
When saving a stage editor file, the extension would contain a few unnecessary symbols. This would prevent the file from saving, and would require the user to manually edit the extension so that you could save the file.

### Before:
image

### After:
image


kiryvajiravuth (approved):
IT WORKS 🎉

Image

EliteMasterEric (approved):

PR #7950, opened 2026-08-16
[ENHANCEMENT] Cut note snap value in half when holding Control

ADA-Funni:
For example, if I'm on 1/16 and hold Control, the note snap becomes 1/32.

### DIFF
--- a/source/funkin/ui/debug/charting/ChartEditorState.hx
+++ b/source/funkin/ui/debug/charting/ChartEditorState.hx
@@ -593,7 +593,12 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
 
   function get_noteSnapQuant():Int
   {
-    return Std.int(SNAP_QUANTS[noteSnapQuantIndex] * (pressingControl() ? 2 : 1));
+    if (pressingControl()) {
+      final result:Int = Std.int(SNAP_QUANTS[noteSnapQuantIndex] * 2);
+      if (SNAP_QUANTS.contains(result)) return result;
+    }
+
+    return SNAP_QUANTS[noteSnapQuantIndex];
   }
 
   /**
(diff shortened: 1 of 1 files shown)


source/funkin/ui/debug/charting/ChartEditorState.hx
```diff
@@ -593,7 +593,7 @@ class ChartEditorState extends UIState // UIState derives from MusicBeatState
 
   function get_noteSnapQuant():Int
   {
-    return SNAP_QUANTS[noteSnapQuantIndex];
+    return Std.int(SNAP_QUANTS[noteSnapQuantIndex] * (pressingControl() ? 2 : 1));
```
EliteMasterEric: I think the logic for this should probably be reworked, since currently you can get values that aren't in the `SNAP_QUANTS` array.

EliteMasterEric (changes requested):

source/funkin/ui/debug/charting/ChartEditorState.hx
ADA-Funni: Ah, I anticipated that, but assumed it wasn't an issue.

I'll do something about that tomorrow

ADA-Funni: Got caught up in school stuff, but I'm able to work on it now.
Is this good?

```haxe
function get_noteSnapQuant():Int
  {
    if (pressingControl()) {
      final result:Int = Std.int(SNAP_QUANTS[noteSnapQuantIndex] * 2);
      if (SNAP_QUANTS.contains(result)) return result;
    }

    return SNAP_QUANTS[noteSnapQuantIndex];
  }
```

ADA-Funni:
Had to do some horrible dark magic to rebase this PR, and apparently Git closed the fucker for me when I didn't want it closed, but my branch works now.

Issue #7937, opened 2026-08-13
Bug Report: Holding opposite directional keys in Character Select keeps selection in place but still triggers sound


GERShark8209:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue, or the issue is related to mods

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.8.6

### Description (include any images, videos, errors, or crash logs)

If you hold both up and down or left and right while in the Character Select Menu it doesn't move which character you're selecting but still behaves as if you're selecting a new character.

### Steps to Reproduce

1. Open the Character Select Menu
2. Just hold left and right or up and down

AnimatingLegend:
interesting.. ill give this a looksie

PR #7956, opened 2026-08-17
[BUGFIX] Prevent `selectSound` from playing when holding opposite directional keys at once

AnimatingLegend:
## Linked Issues
- #7937

## Description
This PR fixes a minor audio issue in the character select menu.

### Changes:
- `update()`: Moved `selectSound.play()`/`resetDeny()` out of the four `UI_*_P` blocks, and only play the sound after if just one of Up/Down or Left/Right was pressed, not both
- `spamOnStep()`: Now calculates `moveX`/`moveY` first, then only moves the cursor and plays the sound if either is non-zero


EliteMasterEric:
This bug still exists internally but the menu got refactored heavily to where this change is no longer compatible.

PR #7983, opened 2026-08-21
[BUGFIX] Change `put()` to `finish()` in Character Select menu to fix compilation

ADA-Funni:
Eric, you missed a spot!


AbnormalPoof:
This was done in the refactored Character Select internally but I'm merging it here anyway

PR #7991, opened 2026-08-22
[BUGFIX] Disable certain plugins while in HotReloadState

ComedyLost:
## Linked Issues
Contributor's server

## Description
So I found out with Kolo's help that one of my prs (https://github.com/FunkinCrew/Funkin/pull/7906) caused a bug where after the first hot reload it required you to press F5 twice for it to work. whoopsies

This pr reverts that and just disables the plugins that would allow you to leave the state early before it has finished loading the mods

Co-authored by Kolo and Ahmed7p since they both showed ways on how to fix it


TechnikTil (approved):
works!

Hundrec (approved):
Yes very good

PR #7995, opened 2026-08-23
[BUGFIX] Preserve scripted state constructor arguments when hot reloading

nykwono:
This PR allows for scripted state to retain their previous constructor arguments when hot reloading, instead of them instantiating before with no arguments.

**Requires this PR to be merged:**
[polymod#477](https://github.com/FunkinCrew/polymod/pull/477)


Hundrec (approved):
Makes the test script work!

PR #8016, opened 2026-08-26
[BUGFIX] Fix non-scoreable notes making default notes unavailable to score

KoloInDaCrib:
## Linked Issues
Me while messing around

## Description
By overloading the note pool with notes that aren't scoreable, reusing these notes wont properly reset their non-scoreable trait for notes that should use the default notekind.

https://github.com/user-attachments/assets/a2aeff70-b19f-43f8-8278-1dcecc139277

## Screenshots/Videos

https://github.com/user-attachments/assets/e22d054d-42c2-4ffc-ab9b-3ef9556b2b5d

### DIFF
--- a/source/funkin/play/notes/Strumline.hx
+++ b/source/funkin/play/notes/Strumline.hx
@@ -1210,7 +1210,7 @@ class Strumline extends FlxSpriteGroup
 
       noteSprite.graphic.destroyOnNoUse = false;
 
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
This actually could've been really bad if we used it before the end of another song and didn't notice, so thanks for fixing this!

PR #8019, opened 2026-08-26
[BUGFIX] Fix Metronome volume toggling when mirroring notes

ahmed7p:
## Description
This PR prevents  Metronome volume from toggeling if u try to mirror notes X Axis or XY Axis using short cuts

## Screenshots
Before
{5CDBDE6B-DD4E-4EB2-8ED2-97A28C515729}
not checking making it toggel even if u don't want to toggel it
{1708F04E-9142-42CA-BDC7-8EC105793A12}


Hundrec (approved):
Good, works!

Issue #4982, opened 2025-05-04
Bug Report: Chart Editor - The waveforms in the "Offsets" and "Freeplay properties" windows are drawn out of bounds when timeline is scrolled very fast


MetaBreeze:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue

### Platform

Itch.io (Downloadable Build) - Windows

### Browser

None

### Version

0.6.4

### Description (include any images, videos, errors, or crash logs)

Pretty self explanatory, scrolling the timelines draws the waveforms out of bounds, the faster the scroll, the further away it gets drawn. The waveform seems to draw the proper section, so I think just the position is the problem

https://github.com/user-attachments/assets/d2ffcc54-2bee-471a-b55d-b4c871fe76ac

I thought this was already reported before, since I'm pretty sure I've seen this in the game for a while, but couldn't find an issue made for it

### Steps to Reproduce

1. Open chart editor
2. Open either Offsets or Freeplay properties window
3. Drag the timeline scrollbar like crazy
4. Bye bye waveform

PR #8021, opened 2026-08-27
[BUGFIX] Fix waveforms rendering outside of Chart Editor toolboxes when first opened

nykwono:
**Fixed issues:**
This probably has been reported before but I wouldn't know the IDs so.
*Note from Hundrec:* Doesn't fix #4982  

Title explains, this bug has existed since the Chart Editor was first made publicly available and still exists somehow lol.

## Old
https://github.com/user-attachments/assets/109d5bd1-e669-4c96-b3bb-2e9efa9a9e26

## New
https://github.com/user-attachments/assets/afe6c685-9907-44d6-97a7-26a987057439


Hundrec (approved):
Good!

Issue #7951, opened 2026-08-16
Bug Report: Properties set to blank or 'Unknown' aren't saved in chart data


SodaPop7720:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue, or the issue is related to mods

### Platform

Windows (Downloadable Build)

### Browser

None

### Version

0.8.6 (also works on the preview build)

### Description (include any images, videos, errors, or crash logs)

If you save a chart with the song artist set to nothing or "Unknown", the game wont save the artist flag to the metadata, causing the song to not appear in freeplay or the chart editor songs list

### Steps to Reproduce

1. Go into the chart editor
2. Load any song
3. Leave the song artist box blank, or set it to 'Unknown'
4. Save the chart and open its metadata
5. What the huh

MistrDJ:
i was running into this exact problem on 0.8.6 and the mod menu build of the game
here are my findings

<img width="1311" height="806" alt="Image" src="https://github.com/user-attachments/assets/3c70c436-57e7-4159-bfa4-823352139783" />

<img width="812" height="196" alt="Image" src="https://github.com/user-attachments/assets/f417b5f6-b5cf-433c-8ed9-b445dc785673" />

<img width="1187" height="144" alt="Image" src="https://github.com/user-attachments/assets/7e1d8fe2-57e3-481e-95e1-a3e58132be03" />

PR #8025, opened 2026-08-27
[BUGFIX] Fix crash when re-opening an exported legacy FNF chart

nykwono:
**Related Issues:**
[#7951](https://github.com/FunkinCrew/Funkin/issues/7951)

Currently, when you try importing a chart from the legacy FNF format, it crashes as the json file doesn't generate some fields like "artist", because when imported they are set to "Unknown", and due to that being the default value for the serialize for `SongData`, it crashes.

More in-deep Explanation from Mr. DJ:
image


Hundrec (approved):
Yay it fixed the crash!

PR #8026, opened 2026-08-27
[ENHANCEMENT] Add time signature support to Chart Editor osu!mania Importer

nykwono:
This PR imports the time signatures used in an osu!mania beatmap when they're imported.

Time signatures in osu!mania beatmaps are supported under the values `meters`, unfortunately osu!mania only support the time signature numerator within this value though they exist nonetheless. The importer right now parses them completely fine but does not add them into the `TimeChange` object when added.

image


Hundrec (approved):
Works!

PR #8027, opened 2026-08-27
[BUGFIX] Fix crash when mirroring an empty note selection in chart editor

realvirtu:
## Description

Basically when you press the shortcut for mirroring notes, there's no check for the amount of notes selected, causing a crash to occur.

## Screenshots/Videos

image


Hundrec (approved):
Perfect!

Issue #8039, opened 2026-08-29
Bug Report: Discord RPC will fail to be applied if the opponent is null.


Honton129:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)
- [x] I have checked the Issues/Discussions pages to see if my issue has already been reported
- [x] I have properly titled my issue
- [x] I did not have any mods installed when I encountered this issue, or the issue is related to mods

### Platform

Windows (Downloadable Build)

### Version

0.8.6

### Description (include any images, videos, errors, or crash logs)

So for example, you got a song that has no opponent defined (or it's file doesn't exists in data/characters)
Well, since now you can set custom RPC icons, having the code in charge to check that behind this is kinda wrong.

<img width="1322" height="843" alt="Image" src="https://github.com/user-attachments/assets/6852c624-1c28-4e5c-9c49-32c86713b778" />

### Steps to Reproduce

1. Play a song with a custom RPC in their metadata (or album) and a null opponent id.
2. Watch how the RPC uses the defaults no matter what.

7oltan:
making a pr to fix this rn

KaiCodesLol:
Thank you 7oltan

PR #8041, opened 2026-08-29
fix discord album not showing if opponent is null

7oltan:
## Linked Issues
https://github.com/FunkinCrew/Funkin/issues/8039

discord album does not get set if the song opponent does not exist, although base game doesn't have any song without an opponent some mods might have songs without an opponent that need discord rpc (idk what to say)


Hundrec:
Thank you for the fix, but Jigsaw says this was already fixed internally through changes he made. These changes won't be transferred to the public branch.
Sorry about that!

PR #8042, opened 2026-08-29
[BUGFIX] Fix duplicate notes on non-zero start position

nykwono:
This PR fixes a bug where if you playtest a song on the exact timestamp as one of the notes, that note will duplicate as if it's a stacked note.


Hundrec (approved):
This does fix the described issue, thanks!

Image

image

PR #8047, opened 2026-08-29
[ENHANCEMENT] Add a motion blur option for note scrolling

ADA-Funni:
## Description
This PR adds a motion blur option for notes going toward the strumline, making the notes' movements look smoother.

## Screenshots/Videos
Without Motion Blur:

https://github.com/user-attachments/assets/1cc80f42-d6ca-4813-8b75-09c232b1050c

With Motion Blur:

https://github.com/user-attachments/assets/ca94e80e-7e7e-408f-a8cb-df8748e6604c

### DIFF
--- a/source/funkin/Preferences.hx
+++ b/source/funkin/Preferences.hx
@@ -115,6 +115,27 @@ class Preferences
   }
 
   /**
+   * If enabled, any notes going toward the strumline have a blur effect placed on them, which makes their movements appear smoother.
+   *
+   * WARNING: This can cause eye strain and/or motion sickness!
+   * @default `true`
+   */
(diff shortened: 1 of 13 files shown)


JackXson-Real:
Honestly there isn't enough of a difference to warrant having an option for this. I tested this both with unlocked fps and 60fps, and couldn't notice anything despite looking very closely. The notes already scroll as smoothly as they possibly can. If they don't feel smooth, then that is most likely an issue with your monitor being low refresh rate.

NotHyper-474:
Ignore that, I put the wrong label :P

ADA-Funni:
> Ignore that, I put the wrong label :P

LMFAO

PR #8049, opened 2026-08-29
[OPTIMIZATION] Set `active` to `false` on `SustainTrail`s

ADA-Funni:
This is the same optimisation used by `NoteSprite` when the graphic isn't animated to prevent FPS drops.
`SustainTrail`s are obviously not animated, so this fix seems like a no-brainer to me.

### DIFF
--- a/source/funkin/play/notes/Strumline.hx
+++ b/source/funkin/play/notes/Strumline.hx
@@ -89,7 +89,19 @@ class Strumline extends FlxSpriteGroup
   public var conductorInUse(get, set):Conductor;
 
   // Used in-game to control the scroll speed within a song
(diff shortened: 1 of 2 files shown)


nykwono (changes requested):
You should make `Strumline.scrollSpeed` a setter function that updates the holds so the hold notes their update size because with `SustainTrail.active` being false will make them be off until you hit them.

ADA-Funni:
> You should make `Strumline.scrollSpeed` a setter function that updates the holds so the hold notes their update size because with `SustainTrail.active` being false will make them be off until you hit them.

That should be a separate pull request

PR #8052, opened 2026-08-29
[BUGFIX] Fix hold notes being off-position for 1 frame after being hit early/late

nykwono:
This PR fixes an issue where if you hit a hold note early/late, there's a chance for the hold note to appear clipped from the center of the strumline for 1 frame before it re-positions itself. It's a lot more common when hitting a note early rather than late but still can happen from both states. 

## Old
https://github.com/user-attachments/assets/a3f19ad1-a329-4f25-b892-07d59e97085a

# New
https://github.com/user-attachments/assets/c4905460-1bd6-4aa0-a613-0d3dcf0337b5

### DIFF
--- a/source/funkin/play/notes/Strumline.hx
+++ b/source/funkin/play/notes/Strumline.hx
@@ -749,7 +749,7 @@ class Strumline extends FlxSpriteGroup
           holdNote.cover.kill();
         }
       }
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
This is correct, yep!

Image

Issue #2282, opened 2024-05-05
Enhancement: make high score text more bright in freeplay


brayliniscool:
sometimes it gets kind of confusing to know which songs ive played and which songs i havent. it would be cool if you could make the freeplay high score text white (or brighter than the normal cyan color) to make it easier to see!

![image](https://github.com/FunkinCrew/Funkin/assets/91850118/dffb6172-2a28-44ec-ba18-c9c34f5b6064)

moxie-coder:
I mean this could be a setting for colorblind people

Starexify:
Wouldnt this be `topic: art and animation` aswell 🤔

PR #8053, opened 2026-08-30
[ENHANCEMENT] Lower opacity of leading zeroes in Freeplay highscore display

kiryvajiravuth:
## Linked Issues
Closes #2282

## Description
leading zeros display as 50% transparent whereas the digits that are behind leading zeros are now display as full visibility. This makes scores more readable.

### Before :

image

### After :
(i will make a new image soon)
image


JackXson-Real (changes requested):
I agree with the change, but this currently looks a bit too strange. Try lowering the brightness a bit.

kiryvajiravuth:
image

i lowered the brightness by 1 , idk if this makes a difference

kiryvajiravuth:
should i just make the leading zeros a little bit more transparent instead ?

JackXson-Real:
> should i just make the leading zeros a little bit more transparent instead ?

Yeah try doing transparent

kiryvajiravuth:
image

does this looks right?

JackXson-Real:
> image
> does this looks right?

Yes that looks a lot better

PR #8055, opened 2026-08-30
[BUGFIX] Fix hotreloading during playtest starting later than the start timestamp

KoloInDaCrib:
## Linked Issues
Me while goofing around again

## Description
Right now it's set that the target song position of the ChartEditorState while hotreloading is the song position of the conductor. However, during playtest, the same Conductor is updated, making hotreloading during playtest start in a later position of the song.

### Old

https://github.com/user-attachments/assets/015cffb2-2d83-45b1-ade6-e8aa7326d80d

### New

https://github.com/user-attachments/assets/1c1611f7-abda-45a9-801c-7976c609cd11


Hundrec (approved):
Good, it works!

PR #8060, opened 2026-08-30
[FEATURE] "Custom" Script Event

TechnikTil:
## Description
This PR adds a "custom" script event to the game. With it comes the `CustomScriptEvent` class, and the `onCustom` script class callback.

Custom script events can be assigned an ID for keeping track of what the event is for, and custom data for fetching values from the event. You can also specify if the event should be cancelable or not!


requazar (approved):
W til

PR #8065, opened 2026-09-01
[POLISH] Unify the `FunkinDebugDisplay` rounded corner radius

FuroYT:
## Description
The corner radius of the debug display was not calculated correctly as the same radius should not used for both the inside and "outside" aka the border

## Screenshots/Videos
image


Hundrec (approved):
Image

PR #8069, opened 2026-09-01
[BUGFIX] Change all ERegs to be multi-thread safe

nykwono:
Turns out there's a bunch of static ERegs used in the game. This, if not careful enough will cause multi-threads to hang due to ERegs not being safe, so this PR makes (most) ERegs localized to help with thread safety.

PR goes with this also [#1](https://github.com/FunkinCrew/thx.semver/pull/1)


NotHyper-474 (approved):

Hundrec (approved):
Oh hell yeah!

PR #2329, opened 2024-05-08
Add pixelPerfect to WiggleEffect [READ DESC!!!]

richTrash21:
### IMPORTANT!!
**This is a source part of WggleEffect fix!**
**Asset part of this fix: https://github.com/FunkinCrew/funkin.assets/pull/2**

### So what this does?
- Makes pixel perfect element of the shader optional.
- If pixel perfect flag is set - applies it to all shader modes.

### Examples
https://github.com/FunkinCrew/Funkin/assets/87835336/c4bee04a-ee6d-4a9f-b2fe-48935fbee1b2

https://github.com/FunkinCrew/Funkin/assets/87835336/f194e7ab-34c1-4618-a524-a594910aad12

### BEFORE
small_wiggle_shader_fix_showcase_BEFORE
### AFTER
small_wiggle_shader_fix_showcase_AFTER

PR #8071, opened 2026-09-02
[ENHANCEMENT/BUGFIX] Rewrite wiggle shader.

### LINKED ISSUE #7976: Bug Report: Wiggle Shader for School (Hate) Behaves Differently on Different GPUs (?)
### Issue Checklist

- [x] I did not have any mods installed when I encountered this issue, or the issue is related to mods

### Platform

Windows (Downloadable Build)

### Browser

_No response_

### Mobile Device Model

_No response_

### Mobile OS Version

_No response_

### Version

0.8.6

This happens in current Funkin' but I'm confident this isn't a new thing, it's just such a minor detail I never took note of it until my friend Rozebud pointed it out.
The shader used to produce the wiggle/wobble effect on the evil school stage behaves strangely on my computer, in comparison to other people, like Rozebud.

The _**first**_ image is how it's supposed to look (captured by Rozebud)
The _**second**_ image and video is how it appears on my end.

The pixels appear to be stretched, warped, and misshapen significantly, making elements like the school and trees lose a lot of their detail.

We think this may be an issue with how my graphics card processes the shader? I'm not entirely sure. I use an AMD Radeon RX 570 if that means anything.

This occurs on both Thorns and Thorns Erect for me, and again, I'm certain this isn't a new issue

Image

Image

https://github.com/user-attachments/assets/fae9e69b-a51b-4497-8f34-8bc33278b704

### Steps to Reproduce

1. (MAYBE) use a specific graphics card, like AMD?
2. Play Thorns on any difficulty
3. Watch background closely

ThatRozebudDude:
Rewrites the wiggle shader to be less confusing and fixes issue #7976 where the wiggle shader doesn't render correctly on certain GPUs due to precision issues. Also adds the ability to enable or disable pixel snapping across the whole shader instead of it being automatically disabled only on the dreamy effect.

Requires FunkinCrew/funkin.assets#468 to also be merged.

Here are 2 videos [MistrDJ](https://github.com/MistrDJ) recorded for me since my GPU doesn't have the issue. The first is how the shader currently looks and the second is how it looks with this fix.

https://github.com/user-attachments/assets/5b492625-a92a-4e1f-a27f-77e1f94b9387

https://github.com/user-attachments/assets/1dedaf7f-d3bb-49d3-be01-9df66d3b2a6a

This shouldn't change how the shader looks at all in the context of how it is used in Week 6 as long as you didn't have the shader issue.

### DIFF
--- a/source/funkin/graphics/shaders/WiggleEffectRuntime.hx
+++ b/source/funkin/graphics/shaders/WiggleEffectRuntime.hx
@@ -66,7 +66,15 @@ class WiggleEffectRuntime extends FlxRuntimeShader
     return time = v;
   }
 
-  public function new(speed:Float, freq:Float, amplitude:Float, ?effect:WiggleEffectType = DREAMY):Void
+  var antialiasing(default, set):Bool = false;
+
+  function set_antialiasing(v:Bool):Bool
+  {
+    this.setBool('antialiasing', v);
+    return antialiasing = v;
+  }
+
+  public function new(speed:Float, freq:Float, amplitude:Float, ?effect:WiggleEffectType = DREAMY, antialiasing:Bool = false):Void
   {
     super(Assets.getText(Paths.frag('wiggle')));
 
@@ -75,6 +83,7 @@ class WiggleEffectRuntime extends FlxRuntimeShader
     this.waveFrequency = freq;
     this.waveAmplitude = amplitude;
     this.effectType = effect;
+    this.antialiasing = antialiasing;
   }
 
   public function update(elapsed:Float)


AppleHair:
#2329 tried to do this a long time ago and got rejected:

"Phantom Arcade prefers the Before style as it is more authentic to the visuals from the graphical style being referenced. Thus, this change is rejected." - Eric

FuroYT:
> #2329 tried to do this a long time ago and got rejected:
> 
> "Phantom Arcade prefers the Before style as it is more authentic to the visuals from the graphical style being referenced. Thus, this change is rejected." - Eric

Rozebud said “This shouldn't change how the shader looks at all in the context of how it is used in Week 6” so this isn’t a visual change but rather a bug fix for AMD Graphics since shaders are not rendered the same way that Nvidia or Intel does.

AppleHair:
> Rozebud said “This shouldn't change how the shader looks at all in the context of how it is used in Week 6” so this isn’t a visual change but rather a bug fix for AMD Graphics since shaders are not rendered the same way that Nvidia or Intel does.

I'm not really sure what he ment by that, because if you look at the footage he provided you can clearly see a visual change that matches exactly to the older PR that was rejected.

If you ask me I don't really mind the change, but I'm just saying that a PR with a similar change was rejected in the past. Just as a heads-up.

ThatRozebudDude:
Considering that the way the shader looks with the bug only happens on certain hardware and the old version of the background with the effect baked as a sprite sheet I doubt that it was the intended look of the stage. Whether he actually preferred the look of the bugged version I don't know but either way that isn't how it looks on the majority of devices so either the bug should be fixed or the shader should be adjusted to make the bugged version consistent on all machines. Again based on how the shader was written and how the old version of the stage looked I believe the fixed version is the original intended look anyways.

AppleHair:
> Considering that the way the shader looks with the bug only happens on certain hardware and the old version of the background with the effect baked as a sprite sheet I doubt that it was the intended look of the stage. Whether he actually preferred the look of the bugged version I don't know but either way that isn't how it looks on the majority of devices so either the bug should be fixed or the shader should be adjusted to make the bugged version consistent on all machines. Again based on how the shader was written and how the old version of the stage looked I believe the fixed version is the original intended look anyways.

Absolutely agree with that, but again, because Phantom Arcade evidently prefers the bugged look, I think it's very possible that the devs would want to make the bugged look consistent on all machines, despite it not being the original look.

It's just my observation though. I don't actually know what they would want to do with this PR.

Hundrec:
Thanks for the discussion, we'll be sure to present this to Dave at some point to see which he prefers now

PR #8074, opened 2026-09-02
Implement `S3TC` (`.dds`) compressed textures for `desktop`

MAJigsaw77:
This PR implements `S3TC` (`.dds`) compressed textures for `desktop`, the format that's currently used is `BC3/DXT5`, the reason why this format is used over smth newer like `BC7` is because `MacOS` doesnt really support anything past `BC3` so for now we are stuck with it, this should make loading times for textures way faster. In order to compress the assets as `BC3` textures you need to download `bc7enc` from [here](https://github.com/MAJigsaw77/bc7enc_rdo/actions) for your current os, then run `./scripts/dds/compress-textures` script and to add the path to `bc7enc` program to it, like `./scripts/dds/compress-textures.sh abs/path/to/bc7enc`, make sure to run this inside the root of the game and not somewhere else.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ temp/
 docs/temp/
 export/
 astc-textures/
+dds-textures/
 RECOVER_*.fla
 shitAudio/
 .build_time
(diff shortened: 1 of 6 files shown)


TechnikTil:
Merge conflicts?

PR #8080, opened 2026-09-04
[BUGFIX] Fix hold note score being inconsistent on different FPS

### LINKED ISSUE #4512: Bug Report:

iLauncherDev:
## Linked Issues

* https://github.com/FunkinCrew/Funkin/issues/4512

## Description

Fix hold note score being inconsistent on different FPS by limiting `elapsed` to `holdNote.sustainLength`

### DIFF
--- a/source/funkin/play/PlayState.hx
+++ b/source/funkin/play/PlayState.hx
@@ -2994,8 +2994,10 @@ class PlayState extends MusicBeatSubState
         // Grant the player health.
         if (!isBotPlayMode && holdNote.scoreable)
         {
-          health += Constants.HEALTH_HOLD_BONUS_PER_SECOND * elapsed;
-          songScore += Constants.SCORE_HOLD_BONUS_PER_SECOND * elapsed;
+          var holdElapsed = Math.min(elapsed, holdNote.sustainLength / Constants.MS_PER_SEC);
+
+          health += Constants.HEALTH_HOLD_BONUS_PER_SECOND * holdElapsed;
+          songScore += Constants.SCORE_HOLD_BONUS_PER_SECOND * holdElapsed;
         }
 
         // Make sure the player keeps singing while the note is held by the bot.


zxcksharks:
I’m curious on how this solves the issue.
The logic seems to still be relying on `elapsed` which is the primary issue, making the logic framerate-dependent. It doesn’t help that the PR description doesn't explain much...

iLauncherDev:
> I’m curious on how this solves the issue.
> The logic seems to still be relying on `elapsed` which is the primary issue, making the logic framerate-dependent. It doesn’t help that the PR description doesn't explain much...

`elapsed` is only the issue when its not clamped correctly

zxcksharks:
> `elapsed` is only the issue when its not clamped correctly

How so? I'm not convinced.
The issue with scoring being inconsistent across framerates is basically:
- Higher framerate = less score than expected
  - (I think) this is because floating points are not perfect analogs for real numbers, as well as `elapsed` only being accurate to the millisecond (as of writing), which means any imperfections compound for each frame of the update loop
  - This much is actually explained in the root issue itself


I looked at your code, and this is basically what it’s doing:

Case 1: If the update skips past the end of the hold note, `elapsed` is capped to the remaining length of the note (in seconds)
 - This fixes an entirely different (but related) issue, being that frames skipping over sustain ends grant more score than expected
 - The current logic for scoring hold notes does not currently account for the end of sustains, which is what is fixed here


Case 2: The update _doesn’t_ skip past the end of the hold note, and behavior is unchanged
- Means less score is _still_ given
- Means the FPS dependency issue is _not fixed_

I think this PR is still useful, but the title is misleading and ultimately does not solve the sustain FPS issue
(also, i find it concerning you can't explain what your own code is doing ?)

Hundrec:
Thanks for the breakdown
I think that makes it clear this PR is still useful, but should be renamed to more accurately describe the issue fixed

Issue #8077, opened 2026-09-04
Crash Report: Moving an object and then exiting the Stage Editor crashes the game


Hundrec:
### Issue Checklist

- [x] I have read the [Contributing Guide](https://github.com/FunkinCrew/Funkin/blob/main/docs/CONTRIBUTING.md)

### Version

0.8.6 / `preview/public-playtest`

### Description

This is very annoying!
Every time I edit something in the Stage Editor and then without saving, the game closes without an error popup.
It has something to do with backups, like the game is trying to do something it shouldn't upon saving a backup.

### Steps to Reproduce

1. Open the Stage Editor
2. Move at least one thing
3. Try to exit
4. Click "Yes" on the confirmation prompt
5. Crash

### Crash logs

<img width="1542" height="595" alt="Image" src="https://github.com/user-attachments/assets/88784aeb-146b-410b-b49c-63d513522c7d" />

PR #8082, opened 2026-09-05
[CRASHFIX] Fix Stage Editor crashing on exit when unsaved changes exist

kiryvajiravuth:
## Linked Issues
Closes #8077

## Description
`FileFilter.extension` returns `"*.fnfs"` (with a wildcard `*` meant for file dialog filtering). Two places in `StageEditorState` were using it as a literal filename extension, producing invalid paths like `stage-editor-Limo Ride-...-.*.fnfs` which crash on file write.

Fixed by replacing `FileUtil.FILE_FILTER_FNFS.extension` with the literal string `'.fnfs'` in backup filename construction and backup file matching.

### Before :

https://github.com/user-attachments/assets/a3802ba5-466d-428c-83c0-0cabfb17891a

### After :

https://github.com/user-attachments/assets/b2ce59b1-ae6d-4676-b91c-d4d9dcf8c0f0

### DIFF
--- a/source/funkin/ui/debug/stageeditor/StageEditorState.hx
+++ b/source/funkin/ui/debug/stageeditor/StageEditorState.hx
@@ -502,7 +502,7 @@ class StageEditorState extends UIState
         var filestats:Array<sys.FileStat> = [];
         if (files.length > 0)
         {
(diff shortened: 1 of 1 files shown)


Hundrec (approved):
Good, one step closer to fixing these Stage Editor crashes!

PR #8084, opened 2026-09-05
[COMPILEFIX] Remove the useless 'dir' field from `grig.audio`

charlesisfeline:
## Description
image

### DIFF
--- a/hmm.json
+++ b/hmm.json
@@ -95,7 +95,6 @@
     {
       "name": "grig.audio",
       "type": "git",
-      "dir": "src",
       "ref": "6409f3c6d1b4c52176813d3ede86c0d34e8af2c1",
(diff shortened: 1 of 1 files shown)


biomseed:
Okay so basically the actual explanation for this is that it creates an unnecessary file called ".dev" in the directory of grig audio that makes the grig.audio library break with this error for specific people when compiling the game, so this is actually really critical
image

Hundrec:
Oh, no wonder why this specific library often breaks for people with no solution for it

MightyTheArmiddilo:
Seeing this specific thing actually get fixed makes me so happy

biomseed (approved):

realvirtu (approved):

Hundrec (approved):
Works normally for me, should be good

PR #8091, opened 2026-09-07
[BUGFIX] Fix spammy legacy asset path errors

nykwono:
The errors happen because when you use the original `funkin.Paths` class when initializing a class field, it tries to retrieve the asset path before the asset system is initialized when the game starts, thus making it error out.

This also makes the `ValidatedPaths` class account for assets in the exclude folder also.


NotHyper-474:
Needs a rebase to remove those extra commits :nails:

nykwono:
The moment I leave home :fear:

nykwono:
done :trollface:

Issue #469, opened 2021-02-20
Enhancement: Score Outline/Shadow


TaffySwordsman:
In-game score text should have some sort of black outline/shadow to make it readable over any color and any background. Weeks 1-4 this isn't really an issue since they all have dark backgrounds but week 6 is okay at best and week 5 is flat out unreadable without really focusing on the text.

![image](https://user-images.githubusercontent.com/8838370/108578380-0a8e2900-72e8-11eb-8d4f-f750035b65b3.png)

ActualMandM:
Duplicate of #228 and fixed with #336

PR #8106, opened 2026-09-11
[REFACTOR] Replace `CharacterDataParser` with `CharacterRegistry`

nykwono:
This PR completely refactors `CharacterDataParser` to be turned into a `BaseRegistry` (now `funkin.data.character.CharacterRegistry`).

**Related PRs:**
[#469](https://github.com/FunkinCrew/funkin.assets/pull/469) (Technically because of the backwards compatibility this PR provides, this doesn't HAVE to be merged but it's better for it to move to the new system nonetheless.)

## Regarding Changes:
Most of this refactor re-uses code from `CharacterDataParser` as it was already compatible. Most of what's done in `CharacterRegistry` besides that is just things for it to work as a regular registry such as replacing some functions from `CharacterDataParser` to instead be overrode functions from `BaseRegistry` (ex. `fetchCharacter -> fetchEntry()`, `listCharacterIds -> listEntryIds()`

## Backwards Compatibility
- Because this is a REALLY big breaking change, I added backwards compatibility for mods that still use `CharacterDataParser` through `funkin.modding.compat.CharacterDataParser` which simply redirects most to all functions from `CharacterDataParser` to `CharacterRegistry`, this should help to mitigate any issues to occur.
- `BaseCharacter.characterId` is now a getter property that redirects to the new `id` field that `BaseCharacter` has thanks to `IRegistryEntry`

## Breaking Changes
- I heard about this from Til so I'm not _far certain_ if this is actually the case, but `CharacterDataParser` handled parsing fields through `haxe.Json`, any unique fields added because of that parsing method will not working anymore. Again, not sure if this is even viable as a concern it's more so just what I've heard


realvirtu:
YES

TechnikTil (approved):

PR #8119, opened 2026-09-16
[ENHANCEMENT] Pressing the DELETE key while holding SHIFT reset the Global offsets back to 0 in the Pause Menu

LeaNimatics:
I was playing around with the Global offsets, and I saw that there where not any way of resseting the Global offsets back to 0.
So now I did that if we hold SHIFT and press the DELETE button, it reset back to 0.

https://github.com/user-attachments/assets/aa872fd2-4d5f-48e9-94c1-0394596748f5


Hundrec:
Unfortunately, this change is outdated due to the Pause Menu rework.
Thank you for your contribution!

